PR #166, opened 2014-12-29
Ignore "error" from a commit with changes.

rhattersley:
Running `git show --quiet ...` on a commit that includes changes (i.e. not a merge commit) gives an exit code of 1. This makes asv think there has been error when in reality everything is fine.

The same applies to `git log ...` as used to determine a range of commits.

Perhaps it would be worth making a more specific change so that only 1 is ignored? E.g. a 128 exit code from not being in a git repo still counts as an error.

### DIFF
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -86,7 +86,7 @@ def run_benchmark(benchmark, root, env, show_stderr=False, quick=False,
                  profile_path],
                 dots=False, timeout=benchmark['timeout'],
                 display_error=False, return_stderr=True,
-                error=False)
+                valid_return_codes=None)
             if errcode:
                 log.add(" failed".format(name))
             else:
--- a/asv/plugins/conda.py
+++ b/asv/plugins/conda.py
@@ -145,7 +145,8 @@ class Conda(environment.Environment):
 
     def uninstall(self, package):
         log.info("Uninstalling {0} from {1}".format(package, self.name))
-        self._run_executable('pip', ['uninstall', '-y', package], error=False)
+        self._run_executable('pip', ['uninstall', '-y', package],
+                             valid_return_codes=None)
 
     def run(self, args, **kwargs):
         log.debug("Running '{0}' in {1}".format(' '.join(args), self.name))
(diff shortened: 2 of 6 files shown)


mdboom:
Can you elaborate on this?  It seems I get an exit code of 0 for both merge and non-merge commits.  I get an exit code of 128 for non-existing commits.

rhattersley:
Perhaps it's my version of git. Using 5457d40edbfc3c62c12ea4f889a65eaf8d4a4198 and 26edc834d6b3dbae8c81292084c58a5e5f60cbf8 from this repo ...
```
$ git --version
git version 1.7.1

$ git show --quiet 5457d40edbfc3c62c12ea4f889a65eaf8d4a4198
commit 5457d40edbfc3c62c12ea4f889a65eaf8d4a4198
Merge: 26edc83 ded5803
Author: Michael Droettboom <mdboom@gmail.com>
Date:   Mon Dec 22 11:11:10 2014 -0500

    Merge pull request #162 from mdboom/fix-check-output-display
    
    Output of args in check_output

$ echo $?
0
$ git show --quiet 26edc834d6b3dbae8c81292084c58a5e5f60cbf8
commit 26edc834d6b3dbae8c81292084c58a5e5f60cbf8
Author: Michael Droettboom <mdboom@gmail.com>
Date:   Fri Dec 19 14:34:57 2014 -0500

    Fix #155: Typo

$ > echo $?
1
```

mdboom:
I've got git 2.1.0, which always returns `0` for the above, so that explains the difference.

I'm not crazy about just ignoring all error codes here.  Would you mind trying #173 to see if it fixes your original issue?

rhattersley:
> I'm not crazy about just ignoring all error codes here

Indeed! :expressionless:

> Would you mind trying #173 to see if it fixes your original issue?

Sure ... I'll give it a go and report back tomorrow.

rhattersley:
I've pushed a new commit which lets you pick specific return codes to ignore.

mdboom:
This looks good.  Thanks!

PR #167, opened 2014-12-29
Strip newline from version

rhattersley:
Without this I'm getting a trailing `\n` on the version which seems to confuse the web interface.


mdboom:
Thanks!

PR #174, opened 2015-01-06
Install dependencies in one go.

rhattersley:
This PR avoids the "version bouncing" of secondary dependencies than can occur when upgrading conda packages one-by-one. It also makes it faster. Win - win! :smile:

For example, assume a project has two dependencies: "numpy" at version 1.7, and "wibble", where "wibble" has only been built against numpy 1.8. First, asv installs numpy 1.7. Then it installs wibble, which drags numpy up to 1.8.

### DIFF
--- a/asv/plugins/conda.py
+++ b/asv/plugins/conda.py
@@ -143,12 +143,6 @@ class Conda(environment.Environment):
         args.append(package)
         self._run_executable('pip', args)
 
-    def upgrade(self, package):
-        log.info("Upgrading {0} in {1}".format(package, self.name))
-        self._run_executable(
-            'conda',
-            ['install', '-p', self._path, '--yes', package])
-
     def uninstall(self, package):
         log.info("Uninstalling {0} from {1}".format(package, self.name))
         self._run_executable('pip', ['uninstall', '-y', package],
--- a/asv/plugins/conda.py
+++ b/asv/plugins/conda.py
@@ -116,11 +116,16 @@ class Conda(environment.Environment):
 
         self.setup()
 
+        # Install all the dependencies with a single conda command.
+        # This ensures we get the versions requested, or an error
+        # otherwise. It's also quicker than doing it one by one.
+        args = ['install', '-p', self._path, '--yes']
         for key, val in six.iteritems(self._requirements):
             if val is not None:
-                self.upgrade("{0}={1}".format(key, val))
+                args.append("{0}={1}".format(key, val))
             else:
-                self.upgrade(key)
+                args.append(key)
+        self._run_executable('conda', args)
 
         self._requirements_installed = True
 
--- a/asv/plugins/conda.py
+++ b/asv/plugins/conda.py
@@ -116,16 +116,17 @@ class Conda(environment.Environment):
 
         self.setup()
 
-        # Install all the dependencies with a single conda command.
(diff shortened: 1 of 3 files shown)


asv/plugins/conda.py
```diff
@@ -116,11 +116,16 @@ def install_requirements(self):
 
         self.setup()
 
+        # Install all the dependencies with a single conda command.
+        # This ensures we get the versions requested, or an error
+        # otherwise. It's also quicker than doing it one by one.
+        args = ['install', '-p', self._path, '--yes']
         for key, val in six.iteritems(self._requirements):
             if val is not None:
-                self.upgrade("{0}={1}".format(key, val))
```
rhattersley: This seems to render the `upgrade` method unused. Since it's not used outside the class (and the same applies for the virtualenv variant) perhaps it should be removed from the abstract `Environment` class too? `Virtualenv.upgrade` could be made a private method.

I have a branch on stand-by with this simplification if you're interested.

mdboom:
This is great.  I think the same simplification can be applied to `virtualenv` as well.  And I agree that then the `upgrade` method can be removed all places (including the base class).  `asv` is still green enough I'm not guaranteeing API stability, so I'm happy to keep things trimmed.

One small issue that Travis-CI discovered -- if `len(self._requirements) == 0`, we should just avoid calling `conda` altogether.

rhattersley:
I've pushed a couple of new commits which fix the zero-dependencies bug and perform the `upgrade()` simplification.

asv/plugins/virtualenv.py
```diff
@@ -124,13 +124,13 @@ def install_requirements(self):
...
-        self.upgrade('setuptools==3.8')
+        self._upgrade('setuptools==3.8')
 
         for key, val in six.iteritems(self._requirements):
             if val is not None:
-                self.upgrade("{0}=={1}".format(key, val))
+                self._upgrade("{0}=={1}".format(key, val))
             else:
-                self.upgrade(key)
+                self._upgrade(key)
```
mdboom: I think we can do the same thing for `virtualenv` that was done for `conda`, i.e. pass all dependencies in one go.  Then we'll have fixed the original bug in both places.

rhattersley: OK. I'll make the necessary tweaks...

rhattersley: Any idea if it's safe to combine "setuptools==3.8" with the other requirements?

asv/plugins/virtualenv.py
```diff
@@ -124,13 +124,17 @@ def install_requirements(self):
...
 
-        self.upgrade('setuptools==3.8')
-
-        for key, val in six.iteritems(self._requirements):
-            if val is not None:
-                self.upgrade("{0}=={1}".format(key, val))
-            else:
-                self.upgrade(key)
+        self._run_executable('pip', ['install', '--upgrade',
+                                     'setuptools==3.8'])
```
rhattersley: I've left this as a seperate operation for now.

mdboom: Yeah -- I think that's probably best.  `setuptools` can end up changing the behavior of `pip` itself, so it probably does make sense to upgrade that first, then move along with the project-specific stuff.

mdboom:
Thanks for this!  Merging.

Issue #169, opened 2014-12-29
Long conda environment name


rhattersley:
I'm trying to benchmark a project which has [several](https://github.com/SciTools/iris/blob/333a5fc8c31746412f684a2137f5cbe8e2685a49/conda-requirements.txt) dependencies. If I want asv to create a conda environment it seems I have to list the dependencies in the `matrix` entry of `asv.conf.json`. Unfortunately, *all* the entries in this matrix end up in the conda environment name. This makes the environment name too long for conda's automatic prefix re-writing(*).

How about only the multi-valued dependencies are used in the environment name instead?

*) For example:
```
> conda create -n this-is-way-too-long-for-the-automatic-prefix-stuff-to-cope-and-causes-an-exiciting-error-to-be-thrown udunits=2.*
Fetching package metadata: ...
Solving package specifications: .
Package plan for installation in environment /data/local/ithr/miniconda/envs/this-is-way-too-long-for-the-automatic-prefix-stuff-to-cope-and-causes-an-exiciting-error-to-be-thrown:

The following NEW packages will be INSTALLED:

    udunits: 2.2.17-4 scitools

Proceed ([y]/n)? 

Linking packages ...
Error: ERROR: placeholder '/home/travis/miniconda/envs/_build_placehold_placehold_placehold_placehold_place' too short in: udunits-2.2.17-4

```

mdboom:
Thanks for the report.  Indeed, that is a problem, which I can confirm.  Will have to play with the best solution -- naming each environment after a particular instance of the dependency matrix is helpful if that matrix is later expanded and you don't want to throw out existing results.  But it may be that that doesn't matter for the environments themselves, only the results.

pv:
Fixed in gh-181

PR #176, opened 2015-01-07
Only include pinned versions in environment name.

rhattersley:
re: #169.

For a dependency matrix such as...
```
"matrix": {
    "a": [],
    "b": ["5"],
    "c": ["2.24", "2.26"]
}
```
... this PR changes the resulting environment names from:
 - py2.7-a-b5-c2.24
 - py2.7-a-b5-c2.26

to:
 - py2.7-b5-c2.24
 - py2.7-b5-c2.26

NB.
 - As mentioned on #169, this affects both the name of the testing environment and the name of the result files.
 - This PR does *not* change the circumstances under which result file names will change (e.g. going from an unpinned version to a pinned version). The obvious exception to this is the initial transition from the current naming scheme to the scheme proposed in this PR.

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -55,8 +55,6 @@ def get_env_name(python, requirements):
     for key, val in reqs:
         if val is not None:
             name.append(''.join([key, val]))
-        else:
-            name.append(key)
     return '-'.join(name)
 
 


mdboom:
Thanks for submitting this.

A couple of thoughts: This doesn't really solve the problem when there are a large number of explicitly-versioned dependencies.  Also, even though it's not supported at present since the existing dependency matrix stuff is very simplistic, it may be the case that the user wants to create environments both with and without a particular dependency (to benchmark a Python vs. Cython implementation of an algorithm, for example).  What if the environment directory were the MD5 hash of the list of dependencies?  Then we can limit the number of characters to something sane in all cases, even in terrible ones.  Sorry to share this brain dump only after you filed the PR -- I didn't have a chance to get back to this yesterday.

There's also the transition problem you mention.  For the environment directories it's not a big deal -- you just throw out the old environments and there's a rebuild time hit for the first time with the new method.  For the results though, the files would all have to be renamed.  Maybe we could do this name simplification --- whether as you've proposed here or with my hash proposal -- only on the environment name (where conda cares) and not on the result files.

rhattersley:
> This doesn't really solve the problem when there are a large number of explicitly-versioned dependencies.

True. But I don't have that problem. :wink: 

> it may be the case that the user wants to create environments both with and without a particular dependency

That certainly sounds plausible.

> What if the environment directory were the MD5 hash of the list of dependencies?

> Maybe we could do this name simplification --- whether as you've proposed here or with my hash proposal -- only on the environment name (where conda cares) and not on the result files.

:+1: If necessary there's always the option of writing out one or more files that describe the contents of the environment directory.

> Sorry to share this brain dump only after you filed the PR -- I didn't have a chance to get back to this yesterday.

Not a problem at all. I needed to make these changes anyway so I can keep making progress. :grinning:

mdboom:
Closed in favor of #181.

rhattersley:
:+1:

Issue #175, opened 2015-01-06
Incremental benchmarking


rhattersley:
Is there a way, or a plan for a way, to build up benchmarking results for a given machine across multiple invocations?

I'd like to be able to do something like the following and have the results combined.
```
$ asv run --bench my_slow_benchmark --quick
$ asv run --bench some_other_benchmark v3..v4
etc.
```
Currently the second invocation just throws away the results from the first. :unamused:

mdboom:
Yeah, this should probably be made to work.  The `--bench` argument wasn't very well thought through and is really only useful for testing that certain benchmarks are working or to get quick one-off results.  The result-saving code pretty much assumes that you're running all of the benchmarks in a single invocation.  This may not be too difficult -- it should just be a matter of loading and existing results and merging the results into it rather than writing a new file each time.

mdboom:
This is totally untested, but I think something like this should do the trick:

```diff
	Modified   asv/results.py
diff --git a/asv/results.py b/asv/results.py
index 93f140b..24af8ea 100644
--- a/asv/results.py
+++ b/asv/results.py
@@ -213,6 +213,10 @@ class Results(object):
         """
         path = os.path.join(result_dir, self._filename)
 
+        if os.path.exists(path):
+            old_results = self.load(path)
+            self.add_existing_results(old_results)
+
         util.write_json(path, {
             'results': self._results,
             'params': self._params,
@@ -248,6 +252,18 @@ class Results(object):
         obj._filename = os.path.join(*path.split(os.path.sep)[-2:])
         return obj
 
+    def add_existing_results(self, old):
+        """
+        Add any existing old results that aren't overridden by the
+        current results.
+        """
+        for key, val in six.iteritems(old.results):
+            if key not in self._results:
+                self._results[key] = val
+        for key, val in six.iteritems(old._profiles):
+            if key not in self._profiles:
+                self._profiles[key] = val
+
     def rm(self, result_dir):
         path = os.path.join(result_dir, self._filename)
         os.remove(path)
```

I may not have time to work this into a proper PR with a test, but maybe that helps show the way of what I'm thinking.

pv:
Seems to be covered by gh-188 and gh-211

PR #177, opened 2015-01-07
Incremental benchmarking

rhattersley:
Extends the patch supplied in #175 to ensure the `results/benchmarks.json` always contains *all* the benchmark definitions, instead of just the benchmarks selected in the last invocation.

(It also removes the apparently unused `regex` argument from `Benchmarks.load()`. This isn't necessary for the PR to work so I'm happy to drop that commit, but the existence of the `regex` argument did slow down the process of finding the right bit of code to modify.)

If this looks to be a sensible change them I'm happy to look at testing.

### DIFF
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -261,8 +261,8 @@ class Benchmarks(dict):
         metadata about the discovered benchmarks, in the results dir.
         """
         path = self.get_benchmark_file_path(self._conf.results_dir)
-        util.write_json(path, self, self.api_version)
-        del self['version']
+        util.write_json(path, self._all_benchmarks, self.api_version)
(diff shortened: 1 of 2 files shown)


mdboom:
Thanks.  This looks good -- the removal of the `regex` thing initially threw me for a loop, but I think it makes sense as you've done it.

Travis tells us that this broke the `rm` command -- since now it removes things but then adds them on top of the existing file, rather than replacing it.  Perhaps the solution there is add an `update` method (which adds results to an existing file as `save` does in this PR) and maintain the old behavior of `save`, which is to replace.  And then update the appropriate uses of `save` to use `update` instead.

mdboom:
Closing in favor of #188.

Issue #178, opened 2015-01-07
Explicitly select virtualenv or conda


cdeil:
The docs say that currently the choice for `conda` is made automatically if that is on the PATH.
This doesn't seem to work for me.

So far I was using `virtualenv` and now I switched to `conda` using `rm -r env` and then `asv` still said it would try to use virtualenv "Creating virtualenv for py2.7-astropy-numpy-scipy." and then failed like this:
https://gist.github.com/cdeil/96eb55690b5288af2b61#file-gistfile1-txt-L6

Maybe it would be better to let the user make an explicit choice for virtualenv / conda via a configuration option? (This might also be better in case in the future other "backends" like e.g. Docker are added.)
And is this info which one was used stored somewhere in the results?

mdboom:
I'm not opposed to this idea, by adding a new key to `asv.conf.json`.  This has caused other related issues as well in the past.

PR #180, opened 2015-01-08
Allow for overriding the environment type

mdboom:
Possible fix for #178.

@cdeil: Would you mind testing this to see if it fixes your original issue?

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -106,6 +106,9 @@ def get_environment_class(conf, python):
         raise ValueError(
             "Unknown environment type '{0}'".format(conf.environment_type))
     else:
+        log.warn(
+            "No `environment_type` specified in asv.conf.json. "
+            "This will be required in the future.")
         # Try the subclasses in reverse order so custom plugins come first
         for cls in list(util.iter_subclasses(Environment))[::-1]:
             if cls.matches(python):
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -72,10 +72,46 @@ def get_environments(conf):
     ----------
     conf : dict
         asv configuration object
+
+    python : str
+        Python version specifier.  Acceptable values depend on the
+        Environment plugins installed but generally are:
+
+        - 'X.Y': A Python version, in which case conda or virtualenv
+          will be used to create a new environment.
+
+        - 'python' or '/usr/bin/python': Search for the given
+          executable on the search PATH, and use that.  It is assumed
+          that all dependencies and the benchmarked project itself are
+          already installed.
     """
     for python in conf.pythons:
         for env in get_environments_for_python(conf, python):
             yield env
+
+
+def get_environment_class(conf, python):
+    """
+    Get a matching environment type class.
+
+    Parameters
+    ----------
+    conf : dict
+        asv configuration object
+    """
+    if conf.environment_type:
+        for cls in util.iter_subclasses(Environment):
+            if cls.tool_name == conf.environment_type:
+                return cls
+        raise ValueError(
+            "Unknown environment type '{0}'".format(conf.environment_type))
+    else:
+        # Try the subclasses in reverse order so custom plugins come first
+        for cls in list(util.iter_subclasses(Environment))[::-1]:
+            if cls.matches(python):
+                return cls
+        raise ValueError(
+            "No way to create environment for '{0}'".format(python))
 
 
 def get_environments_for_python(conf, python):
@@ -100,14 +136,9 @@ def get_environments_for_python(conf, python):
           that all dependencies and the benchmarked project itself are
           already installed.
     """
-    # Try the subclasses in reverse order so custom plugins come first
-    for cls in list(util.iter_subclasses(Environment))[::-1]:
-        if cls.matches(python):
-            for env in cls.get_environments(conf, python):
-                yield env
-            break
-    else:
-        raise ValueError("No way to create environment for '{0}'".format(python))
+    cls = get_environment_class(conf, python)
+    for env in cls.get_environments(conf, python):
+        yield env
 
 
 class PythonMissingError(BaseException):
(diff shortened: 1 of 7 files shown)


asv/template/asv.conf.json
```diff
@@ -36,6 +36,13 @@
     // environments in.  If not provided, defaults to "env"
     // "env_dir": "env",
 
+    // The tool to use to create environments.  May be "conda",
+    // "virtualenv" or other value depending on the plugins in use.
+    // If missing or the empty string, the tool will be automatically
+    // determined by looking for tools on the PATH environment
+    // variable.
+    // "environment_type": "",
```
cdeil: The new "environment_type" option doesn't show up in the reference docs ... I think you have to add it in `docs/source/asv.conf.json.rst`, no?

An unrelated point is that there's this warning:
```
/Users/deil/code/asv/docs/source/asv.conf.json.rst:66: WARNING: Inline literal start-string without end-string.
```
Maybe run `build_sphinx` on travis-ci or if it already does make warnings fail the build?

docs/source/using.rst
```diff
@@ -167,30 +169,32 @@ over and dependencies are installed into the environment.  The
...
-    always runs its benchmarks in a virtual environment, so it will
-    not change what is installed in any of the python environments on
-    your system.
+Environments can be created using different tools.  By default,
+``asv`` ships with support for `anaconda
+<https://store.continuum.io/cshop/anaconda/>`__ and `virtualenv
+<https://pypi.python.org/pypi/virtualenv>`__, though plugins may be
+installed to support other environment tools.  By default, if
+``conda`` is found in your ``PATH``, it will be used, and if not, fall
+back to ``virtualenv``.  To override this behavior, the
```
cdeil: This doesn't work for me.

If I run `asv run EXISTING` for https://github.com/gammapy/gammapy-benchmarks with `conda` in the `PATH` it still uses `virtualenv`, not `conda` as claimed here.
@mdboom - Can you reproduce this behaviour?

asv/template/asv.conf.json
mdboom: I've added this to the reference docs and fixed the syntax error.

docs/source/using.rst
mdboom: No -- I would like to get to the bottom of why it's failing for you.

What does:

```
>>> from asv import util
>>> asv.which('conda')
```

give for you?

cdeil: The issue is that I have `conda` on my PATH without `asv` installed:
```
$ python
Python 3.4.2 |Anaconda 2.1.0 (x86_64)| (default, Oct 21 2014, 17:42:20) 
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from asv import util
  File "<stdin>", line 1, in <module>
ImportError: No module named 'asv'
```

So `asv` uses Macports Python:
```
eduroam-3-163:gammapy-benchmarks deil$ which asv
/Users/deil/Library/Python/3.4/bin/asv
eduroam-3-163:gammapy-benchmarks deil$ cat /Users/deil/Library/Python/3.4/bin/asv
#!/opt/local/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'asv==0.1','console_scripts','asv'
__requires__ = 'asv==0.1'
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.exit(
        load_entry_point('asv==0.1', 'console_scripts', 'asv')()
    )
```
Any way to avoid that "gotcha" or print some warning message?
Or at least mention this in the docs?

mdboom: Perhaps there's some confusion here.  When the docs say `conda` on your `PATH`, I mean explicitly the `conda` executable, not the Python interpreter that `conda` provides.  I should be able to run `conda` or `which conda` at the command prompt and have that work.

In any event, it shouldn't matter which python `asv` is running under (except for the testing example I gave above) in order for it to find `conda`.  Any Python interpreter shouldn't be messing with `PATH`.

Can you try:

```
$/opt/local/bin/python
>>> from asv import util
>>> asv.which('conda')
```

cdeil: Does this help?
```bash
$ which conda
/Users/deil/software/anaconda/anaconda3/bin/conda
$ which python
/Users/deil/software/anaconda/anaconda3/bin/python
$ python -c 'from asv import util; print(util.which("conda"))'
  File "<string>", line 1, in <module>
ImportError: No module named 'asv'
$ /opt/local/bin/python -c 'from asv import util; print(util.which("conda"))'
/Users/deil/software/anaconda/anaconda3/bin/conda
```

mdboom: Hmm... puzzling.  That's exactly what `asv` does to check the existence of `conda`.  It then goes on to do:

```
$ conda create --yes -p /tmp/foo python=2.7 --dry-run
```

Does that fail?

If that's not it, all I can suggest is to try to step through `plugins/conda.py:Conda:matches` to see why it's returning `False`.

asv/environment.py
```diff
@@ -77,6 +77,42 @@ def get_environments(conf):
...
+    """
+    if conf.environment_type:
+        for cls in util.iter_subclasses(Environment):
+            if cls.tool_name == conf.environment_type:
+                return cls
+        raise ValueError(
+            "Unknown environment type '{0}'".format(conf.environment_type))
+    else:
+        # Try the subclasses in reverse order so custom plugins come first
+        for cls in list(util.iter_subclasses(Environment))[::-1]:
```
mdboom: Yes -- we should probably transition to being explicit.  I think in the meantime, I'll just add a warning.

cdeil:
:+1: ... i.e. explicitly setting the environment type to use conda works for me.

PR #181, opened 2015-01-08
Use shorter environment directory names

mdboom:
This is a possible fix for #169 and an alternative to #176.

It uses an md5 hexdigest for the directory name, and then puts information about the environment in a file `asv-env-info.json` within the environment directory (though this isn't currently used for anything).


rhattersley:
Nicely ironic branch name you chose :wink:

Issue #179, opened 2015-01-07
Make it easy to do multiple measurements on a given benchmark


cdeil:
I've started to write benchmarks for Gammapy and decided that mostly high-level benchmarks running compute-heavy end-user analysis tasks are of interest and for each of those I'm interested in:
- CPU time
- Wall clock time
- Memory use

@mdboom Would it make sense to implement a mechanism that makes it easy to do several benchmarks (namely the three I mentioned) without having to write boilerplate code? Maybe via a decorator listing which measurements I want?

mdboom:
This is a good idea.  I'll self-assign if I find some free time to work on it, otherwise all patches welcome!

erwanp:
I have a similar need : a single benchmark for the [spectroscopy RADIS code](https://github.com/radis/radis-benchmark) can take up to 250s ; and computing it twice for PeakMemory and CPU-wall-time takes a lot of time. 

Would it be possible and still accurate to measure both the PeakMemory and the CPU-wall-time on a single run ?

PR #182, opened 2015-01-08
Allow support for multiple benchmark configurations

mdboom:
Possible fix for #179.

@cdeil: Does this address your original need?

### DIFF
--- a/docs/source/writing_benchmarks.rst
+++ b/docs/source/writing_benchmarks.rst
@@ -317,7 +317,7 @@ Or, you can put all of the benchmarks in a suite::
       ]
 
       def multi_range(self):
-          range(100000)
+          return range(100000)
 
       def multi_for_loop(self):
           x = 0
--- a/docs/source/writing_benchmarks.rst
+++ b/docs/source/writing_benchmarks.rst
@@ -276,7 +276,7 @@ and memory usage::
   import asvtools
 
   def multi_range():
-      range(100000)
+      return range(100000)
   multi_range.types = [
       ('process_time', 'time'),
       ('wall_time', 'time', {'timer': asvtools.wall_time}),
@@ -295,11 +295,14 @@ you can assign the types to a variable and reuse that::
   ]
 
   def multi_range():
-      range(100000)
+      return range(100000)
   multi_range.types = my_types
 
-  def multi_xrange():
-      xrange(100000)
+  def multi_for_loop():
+      x = 0
+      for i in range(100000):
+          x *= i
+      return x
   multi_xrange.types = my_types
 
 Or, you can put all of the benchmarks in a suite::
@@ -316,5 +319,8 @@ Or, you can put all of the benchmarks in a suite::
       def multi_range(self):
           range(100000)
 
-      def multi_xrange(self):
-          xrange(100000)
+      def multi_for_loop(self):
+          x = 0
+          for i in range(100000):
+              x *= i
+          return x
--- a/docs/source/writing_benchmarks.rst
+++ b/docs/source/writing_benchmarks.rst
@@ -185,7 +185,9 @@ possible, with as much extraneous setup moved to a ``setup`` function::
   ``time.process_time`` (or a backported version of it for versions of
   Python prior to 3.3), but other useful values are
   `timeit.default_timer` to use the default ``timeit`` behavior on
-  your version of Python.
+  your version of Python.  The ``asvtools`` module, which you can
+  import from your benchmark suite, contains the aliases
+  ``process_time`` and ``wall_time`` which can also be used here.
 
   On Windows, `time.clock` has microsecond granularity, but
   `time.time`'s granularity is 1/60th of a second. On Unix,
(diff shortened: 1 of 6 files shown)


cdeil:
Travis-ci fails: https://travis-ci.org/spacetelescope/asv/jobs/46343860#L162

docs/source/writing_benchmarks.rst
```diff
@@ -244,3 +246,69 @@ garbage collector at a given state::
...
   for display in the web interface.
+
+Multiple
+````````
+
+There is also a special benchmark type for benchmarks that you want to
+test in multiple ways, for example to test both for memory usage and
+run time.  "Multiple" benchmarks use the prefix ``multi`` and have a
+required attribute ``types``, which is a list of benchmark types to
+run.
```
cdeil: `type` is a Python keyword, so maybe rename the `types` attribute to something else like `multi` or `measurements`?
(not sure ... if you like `types` better it's totally fine)

cdeil:
I tried adding the `multi_range` example from the docs to `gammapy-benchmarks` and got this error:
https://gist.github.com/cdeil/12a986b620892b773cc3

mdboom:
> I tried adding the multi_range example from the docs to gammapy-benchmarks and got this error:
https://gist.github.com/cdeil/12a986b620892b773cc3

Did you import `asvtools`?  I should add that to the docs.

cdeil:
Indeed ... I had forgotten to put this line:
```python
import asvtools
```
My editor shows this as an error while I'm writing the benchmarks ... I guess there a way to make `from asv import asvtools` work in the virtualenvs (that's where this is located when installing asv)?

cdeil:
On Mac I see this report a memory usage of 0 in both Python 2 and 3 !?
```python
def multi_range():
    range(100000)
multi_range.types = [
    ('process_time', 'time'),
    ('wall_time', 'time', {'timer': asvtools.wall_time}),
    ('memory', 'mem')
]
```

mdboom:
`asvtools` is the only part of `asv` that's available to the benchmarked environment, so I thought it best not to "fake import" it under the `asv` package (since then it might confuse people into thinking they could get other things in `asv` there.  If you have a better thing to suggest, let me know, but I think `from asv import asvtools` is not a good idea.

> On Mac I see this report a memory usage of 0 in both Python 2 and 3 !?

Yes -- the object should be returned.  I'll add that to the example.

cdeil:
If you think `import asvtools` is better, fine with me.

> the object should be returned

Is there a way to measure peak memory usage?
Otherwise to measure the memory usage for this example doesn't make sense, no?
(I haven't started doing memory measurements with asv ... sorry for my ignorance on how what it measures)

mdboom:
The memory benchmarking tool is designed to measure the size of an object (returned from the benchmark function/method).  There currently isn't a way to measure peak memory usage, but it would be a handy thing to add.  

It's a little tricky to do reproducibly with a Python program, however.  The typical unix way is to look at `ru_maxrss` (the maximum resident set size).  However, Python will often not return memory to the OS unless under memory pressure to do so.  So it's quite possible on a machine with lots of RAM to have a high `ru_maxrss`, even though the actual live objects at any given time is much lower.  And one would expect to get a much lower number on a machine with less RAM.  So some research needs to be done to find a good method.  Note, however, that the Python `memory_profiler` module on PyPI ignores this and many still find it useful.  It probably matters less for code dominated by large Numpy arrays than dominated by lots of little Python objects.

cdeil:
This now works for me.

I think it would be nice to put examples where there's a non-zero memory measurement and where wall and CPU time are obviously different, so that users trying this out for the first time see the difference.

Side comment: do you understand why Python 3 is slower that Python 2 for such a simple list iterator example? Is this real or an artifact of how the measurements are done?

cdeil:
Would `return locals()` be a useful way to roughly get the "memory usage in that function" in many cases?

Or would this be useful?
https://github.com/ianozsvald/ipython_memory_usage

(I'll start looking at how memory measurements work tomorrow and do some tests with asv ...

mdboom:
> I think it would be nice to put examples where there's a non-zero memory measurement 

In the fixed version of the examples, it returns the value, so indeed it has a non-zero memory measurement

> wall and CPU time are obviously different

Yes, that probably makes sense.  Maybe be calling `sleep` or something...

> do you understand why Python 3 is slower that Python 2 for such a simple list iterator example?

I suspect it has to do with using long integers for everything.  Integer arithmetic in Python 3 is generally much slower than Python 2.

> Would return locals() be a useful way to roughly get the "memory usage in that function" in many cases?

It might, but I'm not sure I understand the use case.  Generally, you would want to track the memory usage of some class in your library and ignore all else -- at least that's the use I had in mind when I wrote the memory benchmarker.

cdeil:
I'm more interested in system-benchmarks than micro-benchmarks.

I.e. I'd like to run a science analysis example and it would take a few seconds or a minute ... in this case I'd like asv to tell me that this analysis takes 14 seconds and uses a max of 400 MB RAM:
image

cdeil:
PS: this plot was made with psrecord which uses this to measure memory: https://github.com/astrofrog/psrecord/blob/620b7c1683fe1608b90e1a1b618bcc7cba2c5db7/psrecord/main.py#L177

mdboom:
Now that I understand things better (that we were talking about different kinds of memory usage measurements all along), I think this is probably not the right approach for what @cdeil wants to do.  He doesn't really need the full generality of different kinds of benchmarks that `asv` has.  The data he wants could probably better be obtained from a single benchmarking run, and could be handled with a simpler extension to the `time` benchmark type.  I'll have to go back to the drawing board on this one, but this PR is probably not the solution, particularly because it adds some user-facing complexity which is not all that great.

cdeil:
Right ... I would like to run the benchmarks every few months before making a release or after re-writing some internals in C or Cython to see if the analyses end users care about get faster or slower over time, so I'm still interested in the capability of asv to run and compare different machines, Python versions and commits.

Issue #183, opened 2015-01-08
Support snakeviz


cdeil:
It would be great if `asv` supported https://github.com/jiffyclub/snakeviz in addition to `RunSnakeRun`,
because it's easy to install anywhere (e.g. on Mac).

Long-term using a web-profile-viewer might allow interactivity by making a webapp: click on one or two commits, click "profile" ... asv goes off and executes the commands and opens a new browser window with the results.

cc @jiffyclub

mdboom:
Yes, as a first pass, we could support snakeviz as just another profiling GUI alongside the desktop ones `asv` already supports.

The long term plan is to save the profiling data along with the benchmarking results and be able to get to them directly from the `asv` ui.  There's a couple of obstacles to this: Python cProfile data doesn't contain the source code -- only hardcoded paths to the source code -- so this would have to be converted to something static that did contain the source code so that sharing profiling data across machines/users made sense.  Secondly, `snakeviz` probably isn't the appropriate tool for this as it stands now, since it includes a server-side component.  It may be not be a huge step to modify it to run off of static data, however -- that would be ideal, then it just becomes another part of the static blob that gets placed on a static web server.

cdeil:
Or `asv` grows an optional server-side components in addition to it's current capability of creating static webpages. IMO having the ability to browse benchmarks and then to profile and try to understand the "interesting" features they show easily would be super nice.

jiffyclub:
SnakeViz is fairly amenable to this since it's turning into a static, single-page application. The server does do some pre-processing of the stats, but once it's done it renders/delivers the HTML and then the server's job is over (except to deliver static assets). If you can generate and store that rendered HTML you're on your way (I haven't looked into how to do this). You'd also have to make sure SnakeViz's static assets were on the server.

mdboom:
Yes, but the fact that `asv` has no server-side component is the whole reason I started the project, since none of the available alternatives work that way.  I think a lot would be lost if we moved away from that.  I'm not sure that snakeviz really needs the server side component, except that it was a convenient way to develop it.

mdboom:
@jiffyclub: I think our comments just crossed lines.  That sounds promising that there's a possible way to do this statically.

astafan8:
snakeviz is supported so thia issue can be closed ;)

PR #184, opened 2015-01-08
Support snakeviz

mdboom:
Implements half of what is suggested in #183.  This adds `snakeviz` as a new profiling gui type.

Also fixes a couple of buglets with the profiling command.

@cdeil, @jiffyclub

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -316,7 +316,7 @@ class Benchmark(object):
                 code = method_caller.__code__
 
             profile.runctx(
-                code, {'run': self.run}, {}, filename)
+                code, {'run': self.func}, {}, filename)
 
 
 class TimeBenchmark(Benchmark):
--- a/asv/commands/profiling.py
+++ b/asv/commands/profiling.py
@@ -44,13 +44,13 @@ class Profile(Command):
             description="Profile a benchmark")
 
         parser.add_argument(
-            'benchmark', nargs=1,
+            'benchmark', nargs='?',
             help="""The benchmark to profile.  Must be a
             fully-specified benchmark name.""")
         parser.add_argument(
             'revision', nargs='?',
             help="""The revision of the project to profile.  May be a
-            commit hash, or a tag or brach name.""")
+            commit hash, or a tag or branch name.""")
         parser.add_argument(
             '--gui', '-g', nargs='?',
             help="""Display the profile in the given gui.  Use
@@ -102,7 +102,7 @@ class Profile(Command):
     @classmethod
     def run_from_conf_args(cls, conf, args):
         return cls.run(
-            conf=conf, benchmark=args.benchmark[0], revision=args.revision,
+            conf=conf, benchmark=args.benchmark, revision=args.revision,
             gui=args.gui, output=args.output, force=args.force,
             environment=args.environment, python=args.python)
 
@@ -122,6 +122,10 @@ class Profile(Command):
         if gui is not None and gui not in cls.guis:
             raise util.UserError(
                 "Unknown profiler GUI {0}".format(gui))
+
+        if benchmark is None:
+            raise util.UserError(
+                "Must specify benchmark to run")
 
         repo = get_repo(conf)
 
--- a/asv/commands/profiling.py
+++ b/asv/commands/profiling.py
@@ -8,6 +8,7 @@ import contextlib
 import io
 import os
 import pstats
+import sys
 import tempfile
 
 from . import Command
@@ -185,7 +186,12 @@ class Profile(Command):
                     raise util.UserError(
                         "Environment {0} not found.".format(environment))
 
-            benchmarks = Benchmarks(conf, regex=benchmark)
+            if env.python != "{0}.{1}".format(*sys.version_info[:2]):
+                raise util.UserError(
+                    "Profiles must be run in the same version of Python as the "
+                    "asv master process")
+
+            benchmarks = Benchmarks(conf, regex='^{0}$'.format(benchmark))
             if len(benchmarks) != 1:
                 raise util.UserError(
                     "Could not find benchmark {0}".format(benchmark))
--- a/asv/plugins/kcachegrind.py
+++ b/asv/plugins/kcachegrind.py
@@ -23,4 +23,5 @@ class KCachegrindGui(profiling.ProfilerGui):
     def open_profiler_gui(cls, profiler_file):
         command = util.which("pyprof2calltree")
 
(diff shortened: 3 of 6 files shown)


cdeil:
The first profile command I tried was the one from the docs ... it fails with a "bad marshal data" error because it's trying to open a Python 2.7 profile with Python 3.4:
https://gist.github.com/cdeil/a90ea9106fc06ec9e7e9

cdeil:
Now I'm getting this `KeyError`:
```
eduroam-3-163:gammapy-benchmarks deil$ asv profile time_FluxDistribution v0.1 --python 3.4 
· Discovering benchmarks
·· Creating virtualenv for py3.4-astropy-numpy-scipy..
·· Uninstalling gammapy from py3.4-astropy-numpy-scipy.
·· Installing gammapy into py3.4-astropy-numpy-scipy....
· Profile data does not already exist. Running profiler now.
·· Creating virtualenv for py3.4-astropy-numpy-scipy..
·· Uninstalling gammapy from py3.4-astropy-numpy-scipy.
·· Installing gammapy into py3.4-astropy-numpy-scipy...
·· Benchmarking py3.4-astropy-numpy-scipy
··· Running time_catalog.time_FluxDistribution                                                                                             65.78msTraceback (most recent call last):
  File "/Users/deil/Library/Python/3.4/bin/asv", line 9, in <module>
    load_entry_point('asv==0.1', 'console_scripts', 'asv')()
  File "/Users/deil/Library/Python/3.4/lib/python/site-packages/asv-0.1-py3.4.egg/asv/main.py", line 36, in main
    result = args.func(args)
  File "/Users/deil/Library/Python/3.4/lib/python/site-packages/asv-0.1-py3.4.egg/asv/commands/__init__.py", line 45, in run_from_args
    return cls.run_from_conf_args(conf, args)
  File "/Users/deil/Library/Python/3.4/lib/python/site-packages/asv-0.1-py3.4.egg/asv/commands/profiling.py", line 107, in run_from_conf_args
    environment=args.environment, python=args.python)
  File "/Users/deil/Library/Python/3.4/lib/python/site-packages/asv-0.1-py3.4.egg/asv/commands/profiling.py", line 206, in run
    profile_data = results[benchmark]['profile']
KeyError: 'time_FluxDistribution'
```

mdboom:
> The first profile command I tried was the one from the docs ... it fails with a "bad marshal data" error because it's trying to open a Python 2.7 profile with Python 3.4.

Not sure how to fix that one, but we should document it.  I haven't run into this because I've been doing profiling in the same version of Python that I have runsnake and snakeviz installed for.  I suppose what we have to do is save it in some cross-Python compatible way (if that's even easy to do) and then convert it back to the correct version before passing along to the GUI tool (which means we have to know what version of Python it expects etc.).  Ugh.

> Now I'm getting this KeyError:

For some reason the profiling is failing, but the traceback is not being displayed.  I'll have to poke at this a bit to see if I can reproduce it.

cdeil:
For the Py 2/3 profile issue ... how about catching the error, printing a warning users will understand and go on ... in this case it would continue to do the profiling for Python 3 and display it like I want, no?

mdboom:
There's two issues here (and I was confused about which one you were running into):

- The GUI we load the profiling in has to be the same version of Python that we ran the profiler in.  This one is hard to catch for, since it's difficult to know what version of Python the external GUI tool will run in.

- When the profiling results are just printed as a table to the console, we also have to make sure that the `asv` master process is the same version of Python as the slave benchmarking process.  This is at least something we can catch and control.

mdboom:
This has been updated so:

- It will raise an exception when the master and slave Python versions don't match.  We have to assume that the Python versions of `asv` and the GUI tool do match.

- You get a proper error message when the benchmark name is not fully specified (as the docstring says).

mdboom:
> Now I'm getting this KeyError:

This is because the benchmark name wasn't being fully specified (it's ambiguous without a full path).  This has been fixed so one gets an error message and not a traceback.

cdeil:
This basically works ... here's some minor comments ...

Exiting `snakeviz` by closing the web browser tab results in this error:
```
· Error running /Users/deil/Library/Python/3.4/bin/snakeviz /var/folders/sb/4qv5j4m90pz1rw7m70rj1b1r0000gn/T/tmp7mp9h82s
            STDOUT -------->
            
            STDERR -------->
            
  File "/Users/deil/Library/Python/3.4/bin/asv", line 9, in <module>
    load_entry_point('asv==0.1', 'console_scripts', 'asv')()
  File "/Users/deil/code/asv/asv/main.py", line 36, in main
    result = args.func(args)
  File "/Users/deil/code/asv/asv/commands/__init__.py", line 45, in run_from_args
    return cls.run_from_conf_args(conf, args)
  File "/Users/deil/code/asv/asv/commands/profiling.py", line 108, in run_from_conf_args
    environment=args.environment, python=args.python)
  File "/Users/deil/code/asv/asv/commands/profiling.py", line 217, in run
    return cls.guis[gui].open_profiler_gui(profile_path)
  File "/Users/deil/code/asv/asv/plugins/snakeviz.py", line 26, in open_profiler_gui
    return util.check_call([command, profiler_file])
  File "/Users/deil/code/asv/asv/util.py", line 219, in check_call
    dots=dots, display_error=display_error, shell=shell, env=env)
  File "/Users/deil/code/asv/asv/util.py", line 330, in check_output
    raise ProcessError(args, retcode, stdout, stderr)
asv.util.ProcessError: Command '/Users/deil/Library/Python/3.4/bin/snakeviz /var/folders/sb/4qv5j4m90pz1rw7m70rj1b1r0000gn/T/tmp7mp9h82s' returned non-zero exit status -15
```
Maybe special-case this exit code as normal and don't raise this error?

When I select a benchmark that doesn't exist I get the correct message, but should `asv` create a virtualenv and uninstall / install the project in this case?
```
$ asv profile me_no_exit v0.1 --python 3.4 -g snakeviz
· Discovering benchmarks
·· Creating virtualenv for py3.4-astropy-numpy-scipy..
·· Uninstalling gammapy from py3.4-astropy-numpy-scipy.
·· Installing gammapy into py3.4-astropy-numpy-scipy....
· Could not find benchmark me_no_exit
```

mdboom:
> Maybe special-case this exit code as normal and don't raise this error?

That makes sense.

> When I select a benchmark that doesn't exist I get the correct message, but should asv create a virtualenv and uninstall / install the project in this case?

Yes -- `asv` needs to install the benchmarked project into an environment in order to discover the benchmarks that are available.  So you can't determine if a benchmark exists until you've done that.

PR #188, opened 2015-01-08
Incremental benchmarking (take 2)

mdboom:
This is a follow-on to #177, preserving @rhattersley's original work, that have separate `save` and `update_save` methods so that each can be used in the correct context.


rhattersley:
Nice. Thanks @mdboom :+1:

rhattersley:
\o/

PR #193, opened 2015-01-17
Add mercurial support

MatthewTurk:
This is a WIP PR for adding mercurial support, including refactoring out a little bit of the git-specific logic.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,6 @@ install:
 
 script:
     - pip install virtualenv==1.10
-    - pip install "mercurial>=3.3"
-    - pip install python-hglib==1.5
+    - if [[ $TRAVIS_PYTHON_VERSION == 2* ]]; then pip install "mercurial>=3.3" ; fi
+    - if [[ $TRAVIS_PYTHON_VERSION == 2* ]]; then pip install python-hglib==1.5 ; fi
     - python setup.py test
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,4 +31,6 @@ install:
 
 script:
     - pip install virtualenv==1.10
+    - pip install "mercurial>=3.3"
+    - pip install python-hglib==1.5
     - python setup.py test
--- a/asv/commands/run.py
+++ b/asv/commands/run.py
@@ -156,8 +156,7 @@ class Run(Command):
         elif range_spec == 'NEW':
             latest_result = find_latest_result_hash(
                 machine_params.machine, conf.results_dir)
-            # TODO: This is shamelessly git-specific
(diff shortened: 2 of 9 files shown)


mdboom:
Thanks for this.  Now the "pluggable repo" code is no longer theoretical.  Let me know how I can be helpful to get this over the finish line.

It should be possibly to add some testing for this by adding a new test ala `test_workflow` that points to some random `hg` repo rather than a `git` one (the current one points to `asv` itself on github).

MatthewTurk:
Thanks, @mdboom !  I will add some tests and update the docs and it should be ready.  Right now we're getting set up with asv on our repo so hopefully that will shake out any potential other bugs.

mdboom:
Just a heads up: #199 changes the Repo plugin API very slightly.  See #199's diff of `git.py` for the analogous changes that will be needed here.

mdboom:
Can you add a line to install `hglib` in `.travis.yml` so this gets tested on Travis?

MatthewTurk:
@mdboom if you could take a quick look, I think it's just about ready?

mdboom:
#203 changes the the Repo API slightly again.  I'll resolve those differences in some method or other, just want to notate that.

.travis.yml
```diff
@@ -31,4 +31,6 @@ install:
 
 script:
     - pip install virtualenv==1.10
+    - if [[ $TRAVIS_PYTHON_VERSION == 2* ]]; then pip install "mercurial>=3.3" ; fi
+    - if [[ $TRAVIS_PYTHON_VERSION == 2* ]]; then pip install python-hglib==1.5 ; fi
```
mdboom: Does this indicate that the Mercurial support is only on Python 2.x, since `python-hglib` only works there?  Maybe that should be documented?

MatthewTurk: I will add some notes -- `python-hglib` is being ported to Python 3, but isn't ready yet.  Mercurial is a longer term port; I wasn't sure how to indicate that Mercurial would need to come from system packages if it's a Python 3 test.

mdboom: > I wasn't sure how to indicate that Mercurial would need to come from system packages if it's a Python 3 test.

I think `apt-get install mercurial` should work.  But if `python-hglib` doesn't work on Python 3 yet, the mercurial support here wouldn't work anyway, right?  So perfectly fine to leave as-is for now until the Python 3 support is ready.

MatthewTurk: (Of course that makes sense about `apt-get` ... :open_mouth: )  I will absolutely issue a PR when `python-hglib` is ported.  Thanks!

mdboom:
Merged via #214.

Issue #194, opened 2015-01-19
First impressions porting the pandas repository


qwhelan:
@mdboom First of all, it's really impressive to see a v0.1 with this level of polish in presentation, especially given how even mature open source projects tend to be lacking in this regard.

Due to https://github.com/pydata/pandas/issues/8361, I decided to take a look at converting pandas to asv from https://github.com/pydata/vbench and ran into a few usability issues. Most of the following comments can be standalone issues, but I wanted to put everything in context first. I'm also willing to help fix many of these but I'd like feedback from you first.

The first snag was the relative location of ```asv.conf.json``` with regard to the git repo. I thought it would be something like this:
```
pandas
├── asv.conf.json
├── pandas
└── setup.py
```
But it instead seems to expect this:
```
asv.conf.json
pandas
├── pandas
└── setup.py
```
Which poses some issues with distributing it with the repo, as well as conflicts between multiple projects in the same directory. The former fails due to the assumption that ```./project``` is the project root and thus that ```project/setup.py``` exists. Due to my moving it down a level, ```asv.environment.install()``` tries to install from ```pandas/pandas``` and fails. Many of the following issues result from my patching ```asv``` locally to support the former approach, so please let me know if trying to support this is completely unreasonable.

A few other things I noticed:
* ```asv``` doesn't create a secondary repo to run benchmarks in. Again this is only an issue due to my desire to distribute ```asv.conf.json``` with the repo and this is the direct root cause of many of the following issues.
* ```asv run``` does a ```git checkout -f master``` which discards any uncommitted changes *and* committed changes on a detached HEAD (more on why this is a concern later)
 * This bit me several times while trying to just get ```asv run``` to work - I can guarantee this will cause a developer with a substantial change to lose their work.
 * Probably best to remove the ```-f``` and let git yell at the user
* ```asv run``` does a ```git clean -fxd``` which forcibly discards any untracked files.
 * I realize the need to get into a consistent state, but the lack of a secondary repo means that works-in-progress or personal utility scripts will get nuked.
 * It took a couple tries to realize this was why ```asv.conf.json```, ```benchmarks/``` and ```env/``` were being deleted after every ```asv run```
 * This also nukes the ```envs/``` directory, but that's consistent with everything else
* ```asv run``` doesn't return the repo to the state it was in prior to the run
 * Before: ```On branch master```
 * After: ```HEAD detached at 2ef4ba0```
[...]

PR #195, opened 2015-01-19
Fix example asv configuration in test/

qwhelan:
@mdboom This fixes a couple problems noted in #194 and allows ```cd test/; asv run``` to successfully complete.


mdboom:
Thanks.

PR #198, opened 2015-01-26
Fix #197: Make asv dev work again

### LINKED ISSUE #197: `asv dev` fails to find executable
As mentioned in #194, when running ```asv dev```, I get the following error:
```
$ asv dev
· No executable found for python same
· Creating environments
· Installing dependencies
· No environments selected
```

The root cause is an inconsistency in how ```Environment``` and its subclasses define the ```python``` argument. ```ExistingEnvironment``` expects ```python``` to be a full name (such as ```python2.7```) while ```VirtualEnv``` and ```Conda``` expect ```python``` to be just a version number (specifically in ```get_environments()``` and ```matches()```). These latter two classes also do not handle the ```python='same'``` special case. This leads to a ```util.which('pythonsame')``` call which fails and returns no environment matches.

I didn't submit a patch as it's unclear what the preferred fix is - ```same``` could be handled individually in each of the plugins (hacky but possibly easier) or it could be replaced with ```None``` inside ```environment.py``` and each subclass would just have to handle the no-version-number case.

mdboom:
This should have been part of #180.

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -99,6 +99,9 @@ def get_environment_class(conf, python):
           that all dependencies and the benchmarked project itself are
           already installed.
     """
+    if python == 'same':
+        conf.environment_type = 'existing'
+
     if conf.environment_type:
         for cls in util.iter_subclasses(Environment):
             if cls.tool_name == conf.environment_type:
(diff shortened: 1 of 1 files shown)


qwhelan:
@mdboom This works for me.

Issue #201, opened 2015-01-29
pip confuses ccache


pv:
Just a note which may be useful for people using ccache:

Using virtualenv+pip causes stuff to be built in directories whose names have random components.
As a consequence, ccache seems to produce mostly cache misses when compiling.

A possible workaround is here: https://github.com/pv/scipy-bench/blob/master/run.py#L29
```
# Required to make ccache work properly --- pip builds and installs stuff
# to paths with random components in directory names, which confuses ccache.
#
# To work around this, we need to encourage ccache to be sloppy, and to
# remove the -g flag from compiler flags --- it causes file names to be
# embedded to the object files, which ccache apparently does not like.

import sysconfig
os.environ['CCACHE_SLOPPINESS'] = 'file_macro,time_macros'
os.environ['CCACHE_UNIFY'] = '1'
os.environ['CFLAGS'] = drop_g_flag(sysconfig.get_config_var('CFLAGS'))
os.environ['OPT'] = drop_g_flag(sysconfig.get_config_var('OPT'))
os.environ['LDSHARED'] = drop_g_flag(sysconfig.get_config_var('LDSHARED'))

def drop_g_flag(flags):
    """
    Drop -g from command line flags
    """
    if not flags:
        return flags
    return " ".join(x for x in flags.split() if x not in ('-g', '-g1', '-g2', '-g3'))
```
Not sure whether all of the above are needed, but it appeared that at least UNIFY and dropping `-g` were necessary to make ccache work with Scipy/Numpy. Files produced by Cython however still seem to produce cache misses --- didn't manage to figure that out yet.

Making `asv` define the above seems very dirty, but mentioning the pitfall and workaround in docs might be useful.

mdboom:
Thanks.  An earlier version of `asv` used `pip install --editable` which builds things in the same place everytime.  But that had other issues.  I'll play with this suggestion when I get a chance.

mdboom:
Fixed by #203.

PR #202, opened 2015-01-29
Add ccache plugin

mdboom:
This is an attempt to address #201.

This adds an optional `ccache` plugin that can be turned on by adding `"plugins": ["asv.ext.ccache"]` to `asv.conf.json`.  That way we don't need to add this hacky stuff for everyone, just for those who opt-in.

Unfortunately, however, it doesn't seem to work, and I'm getting all cache misses when enabling these environment variables (even though I've confirmed they are being applied).  @pv since you reported the technique itself as working, do you have any thoughts?

### DIFF
--- a/asv/ext/__init__.py
+++ b/asv/ext/__init__.py
@@ -1,0 +1,3 @@
+"""
+This directory contains plugins that are not enabled by default.
+"""
--- a/asv/ext/ccache.py
+++ b/asv/ext/ccache.py
@@ -1,0 +1,31 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import, division, unicode_literals, print_function
+
+"""
+This plugin makes ccache more effective by making it sloppier and
+removing debugging information from the generated files.
+"""
+
+# Contributed by @pv
+
+import os
+import sysconfig
+
+
+def drop_g_flag(flags):
+    """
+    Drop -g from command line flags
+    """
+    if not flags:
+        return flags
+    return " ".join(
+        x for x in flags.split() if x not in ('-g', '-g1', '-g2', '-g3'))
+
+
+os.environ['CCACHE_SLOPPINESS'] = 'file_macro,time_macros'
+os.environ['CCACHE_UNIFY'] = '1'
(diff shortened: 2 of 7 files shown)


mdboom:
And a couple of notes about details I didn't bother with until the basics are worked out:

- [ ] `sysconfig` isn't available on Python 2.6.  Need to resolve that.
- [ ] We should be using the `sysconfig` of the Python we are about to do the compiling for, not of the master `asv` Python.

pv:
I double checked now that I get cache hits. I have ccache version 3.2.1 --- the 3.1.10 has some bugfix that may or may not be relevant. Also, I still get cache misses for Cython-generated files, even with these variables set. Apart from these, nothing immediately comes to mind.

mdboom:
I tried ccache 3.1.10 and 3.2.1, which gcc 4.9.2 underneath on Fedora Linux 21.  Still the same -- no cache hits at all.  I'll continue to investigate.  Just to confirm: Did you try the plugin as implemented here or enabled somehow differently?  (Just to rule out this vs. something environmental about my ccache setup).

pv:
I did not yet try this plugin, but instead ran the run.py in https://github.com/pv/scipy-bench --- but that only sets the environment variables.

mdboom:
I still get only misses using the `run.py` script in https://github.com/pv/scipy-bench, so that suggests it's something with my ccache environment rather that just how I've implemented things here, I think.

pv:
I get cache hits (preprocessed) with this plugin. The reason why Cython files cause misses is that Cython generates variables that contain source file paths.

mdboom:
Ah -- I see the issue.  Scipy is using Numpy distutils (if I'm not mistaken) which takes `CFLAGS` as an override, not an extension of, the built-in Python compiler flags.  I was testing with astropy, which uses vanilla distutils which only lets you add flags, thus the `-g` flag was still being passed.

pv:
Ok, indeed. I don't know if it's actually possible to remove the `-g` when using the usual distutils. This hack is then not generally so useful... One option that according to ccache man pages would tolerate `-g` is to somehow force pip to build in a known directory (not sure if this is possible), and set `CCACHE_BASEDIR` accordingly.

mdboom:
I suspect #203 (or something like it) will be a better ultimate solution and less full of black compiler-specific magic.

mdboom:
Closing in favor of #203.

PR #203, opened 2015-01-30
Use `setup.py` directly rather than `pip install`

mdboom:
This results in building the project in the same directory every time, giving ccache a better chance of doing its thing.  An alternate solution to #201.

`pip install /some/local/path` has been used in asv to install the benchmarked project on the assumption that `pip` knows better about the details of how to install a Python package beyond just the usual `setup.py install`.  However, I wonder if that's really the case...

Note that I tried to use the `--build` flag for `pip install` to force it to build in a specific directory.  This seems to work for remote packages, but it's ignored when given a local path...

Attn: @pv

### DIFF
--- a/asv/console.py
+++ b/asv/console.py
@@ -308,8 +308,10 @@ class Log(object):
     def set_level(self, level):
         orig_level = self._logger.level
         self._logger.setLevel(level)
-        yield
-        self._logger.setLevel(orig_level)
+        try:
+            yield
+        finally:
+            self._logger.setLevel(orig_level)
 
     def is_debug_enabled(self):
         return self._logger.getEffectiveLevel() <= logging.DEBUG
--- a/asv/console.py
+++ b/asv/console.py
@@ -304,6 +304,13 @@ class Log(object):
         else:
             self._logger.setLevel(logging.INFO)
 
+    @contextlib.contextmanager
+    def set_level(self, level):
+        orig_level = self._logger.level
+        self._logger.setLevel(level)
+        yield
+        self._logger.setLevel(orig_level)
+
     def is_debug_enabled(self):
         return self._logger.getEffectiveLevel() <= logging.DEBUG
 
--- a/asv/wheel_cache.py
+++ b/asv/wheel_cache.py
@@ -1,0 +1,95 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import, division, unicode_literals, print_function
+
+
+import os
+import shutil
+
+from . import util
+
+
+class WheelCache(object):
+    def __init__(self, conf, root):
+        self._root = root
+        self._path = os.path.join(root, 'wheels')
+        self._wheel_cache_size = getattr(conf, 'wheel_cache_size', 0)
+
+    def _get_wheel_cache_path(self, commit_hash):
+        """
+        Get the wheel cache path corresponding to a given commit hash.
+        """
+        return os.path.join(self._path, commit_hash)
+
+    def _create_wheel_cache_path(self, commit_hash):
+        """
+        Create the directory for the wheel cache corresponding to the given commit hash.
+        """
+        path = self._get_wheel_cache_path(commit_hash)
+        stamp = os.path.join(path, 'timestamp')
+        if not os.path.isdir(path):
+            os.makedirs(path)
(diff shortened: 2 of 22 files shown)


pv:
I think pip refuses to uninstall packages installed by `setup.py install` --- at least for the case when setup.py is not using setuptools.

Perhaps just running `python setup.py build` before `pip install .` will work?
At least for Numpy, this appears to skip the build in the second step.

pv:
The second question is then what to do with building dependencies --- if the environment in asv is not torn down every time, using ccache for those is probably not so important?

mdboom:
> I think pip refuses to uninstall packages installed by setup.py install --- at least for the case when setup.py is not using setuptools.

> Perhaps just running python setup.py build; pip install . will work (or just build_ext instead of build)?

At least experimentally on a dummy package that uses only vanilla distutils and no setuptools, this appears to be true.  `python setup.py install` makes something that isn't `pip uninstall`-able.  `python setup.py build` followed by `pip install ./project` works, though.  That seems like a good solution, then.

As for the dependencies, you are correct that they are still being built with random paths, therefore ccache doesn't work.  But they are cached in the `env` directory tree and don't need to be built each time, so it's less of an issue.  If it were an issue, we could pass `--build ./some-static-tmp-directory` to pip when building the dependencies, since `--build` *does* work when installing from downloaded packages, just not when installing from the local filesystem.

pv:
This works fine for me. Does it break the parallel build feature however?

mdboom:
I *does* break the parallel build feature.  For a couple of other corner case reasons, I've been meaning to fix the parallel build feature so that each parallel thread would use its own `git clone` anyway, so this might just be onus to do that...

mdboom:
This now makes a separate clone of the project repository for each environment (these are `--shared` clones, so their `.git` directories don't take up much space).  That way parallel building can work without clobbering another environment's build.  This also includes @pv's wheel caching support, so the last *N* builds are retained and can be reinstalled next time much faster.

The combination of parallel building, ccache working and cached wheels makes things *much* faster overall.  Too bad there isn't a tool to measure that ;)

pv:
Seems to break asv dev, cf pv@1addad0
`asv dev` also seems to unnecessarily clone the repo.

asv/wheel_cache.py
```diff
@@ -0,0 +1,92 @@
...
+        for name in names[self._wheel_cache_size:]:
+            path = os.path.join(self._path, name)
+            if os.path.isdir(path):
+                shutil.rmtree(path)
+
+    def build_project_cached(self, env, package, commit_hash):
+        if self._wheel_cache_size == 0:
+            return None
+
+        if not os.path.isdir(self._path):
```
pv: This check should be removed --- the directory is first created by the call to `_get_wheel` below

pv:
Worksforme apart from those two

mdboom:
Thanks.  I think I've addressed those two issues.

pv:
The isdir check probably needs to be added to `_get_wheel` pv@6762b7d ?

mdboom:
Indeed.  I think it's fixed now.

asv/wheel_cache.py
```diff
@@ -0,0 +1,92 @@
...
+        if not os.path.isdir(path):
+            os.makedirs(path)
+        with open(stamp, 'wb'):
+            pass
+        return path
+
+    def _get_wheel(self, commit_hash):
+        cache_path = self._get_wheel_cache_path(commit_hash)
+
+        if not os.isdir(cache_path):
```
pv: os.path.isdir

asv/console.py
```diff
@@ -304,6 +304,13 @@ def enable(self, verbose=False):
         else:
             self._logger.setLevel(logging.INFO)
 
+    @contextlib.contextmanager
+    def set_level(self, level):
+        orig_level = self._logger.level
+        self._logger.setLevel(level)
+        yield
+        self._logger.setLevel(orig_level)
```
pv: This doesn't set the level back if an exception is raised, may be related to gh-209

pv: pv@044b8d6f9c

mdboom: Thanks.  I've cherry-picked 044b8d6 over to this branch.

pv: I've been using this PR for the Scipy benchmarks, and didn't meet other
problems than what is fixed above.

Issue #36, opened 2013-12-12
Create way to parameterize benchmarks


mdboom:
It's probably handy to do:

```
def time_size(n):
    x = [0] * n
time_size.params = [10, 1000, 100000]
```

cdeil:
+1 ... this would be nice!

pv:
Just a heads up: some work towards this coming up here: https://github.com/pv/asv/tree/param-test
Running & js plot generation should work, x-axis and parameter set selection is still missing from UI but should be done soon-ish. Nevertheless, comments on [parameter specification API](https://github.com/pv/asv/blob/param-test/docs/source/writing_benchmarks.rst#parameterized-benchmarks) could be useful already at this stage  --- I'm just basing this on what would be useful for Scipy benchmarks, and experiences may vary...

pv:
This feature should be there now.

mdboom:
Fixed by #205.

PR #205, opened 2015-02-01
Implement parameterized benchmarks

pv:
Implement parameterization of benchmarks.

Plotting can deal with numerical and categorical parameters, user can select in the UI parameter combinations to show and what x-axis to use.

Addresses gh-36.

Sample output here: https://pv.github.io/scipy-bench/

Some sample code here: https://github.com/pv/scipy-bench/tree/param/benchmarks/

TODO:

- [x] ensure asv compare and other cmdline commands deal with parameterized results
- [x] produce text output (tables) showing results, either always or add a command line switch
- [x] for many benchmark parameters, the selection buttons become too small; switch to vertical view (currently used for commit selection) when many choices
- [x] check javascript performance for large numbers of commits
- [x] it probably loads the data set many times from the server, could be refactored a bit if so
- [x] more careful type check for params and param_names on python side, the javascript code does currently not tolerate garbage input here (or, the javascript code should make these checks)
- [x] better unit testing
- [x] also `number` and `repeats` may need to be parameter-dependent?
- [x] should parameterized benchmarks be considered as separate benchmarks, or as benchmarks with multiple return values as currently? Is the parameterization API elegant and suitable?
- [x] should the API be such that first implement benchmarks with multiple return values, and then build parameterized benchmarks on top of that?

TODO for later (separate PR):

- what would be good summary plot? now it just averages over parameters (should these switch to geom means?)


asv/benchmark.py
```diff
@@ -316,7 +381,7 @@ def method_caller():
                 code = method_caller.__code__
 
             profile.runctx(
-                code, {'run': self.func}, {}, filename)
+                code, {'run': self.do_run}, {}, filename)
```
mdboom: This will add extra outer stack frames to the profiling output.  In general, we may need to think more carefully about how parameterized benchmarks work with profiling -- one profile per entry in the parameter matrix may make more sense.  But that's a detail for now.  Maybe as a stop gap we use `do_run` here if the benchmark is parameterized, otherwise use `func`.

mdboom:
This is highly cool!  Thanks for such an enormous contribution!

One structural thing strikes me right off the bat.  It seems that for each parameterized benchmark, all of its combinations are run in a single process.  This could be problematic.  If the library being benchmarked does memoization, the later-run benchmarks may be artificially advantaged.  Even without that, having a number of Python pools already allocated may impact the timings of later run benchmarks.  I think it's ultimately better and more reproducible to run each benchmark in its own process -- thereby moving the logic of running each parameterized benchmark from `benchmark.py` to `benchmarks.py`.

A minor detail: In the bar graph view, there are little circles at the top of each bar.  In the "traditional" asv line view, these link to commits.  In the bar graph view there's no such logical place to jump to, so maybe it's possible to just not show them (but that's just an aesthetic thing).

I agree with all of your TODO list items above, except for:

> produce text output (tables) showing results, either always or add a command line switch

I've had an unexpressed thought that a command to explore the data from the commandline would be useful.  Essentially a CLI replacement for the Javascript GUI, and it would be a good complement to `asv rm` which lets you modify the data but doesn't let you see it (which I've long thought was a problem).  That might be better than extending existing commands with this functionality.

> what would be good summary plot? now it just averages over parameters (should these switch to geom means?)

I wonder (not fully formed thought here) if that needs to be user-configurable.  For most timing things, I think people are more concerned with log(time) rather than time -- but not sure exactly how that impacts calculating the summary.

> Is the parameterization API elegant and suitable?

I think yes -- though it seems with many elegant designs there's always some use case someone will come up with that can't be expressed.  But I think this is a good starting point -- it's pretty much exactly what I had in mind, and the documentation of it is very straightforward (which is a good sign).

> should the API be such that first implement benchmarks with multiple return values, and then build parameterized benchmarks on top of that?

Given my concern about running multiple benchmarks in the same process (above), I'd say 'no'.

pv:
Ok, I see the issue with the single-process benchmarks. This should be straightforward to change by passing an integer index to subprocess. This also removes the usefulness of separate parameter-specific setup and teardown functions. The API could be simplified by removing `setup_params` and `teardown_params` and just making the basic `setup` and `teardown` parameterized.

There's an use case for printing the results directly --- when writing the benchmark, which maybe is done at the same time as writing the feature to be benchmarked, you'd like to run `asv dev --bench foo` and see all the results immediately, as you see for the non-parameterized ones, an preferably in an easy-to-read form. They cannot be inspected later on since they are not saved. I'd be tempted to just reuse `show_stderr` for this.

The benchmarks-with-multiple values is probably orthogonal question as far as running the benchmarks goes --- the result and display format used here however could be reusable also for such case. There's one benchmark in Scipy that would be most natural to be written as a parameterized, multi-value benchmark. But this feature can maybe be left for later.

- [ ] change benchmarks to run in separate processes

mdboom:
> There's an use case for printing the results directly

Ok.  That's a good point.  My suggestion is somewhat orthogonal then.

Thanks, again.  You've saved me a lot of work :wink:

pv:
As for JS performance, parameterized tests don't seem much slower than plotting other results. Profile says >~ 95% of the time is spent in $.plot, so it in the end depends on how many data points end up being handed to flot.

mdboom:
Good to know about JS performance.

Sorry to have comments trickling in here...  I'm discovering some corner details here as I play with this more.

It seems that if you generate some benchmark results with a particular set of parameters, and then later change those parameters, the old benchmark results are invalidated (in my testing, all results for that benchmark show up as `null` in the output graph json).  But it should be possible to map the old benchmarks into the new parameters wherever the parameter values match.  It seems that everything to make this work already exists in the `results` files, since the parameters used to get the results are duplicated there.  You could take the union of all parameters from all of the commits on a particular benchmark, and then map the existing results into that, leaving as `null` anything that isn't there.  I think this would fit into some other design decisions of `asv` that try to avoid obsoleting old benchmarks.

I also had a crazy idea to parameterize by types that turned out not to work.  It may be too problematic to make work, but I thought I'd share it and get your take.  In `astropy`, we have two different implementations of file readers, one implemented in Python, the other in C.  It would be handy to be able to parameterize based on the type of reader, something like:

```
def time_reader(reader_cls):
    reader_cls().read('somefile.txt')
time_reader.params = [SlowReader, FastReader]
```

Of course, this fails because the types are not JSON-serializable.  To a certain extent though, it may work well enough to serialize the `repr` of them and pass them back to the master asv process.  All it really needs the number of parameters and something unique to refer to them by anyway.  The objects themselves only ever have to exist in the `benchmark.py` subprocess.  Does that make sense?  This can always be a follow-on PR so as not to bog down this one.

pv:
Yes, it could be doable to take those parameters from old results that happen to match the current pattern, only small changes to publish.py required probably. The params are duplicated in the results files to make them more self-contained vs changes in the benchmark.
.
Taking an union of all existing parameters blows up the size of the cartesian product, so that alternative probably would need to be done on the javascript side.
I was thinking about it in the beginning, but it seemed complicated then (now, it seems somewhat less complicated :), and took the simpler route.
One problem is that if you mix strings and numbers, the x-axis can become categorical because of data in the past, even if the benchmark currently has only numerical parameters, so for starters (=this PR) it would be simpler to limit things to the parameters in the current version of the benchmark.
.
Restrictions on allowed parameter values help in the sense that you can write "some_benchmark(1, "foo")" and it's unambiguous from the textual repr  what benchmark it is, which will come up with asv compare et al. This was my experience with nose's parameterized tests, where you need restraint to not make a mess. Allowing classes could be nice, but I don't see clear places to draw the line except at pickle and json, and json seems to me the better option currently, although it requires some contortions in writing the benchmarks.
.
I agree it's best to think this through carefully now that we're at it and everything is fresh in mind, rather than to get something out quick...

mdboom:
>  so for starters (=this PR) it would be simpler to limit things to the parameters in the current version of the benchmark.

I see your point.  I think it will be more common that users *add* parameters than *replace* them, so it's fine to say that we display only the currently specified parameters, pulling any that match from the past, and leaving as "missing" any that were never run in the past.

> Allowing classes could be nice, but I don't see clear places to draw the line except at pickle and json,

I don't think pickling has to work because the object only has to exist in the benchmark process.  But I see your point, and of course there is a workaround to achieve what I describe, which is to use strings in the parameter list and then map those strings to classes in the benchmarked function.  Maybe the thing to do then is simply type check the parameter lists to make sure it's completely JSON-serializable and if not raise a helpful exception that only basic data types are supported.  And we could add an example to the docs of something like:

```
classes = {
    'slow': SlowReader,
    'fast': FastReader
}
def test_params(n):
    cls = classes[n]
    ...
test_params.params = list(classes.keys())
```

pv:
Ok, you're right that from the implementation POV, just repr() would be
fine, we can just special-case the numeric values in the visualization.
It can be done like that too, and then it's up to the users to do what
they think is sensible.

04.02.2015, 17:13, Michael Droettboom kirjoitti:

pv:
I think feature-wise this PR is more or less a complete unit now, unless I missed something.
I'll do some more testing, and probably add some further tests in the test suite.

TODOs for later:

- Dealing with only a single benchmark parameter combination: `asv run`, `asv find` et al.
- Separate profiles for each parameter combination? Now it just combines everything into one.
- The vertical layout of parameter selection buttons takes quite a lot of space. There might be prettier alternatives in bootstrap.
- Rename `show_stderr` to `show_output` since it actually includes also stdout now

pv:
I think this is ready now. Added some more tests, and fixed a number of remaining bugs. There are some additions that could still be done, but those are perhaps best to left for separate PRs.

The JS frontend doesn't have unit tests, but the workflow appears to work: https://pv.github.io/scipy-bench/

mdboom:
This is really cool.

I agree with the need for JS testing -- not sure how best to acheive that, because it's primarily GUI testing, but if you have any suggestions, let me know... (This is my first major web/js project...)

A couple of UI "surprises" I noticed (all of this is in Opera on Linux -- I can try other browsers if it turns out to be browser-specific):

If I choose the `io_matlab.MemUsage.track_savemat` benchmark from the summary page, it shows lines for all of the "size" params, but only the first value for the "compressed" param.  Is there a reason for the discrepancy, and is that documented somewhere.  I understand the need to not plot hundreds of lines at once, so maybe just choose the first of each parameter by default?  I also noticed that if I click on "size" for x-axis and then go back to "commit", only the first value for "size" (1M) is selected.

pv:
Hi, unfortunately I'm not any more familiar with JS testing solutions.
.
Those UI features were intentional although not perhaps completely
thought through; the initial set of curves plotted is defined on
asv.js:560, and turning off selectors when x-axis changes is at
asv.js:612. The motivation for both is to avoid inadvertently plotting a
large number of curves at once. I though it was useful to show more than
one line initially, so it just arbitrarily selects first four values of
the first parameter. The second point avoids suddenly increasing the
number of curves when the x-axis selection is changed.
.
The behavior here is easily changed, although I personally like the
default of initially showing more than one result.

mdboom:
@pv: I agree we don't want to plot many lines by default, but I don't have a great alternative suggestion.

*Perhaps* what we really want to do is turn on enough parameters so that the largest number of lines under *n* (where *n* is, say, 8) is displayed.  So in the case of 2x2 parameters, all would be displayed, in the case of 4x4, the all 4 of the first and 2 of the second would be displayed.  But maybe that's even more magical and surprising.

But I think the part I found most jarring, that we should try to fix if easy before merging, was clicking on "size" then on "commit" doesn't "undo" the last action like I was expecting.  Also note that clicking on "commpression" and then back to "commit" *does* seem to restore to the original state (choosing the first 4 options on the first parameter).  So maybe there is a bug in there somewhere given the inconsistency?

pv:
Allowing also some options from the second parameter is probably OK. I would probably prefer a greedy choice of picking as many of the first parameter as available, and only after that considering whether there's room for more.

Regarding your second issue: it's working as intended --- only the parameter selected for x-axis is reset. Whether what was intended is good behavior is then a different question. We can change x-axis change to not reset any parameters, in case that seems better (to try if you like it, comment out lines 612-620 in asv.js).

pv:
Changed to behave you suggested in e6c0e50

mdboom:
Thanks.  This looks good.

pv:
Thanks!

PR #206, opened 2015-02-01
Cache build results in Python wheels for faster installation

pv:
This may be useful when manually running `asv`. Quite often, you don't get the command line switches right the first time, and some benchmark fails e.g with timeout, so cutting down time spent waiting on rebuilds becomes useful. (For Scipy, the build time is 2 min (with ccache+f90cache) which is already annoying to wait for, whereas wheel install time is 2 seconds.)

Now, one question is how well `pip wheel` works with various Python projects. I'd believe it works almost for everything, but not 100% sure.

### DIFF
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -197,7 +197,7 @@ class Benchmarks(dict):
             repo = get_repo(conf)
             repo.checkout()
 
-            env.install_project(conf)
+            env.install_project(conf, repo.get_hash_from_head())
 
             output = env.run(
                 [BENCHMARK_RUN_SCRIPT, 'discover', root],
--- a/asv/commands/find.py
+++ b/asv/commands/find.py
@@ -127,7 +127,7 @@ class Find(Command):
                     conf.project, commit_hash[:8]))
 
             repo.checkout(commit_hash)
-            env.install_project(conf)
+            env.install_project(conf, commit_hash)
             x = benchmarks.run_benchmarks(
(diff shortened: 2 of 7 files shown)


mdboom:
This is cool.  A few questions:

1. Shouldn't this also be applicable to `conda`?  The `conda` environment installs dependencies through `conda`, but the project is still installed through `pip` as in `virtualenv`.  So maybe this should all go in the base class?

2. It probably makes sense to make this a configuration file option.  That way the number of wheels to save could be controlled, as well as turning it off in any situations where wheel-building may not work.

mdboom:
I've been thinking about this a little in the context of #203, where we need a way to fix parallel building.  There, I think we need a unique build location for each environment so they don't clobber each other.  Once doing that, that build environment can just be cached by just moving it somewhere.  So while I like the idea of caching builds, I'm not sure it needs the overhead of wheels to make happen.  Anyway, I'm going to spend some time kicking at this, but just wanted to give you a heads up that you may want to hit pause on this while I look into it, in case this gets done in a slightly different way.

mdboom:
I've incorporated this (with some reorganization) in #203.

pv:
Seems to break `asv dev` https://github.com/pv/asv/commit/1addad03d777359479260a93e6c65b9eb1dc9d5b

pv:
`asv dev` also clones the project, apparently unnecessarily

pv:
Sorry, commented on wrong issue (also closing this, since it's superceded).

PR #210, opened 2015-02-07
For python=same, do not clone repository or print no-op actions

pv:
The `asv dev` / `asv run --python=same` commands assume the project is already installed in the current overall Python environment. Therefore, we can skip all operations on the repository. Also, skip printing unnecessary things to stdout, and add a test checking `asv dev` works.

The use case I have in mind is to be able to use asv simply as a benchmark runner, without needing its repository/environment management features.

### DIFF
--- a/asv/commands/dev.py
+++ b/asv/commands/dev.py
@@ -46,6 +46,10 @@ class Dev(Run):
 
     @classmethod
     def run_from_conf_args(cls, conf, args):
-        return cls.run(
-            conf=conf, bench=args.bench, show_stderr=True, quick=True,
-            python=args.python, machine=args.machine, dry_run=True)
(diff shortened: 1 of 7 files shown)


mdboom:
This makes a lot of sense.  I wonder, however, if it might be better to have a new no-op subclass of `Repo` that goes into effect when `python=='same'` (much like `ExistingEnvironment` is a mostly no-op implementaton of `Environment`).  That would take all of these changes to not touch the repo out of the many places they are used and into the `repo.py` module.

pv:
Yes, that will probably cover the cases needed here, and more in case future commands will also need this.

pv:
Ok, this should be more sensible now.

mdboom:
Looks good.  Thanks!

PR #211, opened 2015-02-07
Add options to run command to run only missing benchmarks

pv:
Add options to `asv run` to skip benchmarks based on whether there are existing results or not.
This operates on a finer-grained level than the `MISSING` range_spec option.

I assume the intent in the hash comparison code in iter_results_for_machine_and_hash was to allow matching on the shortest initial part, and not require full match always?

### DIFF
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -422,7 +422,8 @@ class Benchmarks(dict):
 
         return cls(conf, benchmarks=d)
 
-    def run_benchmarks(self, env, show_stderr=False, quick=False, profile=False):
+    def run_benchmarks(self, env, show_stderr=False, quick=False, profile=False,
+                       skip=None):
         """
         Run all of the benchmarks in the given `Environment`.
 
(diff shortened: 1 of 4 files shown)


mdboom:
Nice.  Should we also remove the `MISSING` range_spec option, since it's now somewhat redundant.  `--skip-existing` and `ALL` is probably a much better choice (though slightly different because it will *fill in* any new benchmarks that were never run before).

pv:
I guess most of the use cases of `MISSING` are covered by `--skip-existing`. It probably is useful to offer skipping commits where there are already some test results present, but I think at least for a project with 10+ years of history having a separate flag combined with a commit range is more useful than `MISSING`. I didn't add it here since it already adds 3 flags for skipping, but maybe yet another one would be useful?

mdboom:
Yeah -- that makes sense.  We could have a `--skip-existing-commits`, and then remove the now-completely-redundant `MISSING`.

pv:
Rebased, removed `MISSING`, and added `--skip-existing-commits`.

mdboom:
Looks good!

PR #212, opened 2015-02-07
Run very slow timing benchmarks only once without repeats

pv:
The logic by which the timing benchmark choose the number and repeats is probably not very suitable for very slow benchmarks.

Users can in principle set number and repeats themselves to 1, but this is fiddly especially with parameterized benchmarks where the runtimes can vary orders of magnitude following the performance scaling of the algorithm.

This PR changes the logic as follows:

- result from calibration run is always retained
- very slow benchmarks are run only once, by default

The definition for "very slow" is here 5*goal_time, which for the default parameters is 10 seconds.
Presumably, most sources of fluctuations operate on shorter time scales, so that one run is usually good enough (although this point is debatable, maybe the default should still be twice or thrice?).

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -378,8 +378,7 @@ class TimeBenchmark(Benchmark):
         self._load_vars()
 
     def _load_vars(self):
-        self.repeat = _get_first_attr(
-            self._attr_sources, 'repeat', timeit.default_repeat)
+        self.repeat = _get_first_attr(self._attr_sources, 'repeat', 0)
         self.number = int(_get_first_attr(self._attr_sources, 'number', 0))
         self.goal_time = _get_first_attr(self._attr_sources, 'goal_time', 2.0)
         self.timer = _get_first_attr(self._attr_sources, 'timer', process_time)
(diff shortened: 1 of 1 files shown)


mdboom:
The current code here is borrowed whole-hog from IPython's `timeit` magic, for what it's worth.

My *gut* makes me think we should never do fewer than 2 runs, but I don't have any data to back that up -- it probably depends a lot on what the benchmark is actually doing.

I definitely like the idea of not throwing out the work done to perform the estimate.

pv:
The data set at https://pv.github.io/scipy-bench is generated with this patch in.
There *is* a lot of scatter in some of the slow benchmarks eg. https://pv.github.io/scipy-bench/#spatial.Neighbors.time_sparse_distance_matrix but I do not know if this is significantly reduced by increasing the minimum repeat count. If it is, then maybe waiting longer is worth it...

It's probably a good idea to check if the behavior changes with two or more repeats. For exponentially distributed error, you'd expect the std in minimum to scale as 1/repeat, but perhaps there is some hot vs. cold run effect that makes it decrease more.

mdboom:
I think I'm probably fine with merging this and tweaking later to do a minimum of 2 runs if necessary.  Would you mind rebasing?

pv:
Rebased

PR #215, opened 2015-02-11
CSS fix: prevent benchmark tree view carets from ending up at different lines than the labels

pv:
If benchmark names are long, the carets in the benchmark tree view in the web UI may end up at different lines than the labels:

tmp

Add a CSS fix to prevent this -- better to just let the text overflow.
Fixed example: https://pv.github.io/scipy-bench/#sparse_linalg_onenormest.BenchmarkOneNormEst.time_onenormest

### DIFF
--- a/asv/www/asv.css
+++ b/asv/www/asv.css
@@ -86,3 +86,11 @@
 .nav>li>a {
     padding: 0;
 }
+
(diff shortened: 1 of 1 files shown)


mdboom:
Seems simple enough.

PR #217, opened 2015-02-11
Small bugfix to support existing results without parameters

mdboom:
Fixes a small backward imcompatibility created by #205.


pv:
Looks good to me

mdboom:
Yeah -- it does go away when benchmarks.json is regenerated, but might as well be backward compatible anyway.

PR #219, opened 2015-02-11
Downsample summary graphs

mdboom:
The data set in `astropy-benchmarks` is getting quite large.  There doesn't seem to be any reason to store more points than can be seen and displayed for the summary "thumbnail" graphs.  This downsamples them if they get too big.

@pv: Does this have any negative repercussions for parameterized benchmarks?

### DIFF
--- a/asv/graph.py
+++ b/asv/graph.py
@@ -13,6 +13,14 @@ from six.moves import xrange
 from . import util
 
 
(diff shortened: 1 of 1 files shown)


pv:
Shouldn't cause problems, the summary graphs are the same.

Issue #208, opened 2015-02-07
Benchmark discovery project version suboptimal


pv:
The benchmark discovery seems to be run currently against whatever version was last active.
This is sometimes sub-optimal:
- Benchmarks may have imports written for the current project version, which fail for earlier version
- If the last version happens to result to build failure, all following asv operations will just fail

Perhaps best would be to give a config option that determines this, or just always check out master branch first.

mdboom:
This was done to be fast about discovering benchmarks, though with ccache and wheel cache support (almost) in place it's now less of an issue, and I think always checking out master will be the right thing in the majority of cases.

At the end of the day, though, benchmarks need to be written to be importable across all versions of the underlying project's API, and if there's a project commit in which the benchmarks fail to import, the benchmarks haven't been written correctly.  So we should probably:

- Handle ImportError during the importing of the benchmark suite when a benchmark is about to be run as a bigger error than just a "failed test" to flag that the benchmark suite probably needs some tweaking.

- Add a documentation note about that on the "writing benchmarks" page that the benchmark suite should be importable across API versions of the underlying project, and some strategies for that.

pv:
#568 should address this.

PR #221, opened 2015-02-12
Change default installed project version used in benchmark discovery

pv:
Choose either a most recently built version from the wheel cache, or if
none found, use the current head of the master branch.

This speeds up benchmark discovery avoiding rebuilding when possible,
and avoids build failures that occur if the project repo was previously
at an unbuildable commit. Instead, we assume now that the master branch
is at an installable state, and all wheels that build are OK.

If these assumptions break, then the failure mode is somewhat nasty for
the user, however.

Addresses (partially) gh-208

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -296,12 +296,17 @@ class Environment(object):
 
     def install_project(self, conf, commit_hash=None):
         """
-        Install a working copy of the benchmarked project into the
-        environment.  Uninstalls any installed copy of the project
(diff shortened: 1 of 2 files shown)


mdboom:
Hmm...  Not sure why Travis CI isn't reporting on this one.  Can you a trivial rebase and repush to see if that forces it?

pv:
Rebased, travis seems to have run this time.

Issue #220, opened 2015-02-11
Improve parameterized benchmark summary graphs


pv:
A reminder for myself: the parameterized benchmark summary graphs can be sometimes misleading, since they are calculated by averaging results over all benchmark parameters.

If it happens that data series for some parameters is only available starting from a given commit (e.g. new feature), this can causes a sudden jump in the average of the results. Various examples of this here: https://pv.github.io/scipy-bench/ for example in sparse.Getset.track_fancy_setitem --- there is no regression, the benchmark just does not run for all parameters for too old versions.

It's not immediately clear what is optimal here, but the situation probably can be improved.

pv:
This is fixed, geometric mean + filling should be OK.

PR #226, opened 2015-02-17
Improve summary plot for parameterized tests

pv:
Take geometric mean over parameter sets instead of arithmetic mean.
Also fill in data that is missing only for some parameter combinations,
which avoids spurious jumps in the summary graph.

Also do minor refactoring for Graph, ensuring self.data_points format is
the same for scalar and vector series.

The geometric mean is probably the best choice for summarizing all
the results in a single curve --- the main interest in the plot is to see
relative changes, and the geom mean is the choice that gives equal
weights on log-scale. Within the constraint of one curve, I think gh-220 is fixed.

before: http://pav.iki.fi/tmp/asv-226/html-before/index.html
after: http://pav.iki.fi/tmp/asv-226/html-after/index.html (note several spurious jumps removed in sparse&spatial tests)

### DIFF
--- a/asv/graph.py
+++ b/asv/graph.py
@@ -77,7 +77,7 @@ class Graph(object):
         self.summary = summary
         self.path = os.path.join(*parts)
         self.n_series = None
-        self.scalar_series = False
+        self.scalar_series = True
 
     def add_data_point(self, date, value):
         """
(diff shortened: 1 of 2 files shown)


asv/graph.py
```diff
@@ -176,6 +248,24 @@ def _mean_with_none(values):
...
+
+
+def _geom_mean_with_none(values):
+    """
+    Compute geometric mean, with the understanding that None and NaN
+    stand for missing data.
+    """
+    values = [x for x in values if x is not None and x == x]
+    if values:
+        exponent = 1/len(values)
```
mdboom: Did you mean integer truncation for Python 2.x here?

pv: There's `from __future__ import division` on top

mdboom: Oops.  Of course.

PR #227, opened 2015-02-18
Fix #224: Handle 0 values in log scale

### LINKED ISSUE #224: Log-scale in JS goes to infinite loop if zero data
The for loop at https://github.com/spacetelescope/asv/blob/master/asv/www/asv.js#L1069 that computes the tick labels for the log scale does not terminate if the raw data contains a zero, so that `Math.floor(Math.log(0) / Math.LN10) + 1 == -Infinity`.

E.g. go to https://pv.github.io/scipy-bench/#sparse.Getset.track_fancy_setitem and click "log scale"

mdboom:


mdboom:
@pv: If you confirm this works for you, I'll merge.  I would have liked to just skip values <= 0 when log scaling, but it seems the transform function in `flot` doesn't allow this.

pv:
Works for me

PR #228, opened 2015-02-18
Improve CSS

mdboom:
This hopefully improves the style somewhat, though I am no CSS expert.

Mainly, it:

- Removes the gradients and provides an overall flatter and more "raw Bootstrap" look
- Makes the carets "turn" when the layers of the tree view are opened and closed
- Improves the look of the summary page
- Fixes a bug where selection panels wouldn't switch to vertical when the text of each button was quite long
- Adds a scrollbar to selection panels only when necessary
- Makes the tooltip style consistent

screenshot

Cc: @pv  Any feedback?

### DIFF
--- a/asv/www/asv.css
+++ b/asv/www/asv.css
@@ -1,5 +1,7 @@
-.navbar-default .navbar-text {
-    color: #000;
+/* Basic navigation */
+
+.asv-navigation {
+    padding: 2px;
 }
 
 .navbar-default .navbar-link {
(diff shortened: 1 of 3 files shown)


pv:
Looks OK to me. One comment: the almost white f5f5f5 background color of the sidebar headers makes them not stand out well, combined with the very dark selected color. Darker header color (eg ddd) and lighter selection color could be easier on the eyes.

mdboom:
Ok -- I've toned down the contrast between those colors, and indeed I think it looks better now.

PR #229, opened 2015-02-21
Fix printing cloning and building log messages in wrong order 

pv:
Cosmetic fix: the "Cloning..." message was printed after "Building...", which made it look as if cloning took a long time when in reality it was building at the time.

Print the build message immediately before the build command, in case
checkout() also prints something.  Do not print "Cloning..." message
when cloning the repositories from the local mirror repo into the
environments, as this action is likely to be fast.


mdboom:
Sure.  Seems fine to me.

PR #231, opened 2015-02-22
Make difference between failed and skipped benchmarks

pv:
Represent explicitly skipped benchmarks with NaN values, and failed benchmarks with None/null.

The practical difference here is that later on, you might want to use `--skip-existing-successful` to re-run failed tests e.g. due to fixes in the benchmark, but not re-run tests that you explicitly skipped in the benchmark suite.


mdboom:
Thanks.

PR #232, opened 2015-02-26
Add support for branches

pv:
This adds support for branches:

- config option for specifying the branches, defaults to only the 'master' branch
- separate lines in graphs for each branch, implemented via the generic machine parameter set thing (no JS changes needed, nice design!)
- run NEW and run ALL run benchmarks for all specified branches. The `--steps` option applies to each branch separately
- branch information is not stored in asv result files, as it can be obtained from the repository when needed (for subversion this can be cached)

This is still a bit work in progress:

- [x] some more manual testing is needed
- [x] update docs and sample config

### DIFF
--- a/asv/commands/publish.py
+++ b/asv/commands/publish.py
@@ -140,9 +140,6 @@ class Publish(Command):
 
             branch_cache = BranchCache(conf, repo)
 
-            for branch in conf.branches:
-                params.setdefault('branch', set()).add(branch)
-
         log.step()
         log.info("Loading results")
         with log.indent():
@@ -195,6 +192,7 @@ class Publish(Command):
             val = list(val)
             val.sort(key=lambda x: x or '')
             params[key] = val
+        params['branch'] = conf.branches  # maintain same order as in conf file
         util.write_json(os.path.join(conf.html_dir, "index.json"), {
             'project': conf.project,
             'project_url': conf.project_url,
--- a/asv/commands/publish.py
+++ b/asv/commands/publish.py
@@ -18,6 +18,7 @@ from ..graph import Graph
 from ..machine import iter_machine_files
 from ..repo import get_repo
 from ..results import iter_results
+from ..branch_cache import BranchCache
 from .. import util
 
 
@@ -128,6 +129,21 @@ class Publish(Command):
                 machines[d['machine']] = d
 
         log.step()
+        log.info("Getting tags and branches")
+        with log.indent():
+            repo = get_repo(conf)
+            repo.pull()
+            tags = {}
+            for tag in repo.get_tags():
+                log.dot()
+                tags[tag] = repo.get_date_from_name(tag)
+
+            branch_cache = BranchCache(conf, repo)
+
+            for branch in conf.branches:
+                params.setdefault('branch', set()).add(branch)
+
+        log.step()
         log.info("Loading results")
         with log.indent():
             for results in iter_results(conf.results_dir):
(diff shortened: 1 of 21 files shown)


asv/commands/publish.py
```diff
@@ -128,6 +129,28 @@ def run(cls, conf):
...
+            repo = get_repo(conf)
+            repo.pull()
+            tags = {}
+            for tag in repo.get_tags():
+                log.dot()
+                tags[tag] = repo.get_date_from_name(tag)
+
+            branch_cache = BranchCache(conf, repo)
+
+            def iter_branches(results):
```
mdboom: It's a bit confusing to have this function defined here but used a number of lines below.

asv/commands/run.py
```diff
@@ -42,6 +43,21 @@ def _do_build_multiprocess(args):
...
+    """
+    if not n > 0:
+        return items
+    spacing = max(float(len(items)) / n, 1)
+    spaced = []
+    i = 0
+    while int(i) < len(items) and len(spaced) < n:
+        spaced.append(items[int(i)])
+        i += spacing
+    return spaced
```
mdboom: Might as well throw this in `util`, since it's so general.

pv:
Ok, fixed up a bit.

asv/commands/run.py
```diff
@@ -151,31 +152,40 @@ def run(cls, conf, range_spec=None, steps=0, bench=None, parallel=1,
...
+                new_hashes = repo.get_hashes_from_range(spec)
+                branch_commit_hashes.append(util.pick_n(new_hashes, steps))
+
+            # Splice hashes from branches
+            seen = set()
+            commit_hashes = []
+            for commit_hash in util.splice(*branch_commit_hashes):
+                if commit_hash not in seen:
+                    seen.add(commit_hash)
+                    commit_hashes.append(commit_hash)
```
mdboom: It looks like `steps` no longer applies when `NEW` or `ALL` is given...  I think steps should always apply (and probably just be handled below as it was before, rather than in select branches of the `if` statement here.

asv/template/asv.conf.json
```diff
@@ -13,6 +13,9 @@
     // project being benchmarked
     "repo": "",
 
+    // List of branches to benchmark. If not provided, defaults to "master".
+    // "branches": ["master"],
```
mdboom: Maybe somehow indicate this should be `tip` for Mercurial?

asv/commands/run.py
pv: It applies (see the call to pick_n above), however, it's per-branch.
However, it may be less confusing if it's applied globally.

asv/template/asv.conf.json
pv: It probably should be `default` for mercurial -- `tip` is something more similar to `HEAD` as it always points to the most recent commit.

asv/commands/run.py
mdboom: Sorry -- I missed the call to pick_n in this branch.  I'm a little surprised that `--steps` applies per-branch, though.  
I think `--steps` should be global (though, ideally, would distribute its picks evenly between branches).

pv:
Updated. `--steps` is now global. The straightforward implementation here pick commits from branches in proportion to their total commit counts, which is probably sensible.

asv/commands/run.py
```diff
@@ -52,6 +54,14 @@ def setup_arguments(cls, subparsers):
...
             description="Run a benchmark suite.")
 
+        def positive_int(string):
+            try:
+                value = int(string)
+                if not value > 0:
+                    raise argparse.ArgumentTypeError("%r is not a positive integer" % (string,))
+            except ValueError:
+                raise argparse.ArgumentTypeError("%r is not an integer" % (string,))
+
```
mdboom: Can you put this in `asv/commands/common_args.py` so it can be used by other commands in the future?

pv: Moved

PR #234, opened 2015-03-01
Clean up command line option nargs usage

pv:
Use nargs='?' in dash-dash options only if const= is set, i.e., only for --parallel. The other dash-dash options require an argument, e.g. `asv run --machine` requires a machine name to make sense.

Remove nargs='?' from the benchmark name argument in asv profile, as that is a required argument.

For the --bench option, use action="append" rather than nargs='*'. Having --bench slurp command line arguments greedily is confusing to use with commands that have positional arguments. For example `asv dev --bench somename master..HEAD` [behaves in a confusing way](https://github.com/scipy/scipy/issues/4574#issuecomment-76538641); `asv dev` doesn't actually take positional arguments but the commit name is slurped by `--bench`.

### DIFF
--- a/asv/commands/compare.py
+++ b/asv/commands/compare.py
@@ -57,15 +57,15 @@ class Compare(Command):
             description="Compare two sets of results")
 
         parser.add_argument(
-            'revision1', nargs=1,
+            'revision1',
             help="""The reference revision.""")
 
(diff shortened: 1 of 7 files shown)


mdboom:
Thanks.

Issue #191, opened 2015-01-09
Labelling specific commits


jbednar:
asv looks great, and we're currently considering using it for performance monitoring for our simulator instead of our own handcrafted and long-broken scripts.  From what I can see, the only feature our scripts had that I don't see in asv is the ability to label specific commits with an explanation of what happened with them.  We found that when we investigated the commit history, there were a few commits that caused major changes in performance (whether for good or for bad), and once we had completed our investigation we labelled those points on the performance charts, with a key that explains what happened.  (E.g. "Added new optimized component XX", "Switched to slower but more general component XX", "Removed potentially unsafe optimization in XX.x", "Rewrote C component in Python using numpy").  Adding these explanations made it clear that we had understood what happened on those specific commits, so that when we came back to look later we didn't start all over again wondering what the big performance drop (or jump) was from.  Does such a facility exist in asv?  If not, would it be difficult to add?  We certainly found it very useful, and appear to have a similar use case to yours...

mdboom:
Such a feature doesn't currently exist, but it seems like it would be very useful.  It does support showing any `git tags` in the graph -- but I understand that's not exactly what you want, but that feature could be extended to support what you want fairly easily, I would think.

You could probably experiment with it from outside in fact.  When you do an `asv publish` to generate the html/json, in `html/index.json` there should be a `tags` entry, like:

```
"tags": {
        "v0.1": 1340086193000, 
        "v0.2": 1361312808000, 
        "v0.2.1": 1365018342000, 
        "v0.2.2": 1369169818000, 
        "v0.2.3": 1369943695000, 
        "v0.2.4": 1374690436000, 
        "v0.2.5": 1382718598000, 
        "v0.2b1": 1356381732000, 
        "v0.2b2": 1359508111000, 
        "v0.2rc1": 1360776151000, 
        "v0.3": 1384996150000, 
        "v0.3.1": 1393955310000, 
        "v0.3.2": 1399997164000, 
        "v0.3b1": 1383542385000, 
        "v0.3rc1": 1384388274000, 
        "v0.4": 1405527129000, 
        "v0.4.1": 1407512129000, 
        "v0.4.2": 1411505522000, 
        "v0.4rc1": 1404233912000, 
        "v0.4rc2": 1405176393000
    }
```

The first is a string, the second is a javascript time stamp of an associated commit.  You could add comments in the string part as long as you know the date time stamp, and they should appear in the graph.

[...]

PR #236, opened 2015-03-04
Automated regression detection

pv:
Here's some work on detecting and listing benchmark regressions automatically:
Sample output: https://pv.github.io/scipy-bench/#/regressions
http://pav.iki.fi/tmp/asv-236/

It works by fitting piecewise constant functions to all graphs, and using the result to determine the best value  and the date+commit where it regressed.

Fitting piecewise constant functions has exact solutions in `O(n^2)` time, for minimizing objective functions of the form `f = gamma * {#jumps in y_fit} + sum((y - y_fit)**p)` for p=1 or 2. With some reasonably sounding heuristics apparently good solutions can be found in linear time. The parameter gamma is picked by minimizing `beta {#jumps} ln(n)/n + log sum((y - y_fit)**p)` which apparently is some information criterion for gaussian models (no rigor is attempted, it seems to work reasonably with beta=3). 

For 2-norm, all this seems to work fast enough up to 50k+ commits. The 1-norm is approx 10x slower, as I didn't manage to optimize it better in pure Python.

- [x] Some more testing with real-world data could be useful.
- [x] Better unit tests for generated data
- [x] Maybe the plugin should be made optional?
- [x] The "regressions" web page was factored out of asv.js, making the setup more extensible. I didn't go all the way in factoring everything to pieces. Looks reasonably future-proof to me, but ymmv?
- [x] Not sure what's the best way of reporting the results, e.g., should it only report the most recent regression rather than comparing to the best value?
- [x] Preserve table sort if user presses back button
- [x] show commit ranges instead of commits
- [x] way to specify earliest commit to regress in asv.conf
- [x] deal properly with branches and multiple machines
- [x] way to ignore specific regressions
- [x] fine-tune step detection algorithm --- it's not fully satisfactory as of f33b30e, and tends to report spurious regressions at end of the time series


mdboom:
This is really cool.

This assumes that all commits have been benchmarked, correct?  For example, on the astropy benchmarks (which you've kindly provided an example of), the commit link for the regressions is irrelevant because the commits are only subsampled.

It seems that ideally this search of the regression space would be combined with the `asv bisect` command to "dig down" and find the specific commits that are to blame.

pv:
Yes, having the commit links assumes everything is benchmarked. It would be straightforward to display commit ranges instead of commits, if the regressing commit is ambiguous, or give e.g. copypasteable asv bisect command lines.

A different practical question is: a typical situation seems to be that somewhere in history the performance obtained its best value, and then regressed. After that, the regression was fixed (or something else happened), but it's still worse than the original value. What's the most useful way to report this?

The current way probably isn't too bad, although it would be useful to mark regressions as "known" to exclude them from the report, a bit similar to the suggested idea to add labels in gh-191

mdboom:
> Yes, having the commit links assumes everything is benchmarked. It would be straightforward to display commit ranges instead of commits, if the regressing commit is ambiguous, or give e.g. copypasteable asv bisect command lines.

That sounds like a good compromise, and I think would avoid initial confusion.  Something that automatically runs git bisect based on a reasonable place to look can always be done as a follow-on.

> although it would be useful to mark regressions as "known" to exclude them from the report, a bit similar to the suggested idea to add labels in gh-191

Would a way of marking the "earliest" commit to report on (on a benchmark-by-benchmark basis) make sense for that?

pv:
Ok, I think this PR is converging now. The performance seems reasonable even for the more robust 1-norm fitting, and false positive rate seems low enough. I'd rather waste a minute of computer time for analysis than wasting developer time for eyeballing squiggly lines. The UI now reports all detected "significant" breakpoints. The significance threshold in principle could be bumped up, but the false positive rate seems low enough currently, at least based on the astropy/scipy results linked to above.

After getting this work in practice, I also added some amateur-level statistical analysis arguments to the docs to get it work also in theory.

mdboom:
This is cool.  It's too bad the ignore buttons use local storage (and thus other users viewing the same results will still see this) but I don't see a way around that without a server-side component.  Is there a way the ignore selections could be exported and then committed to the git repo containing the results?

pv:
Yes, it could be useful to have the web frontend output something that
can just be copypasted to asv.conf.json. The main question here then are
(i) whether such builtin config should just hide the results by default
so that you can still unignore them, or, completely remove them, and
(ii) whether to put the config to asv.conf.json or some other config file.

pv:
Ok, added a widget showing copypasteable config file excerpt for ignoring the currently ignored regressions.

The config file option limits the commit ranges considered in computation, so that it removes the regressions in question rather than just hiding them. I didn't want to have two ways of ignoring regressions in the config file, and since limiting computation might be required in some cases, I just went with that.

mdboom:
> I didn't want to have two ways of ignoring regressions in the config file, and since limiting computation might be required in some cases, I went with that.

That makes sense.

mdboom:
Sorry this has sat here for a while.  @pv: Good to merge in your opinion?  (Looks good to me).

pv:
I think this has converged, and should be fairly usable currently. I don't have remaining TODO items currently.

I know there may be more robust change point detection algorithms in the literature, but within the constraints here (pure-Python, no special functions, must be fast) I think this is reasonable effort for the purpose.

pv:
Ok, merging.

PR #239, opened 2015-03-08
Don't use --python=same as the default value -- breaks asv run etc

pv:
Set the different default value for 'asv dev' in a different way

### DIFF
--- a/asv/commands/common_args.py
+++ b/asv/commands/common_args.py
@@ -35,10 +35,10 @@ def add_machine(parser):
         ~/.asv-machine.json, that one entry will be used.""")
 
 
-def add_python(parser):
+def add_python(parser, default=None):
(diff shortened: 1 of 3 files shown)


test/test_commands.py
```diff
@@ -0,0 +1,22 @@
...
+    args = parser.parse_args(argv)
+    assert args.python == 'same'
+
+    argv = ['dev', '--python=foo']
+    args = parser.parse_args(argv)
+    assert args.python == 'foo'
+
+    argv = ['run', 'ALL']
+    args = parser.parse_args(argv)
+    assert args.python is None
```
mdboom: Can you put this in `test_dev` instead?

pv:
Ok, fixed up.

PR #241, opened 2015-03-08
Avoid direct file path manipulation in plugin loading, and use standard import mechanism

pv:
The plugin mechanism apparently conflicts with the py.test test collection in some cases.
https://travis-ci.org/spacetelescope/asv/jobs/53561921  (this was caused by having `from asv import command` in the test suite on top level)

It seems the plugin loading code can be simplified somewhat, by relying on a more standard scheme on iterating over modules and importing. I'm not sure this will help with py.test, but it looks like it's better to do things this way.

### DIFF
--- a/asv/plugin_manager.py
+++ b/asv/plugin_manager.py
@@ -4,9 +4,8 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
-import imp
-import os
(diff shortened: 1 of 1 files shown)


mdboom:
Thanks.  I wasn't aware of `pkgutil`.

PR #243, opened 2015-03-09
Generate test repositories on the fly.

### LINKED ISSUE #233: Proper test repository data
Currently the test suite uses live repositories as test data (asv and some other project for mercurial).
This is slightly awkward for writing tests since you need to find commits for existing test cases rather than just adding new test commits, and it's also somewhat slow due to cloning over the network.

It would be better if the test repositories were git and hg repositories, stored for example in zip files so that they are reproducible and can be manipulated without too much trouble. (You cannot store git repo in a git repo.)

mdboom:
Fix #233.

Rather than the suggestion in #233 -- to use zipped repositories -- this just generates them in a temp directory on the fly.  This should make it easier to inject new kinds of synthetic content in the repositories in the future (and creating these repositories isn't much slower than unzipping would be, at least in the case of git).

This also really simplifies how some of the tests are written, and makes the test of `asv find` much "truer".

Cc: @pv


pv:
Looks good to me. I'd maybe rename 'test/test_util.py` to `test/util.py` since tests for `asv/util.py` would perhaps go to `test_util.py`.

mdboom:
Good point about naming.  I actually didn't want to name it util because then it's confusing within the files that import both `asv.util` and `test.util`, but I can come up with something better...

mdboom:
NB: It does seem like this change may cause some merging grief with some of your open PRs.  Sorry about that.

pv:
Maybe `tools.py`?

pv:
I think it's best if this PR goes in first --- it's easier for me to work the branch tests etc based on this rather than vice versa.

PR #245, opened 2015-03-14
Fix mismatch in index json and graph branch names

pv:
Another bug from gh-232, in the case when no branches are specified in config

### DIFF
--- a/asv/commands/publish.py
+++ b/asv/commands/publish.py
@@ -83,6 +83,18 @@ def check_benchmark_params(name, benchmark):
             raise ValueError(msg)
 
 
+def safe_branch_name(branch):
+    """
+    Convert a branch name to a string, dealing with the None default value
+    """
+    if branch is None:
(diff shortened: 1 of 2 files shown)


asv/commands/publish.py
```diff
@@ -83,6 +83,18 @@ def check_benchmark_params(name, benchmark):
...
 
 
+def safe_branch_name(branch):
+    """
+    Convert a branch name to a string, dealing with the None default value
+    """
+    if branch is None:
+        # Occurs only if no branch is set in the configuration, and is
+        # not visible to the user.
+        return "master"
```
mdboom: Possibly naive question: Does this work with Mercurial?

pv: yes, it only affects the directory name. it''s not visible to the user or passed back to the repository

mdboom: Got it.  Seems fine then.

PR #249, opened 2015-03-18
Javascript integration testing

pv:
Some work on getting full-stack integration test framework up. Based on some googling, Selenium WebDriver seems to be a mainstream approach to the problem. The Python bindings seem to be fairly complete and are documented. PhantomJS is installed on Travis-CI by default, so the tests are also run automatically.

Currently, this only has a simple smoke test, more could be added later on (separate PRs).

The javascript code as it is currently written is not unit testable. Some reorganization could be done, and using a javascript unit test framework might be worth considering.

I'll file this PR early to see if the Travis-CI integration works (EDIT: seems to work); there may still be some things left to fix here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ install:
     fi
     if [[ $TRAVIS_PYTHON_VERSION == 2* ]]; then $TRAVIS_PIP install "mercurial>=3.3" ; fi;
     if [[ $TRAVIS_PYTHON_VERSION == 2* ]]; then $TRAVIS_PIP install python-hglib==1.5 ; fi;
+    $TRAVIS_PIP install selenium
 
 script:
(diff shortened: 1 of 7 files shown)


mdboom:
Thanks for getting the ball rolling on this.

mdboom:
@pv: Do you still think there's some things left to fix here, or is this ready to merge?

pv:
I have some minor cleanup pending, didn't yet find time to come back to
this.

pv:
Should be OK now.

Asynchronous wait for an event that should *not* happen (appearance of the error box here) turns out to be difficult to do, so I had to leave the explicit time.sleep in in the smoketest.

pv:
I'm planning to merge this soon, should be OK as it is currently.

PR #254, opened 2015-03-21
Ensure subprocesses are terminated on timeout even if they have spawned their own subprocesses

pv:
It's possible for `check_output` to not exit within a specified timeout, if the subprocess spawns a sub-sub-process process that does not exit. The termination signal sent to the subprocess is not propagated automatically to the sub-sub-one on posix, and the wait for the subprocess doesn't finish before the sub-sub-process exits.

To fix this, launch the subprocess in a [new process group](https://en.wikipedia.org/wiki/Process_group) using [setpgid](http://pubs.opengroup.org/onlinepubs/009695399/functions/setpgid.html). On error/exception, kill the subprocess by sending the signal to the whole process group.

Also move the termination to a `finally:` block, ensuring ASV never leaves subprocesses running in the background.

(Real-world example:
```
cd scipy/benchmarks
asv run --bench interpolate.Leaks.track_leaks 6a035d6a^..6a035d6a
```
)

### DIFF
--- a/asv/util.py
+++ b/asv/util.py
@@ -18,6 +18,7 @@ import subprocess
 import struct
 import sys
 import time
+import errno
 
 try:
     from select import PIPE_BUF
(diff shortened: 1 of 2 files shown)


mdboom:
A somewhat related question -- does the `process_time` function (which is used to actually time the benchmark) work correctly with child processes.  I think the expectation would be that the time includes the benchmark process and all child processes, but we should maybe confirm that's the case, and possibly add a test. 

Otherwise, thanks for this.  I think scipy's needs are a bit beyond most simpler projects, so it's great to have these corner cases ironed out.

pv:
`CLOCK_PROCESS_CPUTIME_ID` does not count time spent in subprocesses.

The subprocess-spawning benchmarks in Scipy are for measuring peak memory usage, which may be of general interest. Maybe a candidate for new benchmark type for ASV, or maybe for replacing the current memory benchmark? The current memory benchmark is not very useful for me, since it measures only memory usage of a single Python object. asizeof doesn't count memory used by non-python objects (including Numpy arrays!), and memory used by temporary storage allocated during computations also cannot be counted looking at the memory used by the final result only.

mdboom:
I'd be happy to add an additional memory benchmark to track peak memory usage, but `asizeof` is useful in other contexts, so I wouldn't want to give it up.

If there's an easy way to track the CPU time of all child processes, it would be great to use it (don't have time to dig down on that today) ... otherwise maybe we just document somewhere that our time benchmarks don't work that way.

pv:
Time spent in child processes can be obtained in principle, http://linux.die.net/man/2/times 
However, this doesn't seem to count the time spent in grandchildren, and has some other caveats.

It may be simpler to just document that the default is to measure the per-process time, and in cases where one uses e.g. multiprocessing or other approaches that fork processes it's best to set `timer = timeit.default_timer` to measure the wall-clock time instead.

mdboom:
> It may be simpler to just document that the default is to measure the per-process time, and in cases where one uses e.g. multiprocessing or other approaches that fork processes it's best to set timer = timeit.default_timer to measure the wall-clock time instead.

Seems reasonable.

mdboom:
Created #257 to address my question.  Otherwise, this looks fine, so merging.

PR #256, opened 2015-03-24
Remove unused routine & fix race condition

pv:
There's an unused `_create_multiprocess` routine that could be removed.

When replacing `install_requirements` -> `create` in gh-252 I (accidentally) made it call `_create` instead of `_create_multiprocess`. However, now that I look at it again, `_create` and `_create_multiprocess` are the same (they were so [also before](https://github.com/spacetelescope/asv/blob/bb16d18185756a737be6579d17a526d243695242/asv/commands/setup.py#L14)), so one can be removed, seems to work OK currently.

### DIFF
--- a/asv/commands/setup.py
+++ b/asv/commands/setup.py
@@ -24,15 +24,6 @@ def _create(env):
         raise
 
 
-def _create_multiprocess(env):
-    try:
-        return _create(env)
-    except:
-        import traceback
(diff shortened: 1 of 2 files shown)


pv:
Heh, the tests seem to have uncovered (an unrelated) race condition in parallel environment creating. Fix incoming...

PR #257, opened 2015-03-26
Note about timing of spawned child processes.

mdboom:


pv:
Note that I probably was wrong about `os.times()[2]` not measuring grandchild process time, the documentation was somewhat ambiguous.

In principle, it would be possible to measure `process_time() + os.times()[2]` --- however, that timer apparently has a rather poor resolution (10ms on ubuntu 14.10).

mdboom:
Maybe we merge this, but open a new issue that it would be nice to solve the "measuring child process time" problem at some point?

pv:
Sounds good.

PR #259, opened 2015-04-12
Provide user-visible feedback when benchmark times out

pv:
Inform the user if a benchmark was terminated due to a timeout, by appending a message in the error stream. This should make debugging things easier (currently, timeout is signalled only by a `failed` text, with no indication for the reason).


mdboom:
:+1:

PR #260, opened 2015-04-12
Add peak memory benchmarks

pv:
Add a benchmark type for measuring peak memory usage.


mdboom:
Thanks!  I know people have been asking for this.

Issue #261, opened 2015-04-16
self tests stall


rbiswas4:
I am trying to install airspeed velocity on a OS X macbook pro laptop:
Darwin Kernel Version 14.3.0 (Yosemite 10.10.3) 

Install seemed to go without complaints
```python setup.py test
running test
running egg_info
writing requirements to asv.egg-info/requires.txt
writing asv.egg-info/PKG-INFO
writing top-level names to asv.egg-info/top_level.txt
writing dependency_links to asv.egg-info/dependency_links.txt
writing entry points to asv.egg-info/entry_points.txt
reading manifest file 'asv.egg-info/SOURCES.txt'
writing manifest file 'asv.egg-info/SOURCES.txt'
running build_ext
=========================================================================================================================================================================== test session starts ============================================================================================================================================================================
platform darwin -- Python 2.7.9 -- py-1.4.25 -- pytest-2.6.3
collected 28 items / 1 errors

test/test_benchmarks.py ...
test/test_compare.py .
test/test_conf.py ..
test/test_dev.py .....
test/test_graph.py ....
test/test_machine.py .
test/test_publish.py .
test/test_quickstart.py .
test/test_repo.py .x
test/test_results.py .
test/test_rm.py .
test/test_subprocess.py ..
test/test_workflow.py ..
```

While some of the previous tests take a bit of time to complete, I have not been able to get through the test_workflow.py at all, even after a couple of hours, but the process has not thrown any error or warnings aside from the error mentioned at the start, after which the tests commenced.

pv:
Try running with `py.test test/ -svv` and see if the output explains
why/where it stalls.

rbiswas4:
Ok, 
```
 py.test test/ -svv >& test_dump.dat
```
I did that and I can see that the stall happens (still running for about 2 hours) at 
```
7240.00%] · Generating graphs
[7260.00%] · Writing indexPASSED
test/test_workflow.py::test_continuous PASSED
test/test_workflow.py::test_find
```
There are also a few runtime errors in the output. I can't attach the file here, so I just put it in a scratch repository https://github.com/rbiswas4/scratch/blob/master/asv_res/test_dump.dat if you are interested.

pv:
From the log, you can see that what is hanging is calling `anaconda
create`, so it is fairly possible this is a bug in Anaconda. You can try
look into what causes that, and e.g. upgrading your Anaconda installation.

pv:
[...]

PR #262, opened 2015-04-18
Work around multiprocessing bug with ProcessError exceptions

pv:
Multiprocessing on Python 2.7.8 fails to deal properly with some
exception types, leading to a hang if they are raised.  See
https://bugs.python.org/issue9400

Work around this by explicitly wrapping all exceptions raised in
multiprocessed functions in a custom ParallelFailure exception type.
Add some extra code so that util.UserErrors are reraised properly.

Fixes gh-261

### DIFF
--- a/asv/commands/run.py
+++ b/asv/commands/run.py
@@ -7,6 +7,7 @@ from __future__ import (absolute_import, division, print_function,
 import six
 
 import logging
+import traceback
 
 from . import Command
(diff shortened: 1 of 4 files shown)


pv:
Build failure unrelated (apparently conda network failure when trying to do Conda.match)

pv:
Planning to merge soon, as the issue is triaged.

PR #264, opened 2015-04-18
Indicate in ProcessError if the command timed out

pv:
It can be helpful if the fact that a command timed out is indicated also in error messages.


pv:
Straightforward, so merging. (Travis failure is due to conda timeouts.)

PR #266, opened 2015-05-01
Use py.test stdout/stderr capture mechanism in tests

pv:
Use py.test native stdout/stderr capture mechanism in tests, rather than manually changing sys.stdout.

This has the advantage that if an exception is raised during the time the output is redirected, py.test will print the output (e.g. from ProcessError logging) which makes debugging easier. This does not occur automatically when sys.stdout is redirected manually; if we do the printing manually, some output can still be lost when using multiprocessing. The fd-level py.test output capture seems to solve this issue.

### DIFF
--- a/test/test_compare.py
+++ b/test/test_compare.py
@@ -13,7 +13,6 @@ import six
 from asv import config
 
 from asv.commands.compare import Compare
-from io import StringIO
 
 from . import tools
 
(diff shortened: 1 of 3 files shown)


pv:
Straightforward (and needed to make gh-261 debuggable), so merging.

mdboom:
Thanks.  I wasn't aware of that feature in py.test.

PR #267, opened 2015-05-01
Prepare for making releases

mdboom:
Adds a few things, like a changelog, manifest, zest.releaser plugins etc. to prepare us to start making releases.


setup.py
```diff
@@ -53,38 +45,64 @@ def get_version():
...
     if content != old_content:
-        with open(fn, 'w') as f:
+        with open(filename, 'w') as f:
             f.write(content)
 
-    # The revision is not added to the version given to setup(), since
-    # that disturbs 'setup.py develop'
-    return version
+
+version = '0.1dev'
```
pv: Note that PEP 440 prescribes the `dev0` suffix instead of `dev`, and pip/setuptools complain if it's not there.

mdboom: It gets appended with the first few characters of the githash below.  However, reading PEP 440 again, it seems this is supposed to be a positive integer, and by that, I suspect an ordinal one, too.  I've updated this to use `rev-list --count`.

pv: `rev-list --count` does not tell the user which git version they are running. Better (and PEP440) compatible option would be to either use the form `dev0+HASH` or `devN+HASH`, where `N` is the rev-list count, and `HASH` the git hash.

mdboom: Ah, ok.  That makes sense.

PR #268, opened 2015-05-02
Document that 'repo' config value can also be a relative path

pv:


docs/source/using.rst
```diff
@@ -45,11 +45,17 @@ suite are:
 
 - ``project_url``: The project's homepage.
 
-- ``repo``: The URL to the DVCS repository for the project.  This
+- ``repo``: The URL or path to the DVCS repository for the project.  This
   should be a read-only URL so that anyone, even those without commit
   rights to the repository, can run the benchmarks.  For a project on
   github, for example, the URL would look like:
-  ``https://github.com/spacetelescope/asv.git``
+  ``https://github.com/astropy/astropy-benchmarks.git``
```
mdboom: This isn't a good example, since `astropy-benchmarks` contains only benchmarks -- it's not the project being benchmarked.  `https://github.com/astropy/astropy.git` would be fine, however.

pv: Ok, indeed.

PR #269, opened 2015-05-09
Minor fixups vs tarball

pv:
The asv tarball is missing some files, the license file and an image file in the docs.
The `publish_docs.sh` script is obsolete, as the docs are on rtd.
Also, add links to Github & PyPi to the front page of the docs --- currently navigating to these is not so easy from the readthedocs page

(This PR is based on the merge-base of the stable&master, so should be mergeable to both. The v0.1.x and stable branches however don't seem to be in sync currently on github.)


mdboom:
Thanks.  Looks like it needs a rebase, then I think this is ready to merge.

pv:
Ok, merged

Issue #275, opened 2015-06-16
commands sub-package shadows the builtin commands module for python2


cfarrow:
This is for `ASV 0.1.1`.

This causes import issues when benchmarking packages that use the builtin `commands`. For example, when profiling a package that uses `Biopython 1.65` I see the following...

```python
...
     from Bio import Seq
File "/Users/cfarrow/Library/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/Bio/Seq.py", line 27, in <module>
    from Bio._py3k import range
File "/Users/cfarrow/Library/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/Bio/_py3k/__init__.py", line 255, in <module>
    from commands import getoutput
File "/Users/cfarrow/projects/genia/genia-test-cases/env/lib/python2.7/site-packages/asv/commands/__init__.py", line 12, in <module>
    from .. import config
ValueError: Attempted relative import beyond toplevel package
```

An "easy" way to deal with this is to rename the `commands` subpackage. Unfortunately, this is likely to break plugins that are not in this repo.

cfarrow:
I'm happy to submit a renaming patch, but I'd rather have a new subpackage name chosen ahead of time.

pv:
I think it's first necessary to understand how it ends up in `sys.modules` at all, and to fix that.
As it's a sub-module, renaming shouldn't be necessary.

For me,
```
>>> import sys, asv.plugins
>>> sys.modules['asv.commands']
<module 'asv.commands' from '/home/pauli/.local/lib/python2.7/site-packages/asv/commands/__init__.pyc'>
>>> sys.modules['commands']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'commands'
>>> import commands
>>> sys.modules['commands']
<module 'commands' from '/usr/lib/python2.7/commands.pyc'>
```
Do you understand how your case is different?

pv:
Ok, indeed, it shows up when run in a benchmark. Probably has to do with the fact that  the `benchmark.py` script is located in the `asv` directory. The issue is probably solved by moving it to a non-package subdirectory.

pv:
Fix in gh-276

cfarrow:
#276 indeed fixes my issue. Thanks!

PR #276, opened 2015-06-16
Ensure benchmark.py script doesn't have asv package directory in sys.path

pv:
Fixes gh-275


mdboom:
:+1:

pv:
Will merge after travis finishes...

PR #277, opened 2015-06-17
Fix #274: Add setup_cache method support

### LINKED ISSUE #274: Add per-class setup / teardown
Similarly to what modern unittest offers, it would be nice to have per-class setup and teardown functions, to set up class-wide resources (such as data samples, etc.)

(see https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass)

mdboom:
Caches results in a shelf between benchmark runs.

Cache is created/destroyed once per environment.

### DIFF
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -497,23 +497,26 @@ class Benchmarks(dict):
             with log.indent():
                 benchmarks = sorted(list(six.iteritems(self)))
 
+                if skip:
+                    benchmarks = [
+                        (name, benchmark) for (name, benchmark) in
+                        benchmarks if name not in skip]
+
                 setup_caches_done = set()
                 for name, benchmark in benchmarks:
                     key = benchmark.get('setup_cache_key')
                     if key is not None and key not in setup_caches_done:
                         log.info('Caching {0}'.format(key))
                         setup_caches_done.add(key)
-                        env.run(
+                        out, err, errcode = env.run(
                             [BENCHMARK_RUN_SCRIPT, 'setup_cache',
                              self._benchmark_dir, name,
                              cache_file.name],
-                            dots=False, display_error=True,
-                            return_stderr=True)
+                            dots=False, display_error=False,
+                            return_stderr=True, valid_return_codes=None)
 
                 times = {}
                 for name, benchmark in benchmarks:
-                    if skip and name in skip:
-                        continue
                     times[name] = run_benchmark(
                         benchmark, self._benchmark_dir, env,
                         cache_file.name, show_stderr=show_stderr,
@@ -529,7 +532,7 @@ class Benchmarks(dict):
                             [BENCHMARK_RUN_SCRIPT, 'teardown_cache',
                              self._benchmark_dir, name],
                             dots=False, display_error=True,
-                            return_stderr=True)
+                            return_stderr=True, valid_return_codes=None)
 
         finally:
             if os.path.isfile(cache_file.name):
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -519,7 +519,6 @@ class Benchmarks(dict):
                                 # TODO: Store more information about failure
                                 times[name] = {'result': None}
                             continue
-                        assert os.path.isfile(os.path.join(tmpdir, 'cache.pickle'))
 
                     for name, benchmark in benchmark_set:
                         times[name] = run_benchmark(
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -29,7 +29,8 @@ BENCHMARK_RUN_SCRIPT = os.path.join(
     os.path.dirname(os.path.abspath(__file__)), "benchmark.py")
 
 
-def run_benchmark(benchmark, root, env, show_stderr=False, quick=False,
+def run_benchmark(benchmark, root, env, setup_class_cache_file,
+                  show_stderr=False, quick=False,
                   profile=False):
     """
     Run a benchmark in different process in the given environment.
@@ -42,6 +43,9 @@ def run_benchmark(benchmark, root, env, show_stderr=False, quick=False,
         Path to benchmark directory in which to find the benchmark
 
     env : Environment object
+
+    setup_class_cache_file : str
+        Path to class-level setup cache file.
 
     show_stderr : bool
         When `True`, write the stderr out to the console.
@@ -96,8 +100,10 @@ def run_benchmark(benchmark, root, env, show_stderr=False, quick=False,
 
         for param_idx, params in param_iter:
             success, data, profile_data, err, out, errcode = \
-                _run_benchmark_single(benchmark, root, env, param_idx,
-                                      quick=quick, profile=profile)
+                _run_benchmark_single(
+                    benchmark, root, env, param_idx,
+                    quick=quick, profile=profile,
+                    setup_class_cache_file=setup_class_cache_file)
 
             total_count += 1
             if success:
@@ -184,7 +190,8 @@ def run_benchmark(benchmark, root, env, show_stderr=False, quick=False,
         return result
 
 
-def _run_benchmark_single(benchmark, root, env, param_idx, profile, quick):
+def _run_benchmark_single(benchmark, root, env, param_idx, profile, quick,
+                          setup_class_cache_file):
     """
     Run a benchmark, for single parameter combination index in case it
     is parameterized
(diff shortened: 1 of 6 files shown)


pv:
Some points to consider:
.
- This is somewhat magical.
- The documentation probably needs to explain the benchmark isolation
mechanism in more detail, the fact that this setup method would not
always be run if cached data exists, and the lifetime of cached data.
- It probably should always set the attributes by unpickling them, so
that the behavior is the same regardless of whether it's the first run,
or a later run.
- Maybe better to not follow unittest camelCase naming conventions, but
just go with `setup_class` a la Nosetests?
- Maybe not call it `setup_class`, since the behavior differs from the
test frameworks (e.g., the method is not run if cached data exists).
- Maybe the fact that pickling is done should be made more explicit, by
e.g. pickling a returned dict. Or perhaps don't pickle at all, but pass
in a name of cache file/directory somehow?
- Module-level methods (cf Nose), since there are also module-level
benchmarks?

mdboom:
All good points.  This was a fifteen minute hack -- probably shouldn't have been made into a pull request just yet.

@pv wrote:
> - This is somewhat magical.

True -- but I think somewhat less problematic than trying to run multiple benchmarks in the same process where each could fail independently etc.

> - The documentation probably needs to explain the benchmark isolation
mechanism in more detail, the fact that this setup method would not
always be run if cached data exists, and the lifetime of cached data.

True.  It's already discussed elsewhere, but it's worth repeating in the discussion of the new `setup_class` method.

> - It probably should always set the attributes by unpickling them, so
that the behavior is the same regardless of whether it's the first run,
or a later run.

Agreed.  And beyond the difference in data structures that may result due to pickling there may be a difference in memory usage which may affect the timing of the benchmark.  I think we should probably run all of the `setup_class` setup in a separate process, then run all of the benchmarks in their own to prevent things like that.

> - Maybe better to not follow unittest camelCase naming conventions, but
just go with `setup_class` a la Nosetests?

Sure.  Also I think perhaps "setup_cache" is a better name than "setup_class" if this will also be module level or function-attribute level.

> - Maybe not call it `setup_class`, since the behavior differs from the
test frameworks (e.g., the method is not run if cached data exists).

Agreed.  I think "setup_cache" maybe gets that point across...  though I'm open to other names...

> - Maybe the fact that pickling is done should be made more explicit, by
e.g. pickling a returned dict. Or perhaps don't pickle at all, but pass
in a name of cache file/directory somehow?

It turns out this is also required for module level functions (since there will be no `cls` or `self` passed in).  I think `setup_cache` will have to return an object (most likely a dictionary, but not required) and this object will get passed to the benchmarking function.

I think pickling a returned object will be most convenient for the majority of use cases.  For the case where controlling the serialization is important, the user can always write to a `BytesIO` buffer and return the resulting string.

> - Module-level methods (cf Nose), since there are also module-level
benchmarks?

Agreed.  We already do that for other special methods, so we should do that here, too.

pv:
Ok, sounds useful. Running cache setup in separate process would also
make things a bit easier for the peak memory benchmarks. Having
`teardown_cache` could also be useful, in case you need to create
temporary files.

asv/benchmarks.py
```diff
@@ -462,26 +468,67 @@ def run_benchmarks(self, env, show_stderr=False, quick=False, profile=False,
...
+                    key = benchmark.get('setup_cache_key')
+                    if key is not None and key not in setup_caches_done:
+                        log.info('Caching {0}'.format(key))
+                        setup_caches_done.add(key)
+                        env.run(
+                            [BENCHMARK_RUN_SCRIPT, 'setup_cache',
+                             self._benchmark_dir, name,
+                             cache_file.name],
+                            dots=False, display_error=True,
+                            return_stderr=True)
```
pv: Failure in cache setup (or teardown) routine probably shouldn't result to benchmark run failure, just to marking that benchmark as failed. The call to setup_cache might be best done actually in `run_benchmark` or just before calling it; it probably should respect the `skip` list here.

asv/benchmarks.py
```diff
@@ -462,26 +468,67 @@ def run_benchmarks(self, env, show_stderr=False, quick=False, profile=False,
...
-                    benchmark, self._benchmark_dir, env, show_stderr=show_stderr,
-                    quick=quick, profile=profile)
+
+        cache_file = tempfile.NamedTemporaryFile(suffix='dbm', delete=False)
+        cache_file.close()
+        # We have to actually delete the file here, since the database
+        # type can not be determined from a zero-length file.  We
+        # can't create the database in the master process, since the
+        # file formats available may vary between Python environments.
+        os.remove(cache_file.name)
```
pv: I think this allows a local race condition attack with arbitrary code execution.
`mkdtemp` is probably a cleaner solution without such problems.

mdboom: Good point.

pv:
This looks somewhat similar to `py.test` fixtures. An approach modeled more closely after that (but with less magic) could look like
```
def my_big_list():
    return [0]*500

class MyClass:
    fixtures = [my_big_list]

    def time_len(self, p):
        len(p)
```
The implementation would be pretty much as above, with minor extensions. The fixtures might also be passed to the setup/teardown routines similarly as done for parameters.

asv/benchmarks.py
mdboom: Agreed about the failure.  I'll fix that.

I do all of this upfront to be careful about spawning as few processes as possible (by tracking the `setup_caches_done` set).  That could probably done inside of `run_benchmark` but would be more cumbersome.  Is there any advantage to that?

Agree that it needs to track the skip list.  I might just pre-filter the benchmark list.

pv: Maybe not that much advantage. I was thinking about keeping the return value construction more localized, but that function is already somewhat too big.

mdboom:
The fixtures suggestion is an interesting one.  I guess what you mean by "less magic" is that rather than matching fixtures to the arguments by the name of the argument, they would just be listed explicitly in the fixtures list?  That seems more tenable than what py.test does.

pv:
Yes, the way py.test does the argument matching by name + global registry is a bit magical. It would be more explicit to list them explicitly in an attribute and then just concatenate them to the argument list, in the same way parameters are handled.

pv:
Ah, I forgot about fixture teardown... Could be an attribute to the fixture method, or maybe specified as a tuple etc., but that gets a bit more ugly.

mdboom:
Thinking about this a bit more, I'm not sure `teardown` is terribly useful in this context anyway.  There's sort of two scenarios here:

1. `setup_cache` creates some things that are easily picklable, and asv manages their caching automatically

2. `setup_cache` wants to do things on the filesystem directly, in which case it needs to clean up after itself.

I'm not sure there are really any more complicated use cases for teardown (as there are in py.test/nose etc.) since the whole process is torn down anyway.  Maybe if setup_cache starts up some daemon that the benchmarks then communicate with or something...

What if `setup_cache` is passed a tempdir path and the convention is "save anything you'll need there -- asv will delete the tempdir when we're done automatically".  Then it just becomes a question of how the benchmarking function obtains the tempdir path...

pv:
Yes, temporary files seem the main use case for cache teardown. I just
noticed this cannot be used for addressing the peakmem setup issue, at
least for parameterized peakmem benchmarks. The subprocess thing seems
still the simplest approach here...

mdboom:
@pv: Just curious: why doesn't this address the peakmem setup issue?

pv:
@mdboom: What the benchmark setup in question does is to create a `.mat` file containing an array of a given size, with compression on or off. The size/compression are a parameter of the benchmark, so the cache setup would need to be parameterized. Of course, I could in principle create all of the files necessary in one go (requires more spare disk space for the combinations, though).

mdboom:
Ah, I see.  In the unparameterized case it would work, though, if I understand correctly.  Maybe as a follow-on to this we could add support for parameterizing `setup_cache` functions...

mdboom:
This has been simplified in the following ways:

- No more `teardown_cache`.  I don't think it's really needed.

- `setup_cache` may save files to the current working directory, which is a temporary working directory thrown out after ``asv`` knows it's done with the results

- The setup caches are no longer in a shelf, but in a per-setup-cache pickle.  This data is thrown out immediately after being finished with, rather than after a whole set of benchmarks, to minimize the disk space used at any given time.

pv:
Looks good to me FWIW. The main missing piece indeed was using a single
semi-persistent cwd.

mdboom:
I've rebased this, which should hopefully give it some testing on Appveyor now.

@pitrou: It would be great to hear whether this addresses your original use case.

pitrou:
It wouldn't help my use case, since my setup() is JIT-compiling a function with Numba. It may address other people's use cases, though :-)

pv:
If you like ugliness, you can in principle run benchmarks in `setup_cache` with custom timing routines, and then load and return the results from track_ entry points... I would be tempted to do it this way in scipy optimizer benchmarks, to return several measurements without having to run everything N times.

PR #281, opened 2015-06-25
Add Appveyor support

mdboom:
I expect this to fail without the fixes in #280.

Note this goes a little beyond what we need by including support for compiling C extensions (which we don't have), but it was easy enough to do since it was already in astropy's recipe.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,47 @@
+# AppVeyor.com is a Continuous Integration service to build and run tests under
+# Windows
+
+environment:
+
+  global:
+      PYTHON: "C:\\conda"
+      MINICONDA_VERSION: "latest"
+      CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\.continuous-integration\\appveyor\\windows_sdk.cmd"
+
+  matrix:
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -0,0 +1,47 @@
...
+
+install:
+    # Install miniconda using a powershell script.
+    - "powershell .continuous-integration/appveyor/install-miniconda.ps1"
+    - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+
+    # Install the build and runtime dependencies of the project.
+    - "conda update --yes conda"
+    # Create a conda environment
+    - "conda create -q --yes -n test -c python=%PYTHON_VERSION%"
```
pv: Probably needs some name after `-c`, cf https://github.com/astropy/astropy/blob/master/appveyor.yml#L36

mdboom:
This is now hitting one of the failures addressed by #280, so it's probably ready to be merged into that one.

pv:
+1

mdboom:
See rcgibbo/asv#1

PR #282, opened 2015-06-27
Pr 280 support windows

pv:
Rebased https://github.com/spacetelescope/asv/pull/280 plus some additional fixes for Windows.
New PR so that Appveyor tests run.

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -259,7 +259,7 @@ class Environment(object):
 
         if not self.check_presence():
             if os.path.exists(self._path):
-                shutil.rmtree(self._path)
+                util.long_path_rmtree(self._path)
 
             if not os.path.exists(self._env_dir):
                 try:
@@ -279,7 +279,7 @@ class Environment(object):
             except:
                 log.error("Failure creating environment for {0}".format(self.name))
                 if os.path.exists(self._path):
-                    shutil.rmtree(self._path)
+                    util.long_path_rmtree(self._path)
                 raise
 
         self.save_info_file(self._path)
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -13,6 +13,7 @@ import hashlib
 import os
 import shutil
 import sys
+import subprocess
 
 import six
 
@@ -20,6 +21,9 @@ from .console import log
 from .repo import get_repo
 from . import util
 from . import wheel_cache
+
+
+WIN = (os.name == "nt")
 
 
 def iter_configuration_matrix(matrix):
@@ -244,7 +248,18 @@ class Environment(object):
             'python': self._python,
             'requirements': self._requirements
         }
+
         if info != expected_info:
+            return False
+
+        for executable in ['pip', 'python']:
+            exe = self.find_executable(executable)
+            if not os.path.isfile(exe):
+                return False
+
+        try:
+            self.run_executable('python', ['-c', 'pass'])
+        except (subprocess.CalledProcessError, OSError):
             return False
 
         return True
@@ -275,7 +290,7 @@ class Environment(object):
                     pass
 
             try:
-                self._setup()
+                self.setup()
             except:
                 log.error("Failure creating environment for {0}".format(self.name))
                 if os.path.exists(self._path):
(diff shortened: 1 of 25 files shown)


pv:
Tests pass currently; discussion on the one additional xfail is in https://github.com/spacetelescope/asv/pull/280#issuecomment-116301165

asv/plugins/conda.py
```diff
@@ -117,11 +121,26 @@ def _install_requirements(self):
...
+        """Find an executable in the environment"""
+        if WIN:
+            executable += ".exe"
+
+            exe = os.path.join(self._path, 'Scripts', executable)
+            if os.path.isfile(exe):
+                return exe
+            exe = os.path.join(self._path, executable)
+            if os.path.isfile(exe):
+                return exe
```
mdboom: Maybe move this to util or a base class, since it's repeated in virtualenv.py?

asv/util.py
```diff
@@ -346,75 +357,103 @@ def get_content(header=None):
...
+        watcher.start()
+        try:
+            stdout, stderr = proc.communicate()
+        finally:
+            if proc.returncode is None:
+                proc.terminate()
+                proc.wait()
+            watcher.join()
+
+        is_timeout = was_timeout[0]
```
mdboom: Nice to have timeout support on Windows!

test/test_environment.py
```diff
@@ -13,18 +14,21 @@
...
+WIN = (os.name == "nt")
+
+
 try:
     util.which('python2.7')
     HAS_PYTHON_27 = True
-except RuntimeError:
-    HAS_PYTHON_27 = False
+except (RuntimeError, IOError):
+    HAS_PYTHON_27 = (sys.version_info[:2] == (2, 7))
```
mdboom: This logic doesn't seem right.  `HAS_PYTHON_27` is supposed to indicate that python2.7 is available somewhere on the `PATH`, not to indicate what the present version of Python is.

pv: Yes, but if the present Python version is 2.7, then 2.7 is available even if `python2.7` is not in PATH. This is the situation when installing Python from Python.org on Windows.

pv: (Note that this comes from the new logic in `Virtualenv._find_python`)

mdboom: I see.  Sounds good.

pv:
Ok, refactored a bit.

pv:
Ok to merge?

asv/environment.py
```diff
@@ -275,18 +290,18 @@ def create(self):
...
-                    shutil.rmtree(self._path)
+                    util.long_path_rmtree(self._path)
                 raise
 
         self.save_info_file(self._path)
 
         self._is_setup = True
 
-    def _setup(self):
+    def setup(self):
```
mdboom: Why the change to public?  The original intention here was that this should only be called from `create`...

asv/graph.py
```diff
@@ -118,13 +118,15 @@ def resample_data(self, val):
 
         new_val = []
         j = 0
-        for i in xrange(min_time + step_size, max_time + step_size, step_size):
+        i = min_time + step_size
```
mdboom: Why this change?  Less likely for a change to inadvertently fail the loop end condition when using `xrange`...

mdboom:
Great to see this working on Windows!

Other than my few minor comments, I think this is ready to merge.  Thanks for this!

asv/graph.py
pv: Javascript timestamps handled here overflow the 32-bit integers on
win32, and xrange on Py2 only works with native size integers, so using
it is not possible here.

asv/environment.py
pv: Ok, reverted `setup` back to `_setup`.

asv/graph.py
```diff
@@ -116,15 +116,21 @@ def resample_data(self, val):
         max_time = max(x[0] for x in val)
         step_size = int((max_time - min_time) / RESAMPLED_POINTS)
 
+        # This loop cannot use xrange, because xrange on Python2 on
+        # 32-bit systems can only deal with 32-bit integers, and
+        # Javascript timestamps (1000*unix_timestamp) handled here
+        # overflow this range
```
mdboom: Thanks.  The comment addresses this nicely.

PR #284, opened 2015-07-06
Fix potential infinite loop in summary graph resampling + add tests

pv:
Missed a potential infinite loop in gh-282. 
Unlikely to occur in real life, but best to fix.
Merging once CI complete...

### DIFF
--- a/asv/graph.py
+++ b/asv/graph.py
@@ -116,6 +116,9 @@ class Graph(object):
         max_time = max(x[0] for x in val)
         step_size = int((max_time - min_time) / RESAMPLED_POINTS)
 
+        if step_size == 0:
+            step_size = max_time - min_time + 1
+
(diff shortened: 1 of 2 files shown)


mdboom:
Aha!  Good catch!

Crazy idea (and we can merge this after CI passes regardless) -- what about adding a "xrange_long_int" to util (or better name if you've got one) that would do what xrange does only with long integers.  Then these sorts of gotchas would be localized at least.  And I guess it could be aliased to the built-in `range` on Python 3.x?

pv:
That's could be possible too. I guess it's would be sensible, at least if there's another place that needs it.

(In the case here, raising an error for `step_size == 0` would also be OK, which is what xrange would do, since triggering this requires exactly 384 datapoints evenly spaced within 384 ms --- can't actually occur since the git/mercurial timestamp granularity is 1000 ms.)

PR #286, opened 2015-07-14
WIP: Add coveralls support

mdboom:


mdboom:
@pv: Any objections to this?  The coverage isn't great right now, but I think I'll work on actually improving the coverage in subsequent PRs.

mdboom:
One of the issues with this is that full coverage isn't really possible without combining multiple Travis configurations (conda vs. virtualenv, for example).  Will need to investigate if it's possible to combine reports and upload together.

pv:
Looks good to me.

PR #289, opened 2015-07-18
Always call benchmark script with absolute path + actually test this

pv:
This wasn't caught previously, because all tests used absolute paths in the config.

See https://github.com/spacetelescope/asv/commit/7564627ba4af275d0d753fefd22e5fe80a3c75e5#commitcomment-12233571


mdboom:
Thanks.  Makes sense.

Issue #287, opened 2015-07-14
Web tests intermittently timing out


mdboom:
It's not every time.  See here for an example:

https://travis-ci.org/spacetelescope/asv/jobs/69812435

Any thoughts @pv?

pv:
I'm not sure it's only Python 3.4, it might have also occurred previously with other versions. Haven't been able to reproduce this on my machine.

The failure appears to be before it has run the tests (no dots in output). Possibly hangs during Selenium initialization --- in which case it's Selenium/PhantomJS bug, probably race condition.

There's however a bug in the tests in that the scope="session" specifier is in the wrong place. That's fixed in gh-288 but I don't think that is related to this issue.

mdboom:
So still a bit of an engima.  Just for more data: I noticed a failure on Python 2.7 where the first dot after the `test_web.py` is printed.

https://travis-ci.org/spacetelescope/asv/jobs/70935853

pv:
Yes, this is probably not easy to weed out...

The first dot being printed in the above job doesn't yet rule out hang during Selenium initialization --- prior to gh-288 the fixture was initialized before each test. Having session-scope may alleviate the issue slightly, however.

pv:
Potentially related: https://github.com/travis-ci/travis-ci/issues/3251

pv:
Looking at the most recent failure in https://travis-ci.org/spacetelescope/asv/jobs/71682218
it again stalled after outputting one dot. Since this is with single-session selenium, and after fixture init and before shutdown, it then probably either hangs on page load in `browser.get`, or in `httpd.shutdown` / `thread.join` for the preview http server. It seems this can be triggered locally by running multiple instances of the tests in parallel, and it appeared the hang in that case is in condition wait --- probably the condition inside `httpd.shutdown` that blocks until `serve_forever` exits. IOW, the Python builtin socketserver seems to fail to terminate for an unclear reason (waiting on socket?) sometimes.

gh-290 plugs these cases by changing the threads to daemon threads, which are shutdown on program exit, and using timed-out joins. Also, it sets timeouts to selenium (afaik, the default is int_max). So hopefully that fixes the issue finally...

pv:
The tests still seem to fail, but with the changes in gh-290 at least a traceback was produced:
https://travis-ci.org/spacetelescope/asv/jobs/75030204

On the browser side:
```
/home/travis/build/spacetelescope/asv/test/test_web.py:71: 
>           get_with_retry(browser, base_url)
[...]

PR #290, opened 2015-07-18
Add page load timeouts to web tests

pv:
Some googling shows other people are having similar problems:
https://www.google.fi/search?q=selenium+hang+travis

This one contains a potential solution --- set up timeouts, and just retry on failure: https://stackoverflow.com/questions/29108260/geb-selenium-tests-hang-loading-new-page

Related to gh-287

**EDIT**: also avoid shutting down the preview web server in a way that can block. I managed to reproduce a hang at this point, so best to do this defensively. I'm not sure, but perhaps phantomjs kept connections alive sometimes and prevented the server from shutting down; at the same time, the main process cannot send new instructions to phantomjs because it's waiting for the http server to exit -> deadlock (until socket timeout)?

### DIFF
--- a/test/test_web.py
+++ b/test/test_web.py
@@ -19,7 +19,7 @@ from asv.commands.run import Run
 from asv.commands.publish import Publish
 
 from . import tools
-from .tools import browser
+from .tools import browser, get_with_retry
 from .test_workflow import basic_conf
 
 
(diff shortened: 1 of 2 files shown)


pv:
Going to merge, this one should be OK.

PR #295, opened 2015-07-21
Fix #294: Handle global commandline args

### LINKED ISSUE #294: Document global commandline options in docs
The options available to all commands are not documented in the generated documentation.

mdboom:
So they appear in `asv $COMMAND --help`

### DIFF
--- a/asv/commands/__init__.py
+++ b/asv/commands/__init__.py
@@ -11,7 +11,6 @@ import six
 
 from .. import config
 from .. import util
-from .. import __version__
 
 from . import common_args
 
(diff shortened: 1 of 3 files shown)


pv:
argparse seems to be clumsy here. Note that `asv --config=foo run`
becomes disallowed here. There doesn't seem to be a way to allow both
places; the parent parser trick mentioned on some web pages is somewhat
wrong, doesn't work for action='store_true'. Other pages suggest not
add_subparsers at all, but that seems also clumsy.
.
The other option would be not fighting against argparse, including `asv
--help` output on the documentation page, and using "add_parser('run',
prog='asv [global options] run', ...)" to indicate to the user that
there are some global options...
.
This change breaks some of my scripts, because `asv --config=foo run`
becomes disallowed. Not a big deal in itself, though.

pv:
There's a way to do it, actually: https://github.com/pv/asv/tree/pr-295
as copied from https://bugs.python.org/issue24251
(Note that travis didn't run for 746c0bd --- there's also a missing import).

PR #298, opened 2015-08-10
Fix bug parsing revisions in 'asv compare'

qwhelan:
@mdboom I noticed a bug when attempting to run ```asv compare``` on the pandas asv suite. I was seeing the following issue:

```
$ asv compare e118206d 912854 
All benchmarks:

    before     after       ratio
  [e       ] [9       ]
   310.58ns   311.98ns      1.00  attrs_caching.getattr_dataframe_index.time_getattr_dataframe_index
     3.66μs     3.75μs      1.03  attrs_caching.setattr_dataframe_index.time_setattr_dataframe_index
      [...]
```

Note that the before and after hashes have been truncated to the first character. The essence of this bug is that ```Compare.run_from_conf_args()``` has some custom logic that wasn't being covered by tests. Changing the tests to call this function confirms that the ```[0]```s are unnecessary.

### DIFF
--- a/asv/commands/compare.py
+++ b/asv/commands/compare.py
@@ -84,8 +84,8 @@ class Compare(Command):
     @classmethod
     def run_from_conf_args(cls, conf, args):
         return cls.run(conf=conf,
-                       hash_1=args.revision1[0],
-                       hash_2=args.revision2[0],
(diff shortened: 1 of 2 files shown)


pv:
Thanks, merged.

PR #302, opened 2015-08-11
Add utility function to enforce hash prefix uniqueness

### LINKED ISSUE #299: asv does not check if hash prefix is unique
@mdboom I noticed that ```asv``` isn't checking if the short hash is a unique benchmark identifier while investigating #298. This is primarily an issue in ```asv compare``` as the code averages over all benchmarks it finds which means, when combined with the bug described in #298, the first column will contain the average result for *all commits that start with 'e'* and the second will have *all commits that start with '9'*.

Given that the prefix of hash implies nothing about the relationship between the content, I believe the appropriate action would be to raise an ```Exception``` if the prefix is not unique. I'm working on a patch to address this but wanted to confirm that this approach would be more appropriate.

qwhelan:
@mdboom @pv Resolves #299 by adding a utility function to resolve a hash prefix to the "full" hash.

### DIFF
--- a/asv/results.py
+++ b/asv/results.py
@@ -50,11 +50,12 @@ def iter_results_for_machine_and_hash(results, machine_name, commit):
     Iterate over all of the result files with a given hash for a
     particular machine.
     """
+    full_commit = get_result_hash_from_prefix(results, machine_name, commit)
+
     for (root, filename) in iter_results_paths(
             os.path.join(results, machine_name)):
         results_commit = filename.split('-')[0]
-        cmp_len = min(len(commit), len(results_commit))
-        if results_commit[:cmp_len] == commit[:cmp_len]:
+        if results_commit == full_commit:
             yield Results.load(os.path.join(root, filename))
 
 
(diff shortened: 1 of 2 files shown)


pv:
Was this closed accidentally?

qwhelan:
@pv Yeah, I was trying to retrigger the Travis run as a bunch of web tests randomly failed.

pv:
Yes, there's something strange with phantomjs on travis, and the conda environment creation also seem to fail quite often due to web connectivity issues...

qwhelan:
@pv Looks like it's all passing now. It looks like Travis has some caching infrastructure that might help cut down on the network failures, which I'll try to look into this evening.

asv/results.py
```diff
@@ -80,6 +81,34 @@ def get_existing_hashes(results):
     return hashes
 
 
+def get_result_hash_from_prefix(results, machine_name, commit_prefix):
+    """
+    Get the 8-char result commit identifier from a potentially shorter
```
mdboom: Instead of 8-char, don't you mean "full"?

qwhelan: @mdboom Nope, the hash length is explicitly truncated to 8-chars below

pv:
LGTM.

Issue #303, opened 2015-08-19
asv failing with OSError


jreback:
just merged ``asv`` support into pandas: https://github.com/pydata/pandas/pull/9715
doing the back-benchmarks, and running into an odd issue. This is does reproduces (I initially tried on 0.1.1, now on latest master of asv)

```
[ 34.82%] ··· Running groupby.groupby_transform_multi_key4.time_groupby_transform_multi_key4Traceback (most recent call last):
  File "/Users/jreback/anaconda/bin/asv", line 9, in <module>
    load_entry_point('asv==0.2.dev762-b661eaef', 'console_scripts', 'asv')()
  File "/Users/jreback/anaconda/lib/python2.7/site-packages/asv-0.2.dev762_b661eaef-py2.7.egg/asv/main.py", line 36, in main
    result = args.func(args)
  File "/Users/jreback/anaconda/lib/python2.7/site-packages/asv-0.2.dev762_b661eaef-py2.7.egg/asv/commands/__init__.py", line 47, in run_from_args
    return cls.run_from_conf_args(conf, args)
  File "/Users/jreback/anaconda/lib/python2.7/site-packages/asv-0.2.dev762_b661eaef-py2.7.egg/asv/commands/run.py", line 122, in run_from_conf_args
    skip_existing_commits=args.skip_existing_commits
  File "/Users/jreback/anaconda/lib/python2.7/site-packages/asv-0.2.dev762_b661eaef-py2.7.egg/asv/commands/run.py", line 275, in run
    profile=profile, skip=skipped_benchmarks)
  File "/Users/jreback/anaconda/lib/python2.7/site-packages/asv-0.2.dev762_b661eaef-py2.7.egg/asv/benchmarks.py", line 529, in run_benchmarks
    cwd=tmpdir)
  File "/Users/jreback/anaconda/lib/python2.7/site-packages/asv-0.2.dev762_b661eaef-py2.7.egg/asv/benchmarks.py", line 108, in run_benchmark
    cwd=cwd)
  File "/Users/jreback/anaconda/lib/python2.7/site-packages/asv-0.2.dev762_b661eaef-py2.7.egg/asv/benchmarks.py", line 236, in _run_benchmark_single
    valid_return_codes=None, cwd=cwd)
  File "/Users/jreback/anaconda/lib/python2.7/site-packages/asv-0.2.dev762_b661eaef-py2.7.egg/asv/plugins/conda.py", line 125, in run
[...]

PR #304, opened 2015-08-19
Catch killpg failing with EPERM on already terminated group on OSX

pv:
Linux and BSD systems (including OSX) appear to behave differently on whether sending signals to zombie process groups is an error or not. In both cases the end result is the same (the signal has no effect), so we can just ignore the error raised here.

Fixes gh-303

See also https://stackoverflow.com/questions/12521705/why-would-killpg-return-not-permitted-when-ownership-is-correct

### DIFF
--- a/asv/util.py
+++ b/asv/util.py
@@ -387,7 +387,7 @@ def check_output(args, valid_return_codes=(0,), timeout=120, dots=True,
                 # process is in a separate process group so it won't receive
                 # these automatically from the terminal
                 def sig_forward(signum, frame):
-                    os.killpg(proc.pid, signum)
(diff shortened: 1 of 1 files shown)


pv:
Merging, straightforward bugfix. (appveyor failure is unrelated since this change only affects POSIX systems.)

mdboom:
I can't think of an obvious way to test this, but if there is one, maybe we should...

pv:
Didn't manage to get ETERM from killpg on latest OpenBSD and FreeBSD...
Would probably need OSX for properly reproduce etc., but I don't have
access.

Issue #305, opened 2015-08-20
Loading summary page is very slow


pv:
Pandas has 600+ benchmarks, and loading up the summary html page takes up 5+ seconds, which is somewhat too slow. It's probably is possible to speed the JS up on this point.

The summary page probably also should not be generated if you follow links directly to individual graphs (e.g. from the regressions view).

PR #306, opened 2015-08-21
Refactor and speed up JS

pv:
Refactoring:

- move summary graph page to `summary.js`
- move benchmark graph pages to `graphdisplay.js`

Optimizations:

- lazy loading, summary and benchmark graph pages are only initialized if used
- use `display:none` instead of `.toggle()` in benchmark tree construction (major speedup, 2.5 sec -> 0.3 sec for pandas)
- fix up summary page loading all graphs at once in some cases (major speedup, 3 sec -> 0 for pandas in some cases)
- add elements to DOM only after construction (minor speedup, avoids restyling)
- cache most recent graph data on JS side

Fixes gh-305

### DIFF
--- a/asv/www/asv.js
+++ b/asv/www/asv.js
@@ -181,8 +181,12 @@ $(document).ready(function() {
     var window_scroll_positions = {};
     /* Previous window hash location */
     var window_last_location = null;
+    /* Whether benchmark graph display was set up */
+    var benchmark_graph_display_ready = false;
 
     function display_benchmark(bm_name, state_selection, sub_benchmark_idx, highlight_timestamps) {
+        setup_benchmark_graph_display();
+
         $('#graph-display').show();
         $('#summary-display').hide();
         $('#regressions-display').hide();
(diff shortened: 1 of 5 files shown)


mdboom:
> fix up summary page loading all graphs at once in some cases (major speedup, 3 sec -> 0 for pandas in some cases)

Just out of curiosity, can you elaborate on this point?  The original version did try to take care not to load all summary graphs at once.

pv:
It somehow happens on Firefox & Chromium that if you scroll the summary
page a bit down, and then click on a summary graph, scroll events are
triggered for all of the graphs (and this was visible on the network
logs). This does not seem to happen if you don't scroll the summary page
first.
.
You should be able to reproduce this at
http://www.astropy.org/astropy-benchmarks/ (in firefox, enable XHR
request logs in the console).
.
I do not understand why this happens, however.

mdboom:
Ah -- I see that, too.  Yet another Javascript/HTML mystery, I suppose.  The fix seems fine -- I was just curious, really.

Other than that, the refactor looks good to me.  :+1: from me to merge.

pv:
This is dogfooded at https://pv.github.io/scipy-bench/ and seems to work

PR #310, opened 2015-08-23
Instead of replacing, take minimum of new and old results

pv:
Add option to `asv run` for taking minimum of new and old results, rather than just replacing old with new. May be useful for reducing scatter in some cases.

(Requested by @juliantaylor)

### DIFF
--- a/asv/commands/compare.py
+++ b/asv/commands/compare.py
@@ -48,11 +48,6 @@ def unroll_result(benchmark_name, result):
         yield name, result
 
 
-def _isna(value):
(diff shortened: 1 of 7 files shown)


juliantaylor:
neat
another nice addition would be to only rerun those benchmarks that have a high scatter.
for vbench we used the simple method of each benchmark having a counter and each time a new minimum is found it is reset to zero. benchmarks with a high enough counter are not rerun.

pv:
Needs a bit more thought, then.

juliantaylor:
why?

PR #311, opened 2015-08-23
CI: try to get rid of conda timeouts

pv:
Try to get rid of conda timeouts by replacing the conda command with one that always uses `travis_retry` instead.


pv:
Straightforward CI change, so merging.

pv:
Heh, of course there's a mistake when you merge fast...

`conda --yes update conda` probably overwrites the replaced conda script, so it should be above the replacement. Fixing ...

PR #312, opened 2015-08-24
CI: replace conda with retry script only after updating (which may overwrite)

pv:
Take 2 in using travis_retry for all conda commands on Travis-CI.


pv:
Merging, should be OK this time...

PR #314, opened 2015-08-25
No mirror for local repos

pv:
Local repositories don't really need a mirror repository, as the performance will be OK anyway.
In addition, this allows referring to HEAD, which is useful for "repo": "." configs.

In order to preserve backward compat, make checking out working trees more robust, by trying to re-create the work tree repo if it appears to be malfunctioning.

### DIFF
--- a/asv/plugins/git.py
+++ b/asv/plugins/git.py
@@ -88,12 +88,23 @@ class Git(Repo):
         self._pulled = True
 
     def checkout(self, path, commit_hash):
+        def checkout_existing(display_error):
+            self._run_git(['checkout', '-f', commit_hash], cwd=path,
+                          display_error=display_error)
+            self._run_git(['clean', '-fdx'], cwd=path,
+                          display_error=display_error)
+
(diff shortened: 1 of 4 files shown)


asv/plugins/git.py
```diff
@@ -19,20 +19,26 @@
...
-        if not os.path.isdir(self._path):
-            args = ['clone']
-            if _checkout_copy:
-                args.append('--shared')
-            else:
-                log.info("Cloning project")
-                args.append('--mirror')
-            args.extend([url, self._path])
-            self._run_git(args, chdir=False)
+        if Git.is_local_repo(url):
```
mdboom: Any reason to do this rather than `self.is_local_repo`...?  This would seem to make subclassing harder.

mdboom:
:+1:

asv/plugins/git.py
pv: Oh, that works for classmethods, will change.

pv:
Ok, merging.

PR #315, opened 2015-08-29
Use threaded HTTP server in asv preview

pv:
Switch `asv preview` to use a threaded HTTP server.

This also switches the tests to use the threaded server. This may possibly help with the timeouts on Travis-CI, since the threaded server does not get "stuck" if one request fails to complete.

### DIFF
--- a/asv/commands/preview.py
+++ b/asv/commands/preview.py
@@ -96,5 +96,8 @@ class Preview(Command):
             import webbrowser
             webbrowser.open(base_url)
 
-        log.info("Press ^C to abort")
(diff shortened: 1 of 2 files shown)


pv:
Planning to merge soon.

mdboom:
Looks fine to me.  Not sure what the Python 3 Windows failure is about, though.

pv:
The appveyor timeout failure appears to be occurring randomly, and has occurred also for previous commits. Not sure about what causes it, but it seems to always get stuck at `test_workflow`.
Probably unrelated to this PR.

PR #316, opened 2015-09-03
Run benchmark setup on every repeat, similarly as timeit does

pv:
The benchmark setup should be run on every repeat, this is what users probably expect.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,10 @@ New Features
 
 API Changes
 ^^^^^^^^^^^
+Previously, the ``setup`` and ``teardown`` methods were run only once
+even when the benchmark had ``repeat > 1`` specified. This is now
+changed so that they are run on every repeat, and additionally before
+and after profiling runs.
 
 - Mirrors are no longer created for local repositories. (#314)
 - The parent directory of the benchmark suite is no longer inserted
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,10 +11,6 @@ New Features
 
 API Changes
 ^^^^^^^^^^^
-Previously, the ``setup`` and ``teardown`` methods were run only once
-even when the benchmark had ``repeat > 1`` specified. This is now
-changed so that they are run on every repeat, and additionally before
-and after profiling runs.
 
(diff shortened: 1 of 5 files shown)


mdboom:
Where is this behavior defined in Python's `timeit`?  I could look at the code, I guess, but unless I'm missing something, the docs seem to be quiet on this point.

I, personally, would only expect setup to be run once ever.  Is there a use case where it matters?

pv:
That's how timeit works, but it doesn't appear to be defined in the documentation.
It was assumed also asv works like this in https://github.com/pydata/pandas/pull/10927

mdboom:
Ah, I see now why that matters.

Can you also update the `setup and teardown functions`?  Particularly, the statement that "teardown is rarely necessary, since each benchmark runs in its own process" is no longer accurate: if setup is run multiple times, it's more likely one would also need a teardown.

"The setup function is run for each benchmark" is also somewhat misleading, because it may be run multiple times.

It should also be clarified how this interacts with setup_cache.

mdboom:
This should also probably get a note in "bug fixes" and "API changes" in CHANGES.

pv:
Updated docs

pv:
It might make sense actually to move the repeat handling out from `benchmark.py` to the main ASV side. Then, we could do something similar to gh-310 but smarter.

CHANGES.rst
```diff
@@ -6,6 +6,10 @@ New Features
 
 API Changes
 ^^^^^^^^^^^
+Previously, the ``setup`` and ``teardown`` methods were run only once
+even when the benchmark had ``repeat > 1`` specified. This is now
+changed so that they are run on every repeat, and additionally before
+and after profiling runs.
```
mdboom: But the behavior with profiling runs is unchanged, correct?  If so, maybe it's less confusing to not mention it.

pv: The profiling runs previously redo the setup before profiling (the benchmark is run once before the profiling).

pv:
Anything more that should be changed?

mdboom:
This looks good.  Needs a rebase and retest, and then I think it's ready to merge.

pv:
Rebased.

Issue #319, opened 2015-09-05
Windows timeout is not timeout from last output


pv:
On windows, the `timeout` has a different meaning from unix currently, as it's timeout from start and not timeout since last output. The behavior should be made to match.

jorisvandenbossche:
I just wanted to report that on Windows, building pandas times out (as it takes longer than 2 minutes, and manually increasing `timeout` in util.py solves it), but I assume the above is the cause of this?

pv:
Yes, fairly sure that's the cause.

The easiest solution appears to be to move the blocking reads to threads reading 1 byte at a time.
Other option would be to use directly win32 async ("overlapped") i/o.

pv:
Ok, MSDN states async io is not supported on anonymous pipes, so gh-320 is the only way to do this (short of rewriting subprocess module or using temporary files).

PR #320, opened 2015-09-05
Make timeouts also on Windows based on last output timestamp

pv:
I don't have a Windows machine at hand (only WINE), so this isn't fully tested, but should work.

Fixes gh-319

cc @jorisvandenbossche

### DIFF
--- a/asv/util.py
+++ b/asv/util.py
@@ -359,25 +359,50 @@ def check_output(args, valid_return_codes=(0,), timeout=120, dots=True,
     is_timeout = False
 
     if WIN:
-        start_time = time.time()
+        start_time = [time.time()]
         was_timeout = [False]
+
+        def stdout_reader_run():
+            while True:
+                c = proc.stdout.read(1)
+                if not c:
+                    break
+                start_time[0] = time.time()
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Yes, this seems to work!

mdboom:
Makes sense to me (though I don't have Windows to test on, either)

pv:
Probably safe to merge, seems simple in principle.

seth-p:
FYI, using the latest code from master (0.2.dev818+c07fdbf2) I get a timeout after a bit under 4 minutes when running the Pandas asv tests (64-bit Python 3.4.3 on Windows 7). https://github.com/pydata/pandas/pull/9023#issuecomment-148934789

pv:
Perhaps the compilation of `algos.c` just takes more than 120 seconds --- can probably be checked by running the command shown in the log manually?

However, 2 min timeout is probably too short for compilation, and it's not handled correctly anyway (asv quits with error rather than dealing with it).

seth-p:
Perhaps one should be able to specify/override the timeouts (I recall seeing them in three places in the code) on the command line and/or confit file?

Issue #309, opened 2015-08-22
Summary graph doesn't look nice with branches


pv:
The summary graph doesn't know anything about branches, and will display all commits in date order.

If there are branches with active commits and differing performance, this will generate spurious "spikes" as the summary shows the commit performance in date order.

The mechanism that's used for averaging and filling-in results for different parameter combinations might also work for averaging over branches.

PR #321, opened 2015-09-05
Make summary graphs by taking filled-in geometric averages also over run params

pv:
Previously, for graphs run over different run params (eg. computer,
branch) the averaging was simple arithmetic mean without fill-in of
missing values. This introduces spurious spikes in the summary graphs,
which is now avoided using the same geom-mean code for run and benchmark
parameters.

Should address gh-309


mdboom:
:+1:

Issue #322, opened 2015-09-06
Specifying requirements depending on environment/OS


jorisvandenbossche:
In trying to have an asv.conf.json file to be generic so that contributors don't have to adapt it, we have some problems in pandas.

As the required dependencies can in some cases depend on:
- OS: eg for building pandas on Window, there is an extra dependency when using conda (`libpython`)
- environment type: eg pytables is called tables for pip but pytables for conda. (which is more an issue with conda, but we have to deal with it)

It is just a question if this would be within scope to deal with such things. But I can understand the answer would quite possibly be: no, each contributor will have to adapt it to work for his/her machine.

E.g with conda yaml files, you can have lines to be specific for a certain OS using a kind of preprocessed selectors: http://conda.pydata.org/docs/building/meta-yaml.html#preprocessing-selectors

pv:
I think this should be supported --- `asv.conf.json` is often checked in to vcs, so that it should be relatively static. Additionally, it would be useful to be able to override the specified environment type for example on command line.

This could along the lines:
```
"matrix": {
    "[environment=conda,platform=win32]libpython": [],
    "[python=3.*]six": [],
}
```
and possibly more generic syntax could be allowed,
```
"[platform=win32]matrix": {
    "libpython": [],
    "six": [],
},
"[platform=win32]pythons": ["2.7"],
"[platform=linux*]pythons": ["2.7", "3.4"],
```

mdboom:
Yeah -- the whole dependency matrix thing is pretty minimal -- it's just the simplest thing to get something off the ground.  I think any changes we make to the "language" should also try to consider the other major shortcoming, namely no way to include or exclude specific combinations.  Travis-CI is fairly powerful in this regard.

I'm not sure I'm crazy about the additional metalanguage inside of dictionary keys, though.  I'd prefer something that was still parseable just from the tree of JSON objects.

pv:
How about this? https://github.com/spacetelescope/asv/pull/329

This enables specifying platform/environment/python-dependent items in the requirement matrix, although you have to do it "backwards" (specify in matrix, but exclude non-applicable entries).
This may or may not be too clumsy for practice, comments welcome.

pv:
gh-329 is probably ready now, assuming the approach is fine.

For Pandas, the config would read:
```
    "matrix": {
        "numpy": [],
        "Cython": [],
        "matplotlib": [],
        "sqlalchemy": [],
        "scipy": [],
        "numexpr": [],
[...]

PR #329, opened 2015-10-14
exclude/include rules in config

pv:
Exclude/include rules in configuration.

The present approach is sort of a ripoff from how Travis-CI does it.

Addresses gh-322

- [x] fix bugs
- [x] `environment_type` and `sys_platform` perhaps should be allowed in *include*, but act as filters, similarly as in *exclude*? This would allow easy specification of os/environment dependent rules.


mdboom:
Thanks for this.  I agree that ripping off Travis-CI makes a lot of sense -- for familiarity's sake if nothing else.

mdboom:
I really like this.  Very cool.

PR #332, opened 2015-10-17
Fail early if mirror directory is needed, but exists and is not a repository

pv:
Print a friendlier error message if the mirror directory already exists.

One very common case where this occurs is when user puts asv.conf.json on the project top level but uses a remote URL in *repo*. Python packages usually have a directory called *project* on the top level.


pv:
Should be straightforward, will merge soon.

Issue #334, opened 2015-10-18
console.py calling write() with bytes rather than str


seth-p:
```
C:\Users\seth\github\pandas\asv_bench>asv continuous master HEAD
· Creating environments
· Discovering benchmarks
·· Uninstalling from py3.4-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-scipy-sqlalchemy-xlrd-xlwt
·· Building for py3.4-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-scipy-sqlalchemy-xlrd-xlwt
·· Installing into py3.4-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-scipy-sqlalchemy-xlrd-xlwt
· Running 1320 total benchmarks (2 commits * 1 environments * 660 benchmarks)
[  0.00%] · For pandas commit hash cd9c777f:
[  0.00%] ·· Building for py3.4-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-scipy-sqlalchemy-xlrd-xlwt
[  0.00%] ·· Benchmarking py3.4-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-scipy-sqlalchemy-xlrd-xlwt
[  0.08%] ··· Running attrs_caching.getattr_dataframe_index.time_getattr_dataframe_index
[  0.15%] ··· Running attrs_caching.setattr_dataframe_index.time_setattr_dataframe_indexTraceback (most recent call last):
  File "C:\Python34\Scripts\asv-script.py", line 9, in <module>
    load_entry_point('asv==0.2.dev818+c07fdbf2', 'console_scripts', 'asv')()
  File "C:\Python34\lib\site-packages\asv\main.py", line 36, in main
    result = args.func(args)
  File "C:\Python34\lib\site-packages\asv\commands\__init__.py", line 48, in run_from_args
    return cls.run_from_conf_args(conf, args)
  File "C:\Python34\lib\site-packages\asv\commands\continuous.py", line 49, in run_from_conf_args
    **kwargs
  File "C:\Python34\lib\site-packages\asv\commands\continuous.py", line 73, in run
    _machine_file=_machine_file)
  File "C:\Python34\lib\site-packages\asv\commands\run.py", line 276, in run
    profile=profile, skip=skipped_benchmarks)
  File "C:\Python34\lib\site-packages\asv\benchmarks.py", line 529, in run_benchmarks
    cwd=tmpdir)
[...]

PR #336, opened 2015-10-18
Fix encoding-fallback printing on Py3

pv:
Take into account that codecs.StreamWriter require a byte stream to
write to, and that the resulting writer should be fed unicode.

Fixes gh-334


pv:
This code is a bit unclear on the text vs bytes issue, and could maybe use some cleanup later on...

pv:
Should be straightforward, merging.

PR #340, opened 2015-10-18
Fix documentation for the --python option

pv:
The default value differs for asv run and asv dev.  Moreover, currently
only python=same results to use of ExistingEnvironment, so it's not
possible to pass in python executable names any more.

### DIFF
--- a/asv/commands/common_args.py
+++ b/asv/commands/common_args.py
@@ -66,21 +66,28 @@ def add_machine(parser):
 
 
 def add_python(parser, default=None):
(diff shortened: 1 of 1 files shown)


mdboom:
Seems fine.  In the future, would you mind leaving PRs up a little longer so I have a change to review (particularly on the weekend...)

pv:
Hi, sorry, I'll avoid self-merging this fast in future.
There's of course no real hurry with this...

PR #348, opened 2015-11-30
Support binary files in repository template

mdboom:
I came across this bug only because the `test` directly accidentally became populated with `.pyc` files, which obviously don't decode as utf-8.  It isn't strictly needed (since we don't have binary files in the repository template), but we might someday.


pv:
LGTM

PR #350, opened 2015-11-30
Fix #346: Display errors when setup_cache fails

### LINKED ISSUE #346: Failed setup_cache doesn't produce any errors
When `setup_cache` fails, no error is produced to the console.

Run this benchmark (from some .py file):
```
class Foo:
	def setup_cache(self):
		assert False

	def track_me(self, cache):
		return 0
```
with the command `asv dev -b Foo` and you get this output:
```
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[  0.00%] ·· Building for D:\Anaconda3\python.exe
[  0.00%] ·· Benchmarking D:\Anaconda3\python.exe
[  0.00%] ··· Setting up D:\benchmarks\bug.py:69
```

System info: Win7 64bit, Python 3.4, Anaconda, conda 3.18.5, asv 0.2.dev839+7b43616e.

mdboom:


asv/benchmarks.py
```diff
@@ -516,9 +516,14 @@ def run_benchmarks(self, env, show_stderr=False, quick=False, profile=False,
                             return_stderr=True, valid_return_codes=None,
                             cwd=tmpdir)
                         if errcode:
+                            # Dump program output
```
pv: Could maybe also dump stdout, `err += out`

PR #351, opened 2015-11-30
Fix #347: Use dill instead of pickle if available

### LINKED ISSUE #347: Consider using dill instead of pickle for setup_cache
From the [Dill pypi page](https://pypi.python.org/pypi/dill):

> Dill extends python’s ‘pickle’ module for serializing and de-serializing python objects to the majority of the built-in python types.
> So dill is useful if the data-types generated in `setup_cache` are not supported by pickle.

I think that this change only requires adding dill as a dependency and then replacing `import pickle` by `import dill as pickle`.

mdboom:


thisch:
You might be interested in cloudpickle (https://github.com/blaze/dask/issues/847)

mdboom:
@thisch, @yoavram:  I must admit, I don't have much experience with either `dill` or `cloudpickle`, and from reading up, I can't say there's a clear winner in that space.

For `asv`, I'd say the priority is the completeness and correctness in the number of data types supported.  Efficiency in space or time is pretty secondary, since these pickles are all just used to cache to local disk.  Does that help make one option stand out over the other?  Or is this best made as a user configuration option so they can choose which pickle implementation to use?

yoavram:
First time I hear about cloudpickle, sounds good to me.
You don't want to include a test in this PR? Something that fails with pickle and succeeds with dill?

mattip:
I am going to close this since the motivation is unclear and the PR has conflicts. If you wish to reopen and explain why an alternative to pickle is needed, please do.

PR #352, opened 2015-12-09
Add a command-line option for selecting environments

pv:
Add a command-line option `--environment` (or `-E`) for selecting what environment type and Python version to run, overriding what is specified in the configuration file.

- This is useful e.g. when the configuration file is checked in a repository, and run by multiple users, some of which have conda and some not.
- Supercedes the `--python` option, and makes it possible to specify what Python executable to run using ExistingEnvironment (which was not possible previously).
- Add `tool_name` to the hash-name of the environments, so that it's possible to have conda and virtualenv at the same time without conflicts.
- Move calls to `environment.get_environments` to the top-level command classes. (Previously, environments were instantiated in `Benchmarks` for benchmark discovery.)

Some fine tuning may still be needed (although this is in principle complete)

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -28,23 +28,30 @@ from . import wheel_cache
 WIN = (os.name == "nt")
 
 
-def iter_requirement_matrix(conf):
+def iter_requirement_matrix(environment_type, pythons, conf,
+                            process_includes=True):
     """
     Iterate through all combinations of the given requirement
     matrix and python versions.
     """
 
-    if not conf.environment_type:
-        env_classes = get_environment_classes(conf)
+    env_classes = {}
+
+    def get_env_type(python):
+        env_type = env_classes.get(python)
+        if env_type is None:
+            env_type = get_environment_class(conf, python)
+            env_classes[python] = env_type
+        return env_type
 
     platform_keys = {
-        'environment_type': conf.environment_type,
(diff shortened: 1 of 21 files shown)


asv/environment.py
```diff
@@ -311,7 +359,8 @@ def hashname(self):
         """
         Get a hash to uniquely identify this environment.
         """
-        return hashlib.md5(self.name.encode('utf-8')).hexdigest()
+        full_name = "{0}-{1}".format(self.tool_name, self.name).encode('utf-8')
```
mdboom: Would it make sense to just have tool_name included as part of the name property -- since tool_name is now required to uniquely identify the environment...

mdboom:
Looks good, other than my one comment above.

mdboom:
@pv: Sorry I've been incommunicado here for so long.  Want to fix the minor issue, rebase and I'll merge?

pv:
No problem, updated. The environment tool name is now included also in the result file names.

pv:
@mdboom: can you enable rolling builds http://www.appveyor.com/docs/build-configuration#rolling-builds in appveyor, so that it behaves more sensibly vs. updated PRs?

mdboom:
Merged.

I also enabled rolling builds on appveyor.

PR #354, opened 2015-12-13
BUG: fix graph file name generation for default requirement version

pv:
The meaning of requirement equal to "" and null was changed in
b9bdf379e5, but the Javascript code was not updated so graphs won't
load.

Sync the javascript side with Python.

Extend the web tests to run with null and "" requirements, to cover this.


pv:
Going to merge soon, as master is broken without this...

PR #359, opened 2015-12-17
Support python 3 for command "asv quickstart"

edisongustavo:


asv/commands/quickstart.py
```diff
@@ -6,6 +6,7 @@
 
 import os
 import shutil
+import six
```
pv: Maybe simpler to do just `from six.moves import input as raw_input`.

edisongustavo: Nice, didn't know about that! Pushing change now

pv:
thanks, merged

PR #363, opened 2015-12-31
ENH: add support for requirements files with virtualenv

johntyree:
I was fiddling with a project that uses `pip install -r requirements.txt` rather than `install_requires` in `setup.py` and came up with the following extension to the `virtualenv` plugin.

Is the "right" way to achieve this? I don't like the overloading of `environment.build_project`, but this needs to appear between VCS checkout and the call to `setup.py`, which both happen therein.

I'll write some tests if this looks mergeable.

### DIFF
--- a/asv/plugins/virtualenv.py
+++ b/asv/plugins/virtualenv.py
@@ -48,6 +48,7 @@ class Virtualenv(environment.Environment):
         self._executable = executable
         self._python = python
         self._requirements = requirements
+        self._requirements_file = getattr(conf, 'requirements_file', None)
         super(Virtualenv, self).__init__(conf, python, requirements)
 
         try:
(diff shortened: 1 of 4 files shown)


pv:
The implementation looks ok to me, best to be pragmatic. Sometimes people have multiple requirement files --- does it need to be a list of files, or did pip have some #include convention?

I'm wondering about the overlap between `matrix` and `requirements_file` / `install_requires` --- the latter two make the environment depend on the version of the project benchmarked. Probably the difference should be pointed out in the docs.

johntyree:
Yes, pip supports nesting files so you only need one

Good point about the docs. I'll take a look at that.

mdboom:
@johntyree: Want to add documentation to get this over the threshold?

johntyree:
@mdboom bandwidth is tight at the moment. It's on my radar though.

johntyree:
I didn't actually see any kind of tests in place for this stuff. Is the `virtualenv` tool tested? It looks like there are some generic tests that will default to using conda (for which `requirements_file` will do nothing).

pv:
Travis runs the tests both with and without conda. It could be better to organize the tests a bit differently, but that hasn't been done so far.

pv:
It's probably reasonably straightforward to add a test for this feature, probably should go in `test_workflow.py`

johntyree:
Aha. OK thanks. If that's something you think you could do in short order, please feel free. Otherwise this is on hold for awhile longer until I have time to look at how these tests work.

datapythonista:
We'll move towards removing environment support, so not worth adding this. See #1031

PR #373, opened 2016-02-26
ENH: fallback to pip if matrix dependency is not available with conda

andyfaff:
If the `environment_type` is `conda`, then the only matrix dependencies that can be examined are those available via `conda install`.  This is annoying if some, or all, packages are only available via `pip`.
This PR examines all the matrix dependencies. It will install all packages available on `conda` via `conda install`. For packages not available this way, it will try to install via `pip`.

### DIFF
--- a/asv/plugins/conda.py
+++ b/asv/plugins/conda.py
@@ -12,6 +12,7 @@ import six
 from .. import environment
 from ..console import log
 from .. import util
+import re
 
 
 WIN = (os.name == "nt")
@@ -98,18 +99,18 @@ class Conda(environment.Environment):
             # otherwise. It's also quicker than doing it one by one.
             conda_args = []
             pip_args = []
+
             for key, val in six.iteritems(self._requirements):
-                if self.conda_package_exists(key):
+                if key.startswith('pip+'):
+                    if val:
+                        pip_args.append("{0}=={1}".format(key[4:], val))
+                    else:
+                        pip_args.append(key[4:])
+                else:
                     if val:
                         conda_args.append("{0}={1}".format(key, val))
                     else:
                         conda_args.append(key)
-                # not on conda, try installing with pip
-                else:
-                    if val:
-                        pip_args.append("{0}=={1}".format(key, val))
-                    else:
-                        pip_args.append(key)
 
             conda_cmd = ['install', '-p', self._path, '--yes']
             pip_cmd = ['install', '-v', '--upgrade']
@@ -120,14 +121,6 @@ class Conda(environment.Environment):
             # install packages only available with pip
             if len(pip_args):
                 self.run_executable('pip', pip_cmd + pip_args)
-
-    def conda_package_exists(self, package):
-        # see if a conda package exists
-        ret = util.check_output(["conda", "search", '--json', package])
-        if ret == '{}\n':
-            return False
-        else:
-            return True
 
     def install(self, package):
         log.info("Installing into {0}".format(self.name))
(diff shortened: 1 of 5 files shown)


andyfaff:
Note: it's probably better to cache the packages available on conda, so one doesn't have to repeatedly check if they're available.

andyfaff:
Extra rationale: one could have a wide range of numpy or scipy versions you want to benchmark, both of which are available with conda. You then have one extra matrix dependency which is only available on pip. It's faster to use conda for the scipy numpy matrix, but you still need to install that other matrix dependency somehow

pv:
This seems fairly magical. Simpler would be to use a prefix `pip+` to
indicate packages to be installed via pip.

andyfaff:
Sorry, it's not quite clear what you mean. You mean in the config file or the command built in conda.py?

andyfaff:
Can one use the conda command to install pip only packages? If so, then it's not obvious from the conda help.

pv:
I mean in the config file --- one could write `pip+numpy` (or some similar notation) to indicate to asv the numpy package should be installed via pip and not conda. This way the choice of the installation tool would be static.

Another issue is then in the ordering of package installations, and ensuring the built packages will work in the conda environment.

andyfaff:
I'm not overly familiar with the importance of ordering of package installations. What are the pitfalls in this?

My particular situation is trying to benchmark `lmfit`. The `emcee` package is an optional dependency and we don't want to install by default in `setup.py`. However, I do want to benchmark the bits that depend on `emcee`. `emcee` is only available by `pip`. So I want to add `pip+emcee` to `matrix`, as well as benchmarking against different `numpy`'s, which I'd like to enter in matrix, but get installed via conda.

andyfaff:
@pv, the latest commit uses `pip+package-name` to specify that `package-name` is to be installed via `pip`. If one is in a conda environment type, then those matrix dependencies are installed via pip, after the other packages have been installed with conda.
Note that the environment name will look something like: `py3.4-pip+package-name` instead of `py3.4-package-name`. Is that a problem? Or should I clean up those names in `Conda.__init__` and `Virtualenv.__init__`?

pv:
The environment name is OK. (`pip+foo` results to a different environment from `foo`, so the env names should also differ.) Can you add a test to `test/test_environment.py`, needs skipif(not HAS_CONDA)?

Note also that something weird happened with git here.

andyfaff:
I screwed up when rebasing. Hopefully that's fixed now.
I'll add some tests.

andyfaff:
Tests added. It's not terribly clear from the project documentation how to successfully run tests on a local machine. I try running `python setup.py test` and get a lot of output, which isn't easy to understand. I don't know what's failing/passing. Are there extra dependencies I need to install?

pv:
cf http://pytest.org/latest/getting-started.html

andyfaff:
I don't think the test failures are to do with this PR.

asv/plugins/conda.py
```diff
@@ -92,19 +93,34 @@ def _setup(self):
...
+                    if val:
+                        conda_args.append("{0}={1}".format(key, val))
+                    else:
+                        conda_args.append(key)
+
+            conda_cmd = ['install', '-p', self._path, '--yes']
+            pip_cmd = ['install', '-v', '--upgrade']
+
+            # install conda packages
+            if len(conda_args):
```
pv: `if conda_args:`

asv/plugins/conda.py
```diff
@@ -12,6 +12,7 @@
 from .. import environment
 from ..console import log
 from .. import util
+import re
```
pv: stdlib import, so should go to top (together with `import os` et al.)

PR #380, opened 2016-03-04
ENH: Add an option to randomize the order of the hashes

astrofrog:
When running on a repo with e.g. 10,000 commits, this allows one to see the general trends building up slowly rather than doing each commit sequentially.

### DIFF
--- a/asv/commands/run.py
+++ b/asv/commands/run.py
@@ -6,6 +6,7 @@ from __future__ import (absolute_import, division, print_function,
 
 import six
 
+import random
 import logging
 import traceback
 
@@ -104,6 +105,15 @@ class Run(Command):
             "--skip-existing", "-k", action="store_true",
             help="""Skip running benchmarks that have previous successful
             or failed results""")
+        parser.add_argument(
+            "--randomize-order", "-r", action="store_true",
+            default=False,
+            help="""Randomize the order of the hashes to check. This means that
+            for example when running the tests for all commits (ALL), one can
+            already start to see trends well before all the commits have been
+            run. Note that if also using the ``steps`` option, the steps are
+            sampled from the commit hashes *after* re-ordering, so they may not
+            end up being equally spaced.""")
 
         parser.set_defaults(func=cls.run_from_args)
 
@@ -120,6 +130,7 @@ class Run(Command):
             skip_successful=args.skip_existing_successful or args.skip_existing,
             skip_failed=args.skip_existing_failed or args.skip_existing,
             skip_existing_commits=args.skip_existing_commits,
+            randomize_order=args.randomize_order,
             **kwargs
         )
 
@@ -127,7 +138,8 @@ class Run(Command):
     def run(cls, conf, range_spec=None, steps=None, bench=None, parallel=1,
             show_stderr=False, quick=False, profile=False, env_spec=None,
             dry_run=False, machine=None, _machine_file=None, skip_successful=False,
-            skip_failed=False, skip_existing_commits=False, _returns={}):
+            skip_failed=False, skip_existing_commits=False, randomize_order=False,
+            _returns={}):
         params = {}
         machine_params = Machine.load(
             machine_name=machine,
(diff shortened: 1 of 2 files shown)


astrofrog:
I'll add a simple test to make sure the option doesn't crash.

astrofrog:
The test I added doesn't actually check the results for now but just makes sure things don't crash with the option added. I **could** fix the random seed, run it, and check the order of the hashes in the stdout. Should I do that?

astrofrog:
@mdboom @pv - this is now ready for review!

mdboom:
I'm not sure I understand why a random order makes sense.  Wouldn't it be better (in terms of showing trends sooner) to use something like a binary tree ordering?

johntyree:
I would agree, with a slight modification. Just doing a normal binary tree ordering will have first and last commit, arguably the two most important, tested last. So I'd say first do those two, then do the rest with whatever kind of traversal you like.

pv:
I don't immediately see use cases where linear order is desirable, so perhaps it would be OK to just always use tree order, without adding command line options. Also --- I think the ordering should be shuffled after `--steps` has been taken into account, not before.

astrofrog:
Binary tree ordering would be fine too - I've used a lot of Monte-Carlo methods in the past, so I just personally like the idea of the ordering having no pattern, but binary tree ordering would be a fine compromise (with the two end points first).

astrofrog:
@pv - the use case for doing the shuffling (or say binary tree selection) *before* the steps in my case is that I would like to be able to run say 20 commits a night for a few hours to catch up with some older repositories. If I apply steps before reordering, it would be hard to run always 20 new commits to run every night.

At the end of the day, if this is too specialized a use case, I can keep this modification in my fork and just always run from my fork instead of the stable release, but it would be nice to avoid it.

Is there any harm to including the random option, since it's got a very small code footprint?

johntyree:
@astrofrog if you aren't tracking the commits to make sure they haven't been run before, then your approach is going to fall apart. As you complete more and more commits, the likelihood of rerunning one of them will increase until you are almost never doing new work.

If you *are* tracking which commits have been run, then I don't see the harm in doing them in a particular order.

astrofrog:
@johntyree - I agree, so I'd use the ``--skip-existing-commits`` option to ensure that.

In the binary tree ordering scenario, one would actually not want to apply ``steps`` as is after doing the tree ordering, but instead the user would want to take the N first commits after the reordering, correct? So if we do the binary ordering, does this mean that we should then change the meaning of steps to be the N first commits? If so, then the question is, would ``--skip-existing-commits`` be applied before or after the N first commits were selected from the tree? or would the tree be constructed only from non-existing commits?

From my perspective, the random ordering avoids a few of these issues because one could then run:

```
asv --random-order --skip-existing --steps=20
```

and this would select 20 random commits that don't already exist.

I think getting the order right for the binary tree ordering will be a little bit more complex, hence my wish to have the random-order too. Maybe we could have both ``--random-order`` and ``--binary-tree-order`` as options?

mdboom:
It seems like a good compromise to do both binary and random ordering.  In the case of binary and `--skip-existing`, intuitively it seems like if you hit an existing commit, you step forward incrementally until hitting a non existing commit (much like `git bisect skip` does).

johntyree:
Perhaps just remove the already-benchmarked commits from the selection pool prior to choosing?

astrofrog:
Just for information, I'm going to try and finish this up today.

astrofrog:
@pv @mdboom - actually, looking at the current code, it looks like the ordering is now essentially random because ``get_new_branch_commits`` returns a non-ordered set. So the question is, should we change that to return an ordered list/tuple, and then introduce the randomization, or shall we just leave as-is and close this PR?

pv:
I would prefer explicit randomization.
Note also the above comments about randomization and the skip options (the same may also apply to steps etc).

astrofrog:
Ok - in that case I'll try and work on this PR more next week. If we want to allow tree ordering, I'll need to modify get_new_branch_commits to return an ordered list initially.

astrofrog:
Realistically I am not going to have time to work on this anytime in the near or mid-term future, so to prevent stale pull requests, I am going to close this. However, if anyone would like to take this over, feel free to cherry-pick the commits here or start from scratch!

PR #381, opened 2016-03-05
BUG: fix two issues with asv publish

astrofrog:
The first issue occurred if a JSON file contained an integer and another contained a string, for the same parameter (for instance, for RAM, a file might contain “8G” and another file 104727291):

```
  File "/Users/tom/miniconda3/bin/asv", line 9, in <module>
    load_entry_point('asv', 'console_scripts', 'asv')()
  File "/Users/tom/tmp/asv/asv/main.py", line 36, in main
    result = args.func(args)
  File "/Users/tom/tmp/asv/asv/commands/__init__.py", line 48, in run_from_args
    return cls.run_from_conf_args(conf, args)
  File "/Users/tom/tmp/asv/asv/commands/publish.py", line 77, in run_from_conf_args
    return cls.run(conf=conf)
  File "/Users/tom/tmp/asv/asv/commands/publish.py", line 182, in run
    val.sort(key=lambda x: x or '')
TypeError: unorderable types: str() < int()
```

The second issue occurred if one file contained an empty string and another contained None, for the same parameter, since this would result in two empty strings in the final list and would then create too many entries in the generated website:

**Before**

screen shot 2016-03-05 at 10 14 01 am

**After**

screen shot 2016-03-05 at 10 15 00 am

### DIFF
--- a/asv/commands/publish.py
+++ b/asv/commands/publish.py
@@ -177,8 +177,12 @@ class Publish(Command):
         for key in six.iterkeys(benchmark_map):
             check_benchmark_params(key, benchmark_map[key])
         for key, val in six.iteritems(params):
-            val = list(val)
-            val.sort(key=lambda x: x or '')
+            # Values can be stored as strings, integers, floating point values,
+            # or None, so we now convert to a list of strings. We also make 
+            # sure that we convert to a set to remove duplicate items (for 
+            # instance, some files might contain "" and some None, but we need 
+            # to treat them as one).
+            val = sorted(set(str(x) if x is not None else '' for x in val))
             params[key] = val
         params['branch'] = [safe_branch_name(branch) for branch in conf.branches]
         util.write_json(os.path.join(conf.html_dir, "index.json"), {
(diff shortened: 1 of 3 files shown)


astrofrog:
I'll look into how easy it would be to add regression tests

astrofrog:
@mdboom @pv - this is now ready for review!

asv/commands/publish.py
```diff
@@ -177,8 +177,12 @@ def run(cls, conf):
...
             check_benchmark_params(key, benchmark_map[key])
         for key, val in six.iteritems(params):
-            val = list(val)
-            val.sort(key=lambda x: x or '')
+            # Values can be stored as strings, integers, floating point values,
+            # or None, so we now convert to a list of strings. We also make 
+            # sure that we convert to a set to remove duplicate items (for 
+            # instance, some files might contain "" and some None, but we need 
+            # to treat them as one).
+            val = sorted(set(str(x) if x is not None else '' for x in val))
```
pv: The comment should explain why None and "" appear there, and why they are equivalent.

I think the current asv version does not generate Nones there, so the conversion is probably for backward compat reasons --- previously `null` in dependency matrix mean the same as empty string.

pv:
Actually, there's a bug in the JS parameter selector, which requires changes in this part of the code, or alternatively one the JS side, gh-385

pv:
Probably superceded by gh-386

asv/commands/publish.py
astrofrog: Indeed - and I think back-compat is important in this case since we need to be able to include old results in the graphs.

pv:
Going to close, gh-386 should fix these + also fix the bugs in the "missing keys" handling

PR #384, opened 2016-03-07
For conda, install pip and wheel from conda so that they are compatible

pv:
This ensures the pip and wheel versions can work together; installing
`wheel` via pip can end up with pip/wheel combination that does not
work.

(Will merge asap, since currently the wheel cache on conda is broken because of incompatible wheel/pip versions.)

### DIFF
--- a/asv/plugins/conda.py
+++ b/asv/plugins/conda.py
@@ -91,35 +91,41 @@ class Conda(environment.Environment):
         self._install_requirements(conda)
 
     def _install_requirements(self, conda):
-        self.install('wheel')
-        if self._requirements:
-            # Install all the dependencies with a single conda command.
-            # This ensures we get the versions requested, or an error
-            # otherwise. It's also quicker than doing it one by one.
-            conda_args = []
-            pip_args = []
+        # Try to install the wheel package
+        _, _, retcode = util.check_output([conda, 'install', '-p', self._path, '--yes', 'wheel'],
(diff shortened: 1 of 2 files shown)


pv:
Ok, the reason for the failure was in my own environment, so it's not as urgent. (~/.local overrides the pip in the conda environments...). Nevertheless, no reason to install wheel from pip since it's also available in conda.

mdboom:
Seems fine to me, once the appveyor failure is investigated.

pv:
The appveyor failure seems to be a pip bug --- it tries to do something with path names > 260 chars without using long path prefix:
```
WindowsError: [Error 3] The system cannot find the path specified: 'c:\\\\users\\\\appveyor\\\\appdata\\\\local\\\\temp\\\\1\\\\pip-teiq9q-uninstall\\\\users\\\\appveyor\\\\appdata\\\\local\\\\temp\\\\1\\\\pytest-0\\\\test_find_benchmarks0\\\\env\\\\f30407b4f44fec871bacf62ec5e2a74b\\\\lib\\\\site-packages\\\\pip\\\\_vendor\\\\requests\\\\packages\\\\urllib3\\\\packages\\\\ssl_match_hostname\\\\_implementation.py'
```
Replaced with a different approach --- try to install `wheel` from conda, and if it fails, disable the wheel cache.

pv:
Closing, needs work but not going to do that in near future.

Issue #385, opened 2016-03-07
JS parameter selection broken vs requirements


pv:
The parameter selector in `graphdisplay.js` and `publish.py` does not currently handle the requirement matrix fully correctly.

The problem is that:

- The matrix allows combinations where some requirement is either installed or not.
- When the requirement is not installed, the `params` section of the corresponding results file contains no key for the requirement. If the requirement is installed, the key is present.
- The JS side and `publish.py` assume graphs have exactly the same set of possible parameters, and making no allowance for missing keys. As a consequence, graphs with missing keys are not loaded.

Possible solution:

In `publish.py`, find out the set of all parameter keys in a separate pass through the results, before constructing the graphs. After that, when constructing the graphs, set value of `"none"` for all missing keys in the graph parameters.

PR #386, opened 2016-03-07
Fix missing key handling in graph generation + parameter handling

pv:
Graph parameter handling needs to take into account that some parameter
keys can be missing in some of the result files. (E.g. as a result of
null entries in requirement matrix, or some requirements having been
added after some result files had already been generated.)

Deal with this as follows:

- In `asv publish`, for each missing key, assign None value in
  the corresponding graph.

- Make None translate to "null" in the graph file path.

- Fix parameter value sorting on Py3

- Update the JS side accordingly.

- Fix the JS parameter selector to deal with null

- Add additional tests

Fixes gh-385

### DIFF
--- a/asv/commands/publish.py
+++ b/asv/commands/publish.py
@@ -120,15 +120,22 @@ class Publish(Command):
         with log.indent():
             graph_groups = {}
 
+            # Determine first the set of all parameters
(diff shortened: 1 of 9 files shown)


mdboom:
LGTM.

PR #387, opened 2016-03-09
BUG: use branch list from config by default

johntyree:
@pv does this look right to you? I wrote this code awhile ago and have forgotten the details about how asv decides on commits to choose, but the issue it is intended to solve is listed below.

Closes https://github.com/spacetelescope/asv/issues/364

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -444,7 +444,9 @@ class Environment(object):
         if commit_hash is None:
             commit_hash = self._cache.get_existing_commit_hash()
             if commit_hash is None:
-                commit_hash = self.repo.get_hash_from_master()
+                commit_hash = self.repo.get_hash_from_name(conf.branches[0])
+                if commit_hash is None:
+                    commit_hash = self.repo.get_hash_from_master()
 
         self.uninstall(conf.project)
 
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -444,9 +444,9 @@ class Environment(object):
         if commit_hash is None:
             commit_hash = self._cache.get_existing_commit_hash()
             if commit_hash is None:
-                commit_hash = self.repo.get_hash_from_name(conf.branches[0])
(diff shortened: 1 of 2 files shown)


asv/environment.py
```diff
@@ -444,7 +444,9 @@ def install_project(self, conf, commit_hash=None):
         if commit_hash is None:
             commit_hash = self._cache.get_existing_commit_hash()
             if commit_hash is None:
-                commit_hash = self.repo.get_hash_from_master()
+                commit_hash = self.repo.get_hash_from_name(conf.branches[0])
```
johntyree: Hmm... is it the case that this is always nonempty?

pv: Yes, cf https://github.com/spacetelescope/asv/blob/master/asv/config.py#L69

asv/environment.py
```diff
@@ -444,7 +444,9 @@ def install_project(self, conf, commit_hash=None):
...
+                if commit_hash is None:
```
pv: get_hash_from_name will raise an exception if it can't resolve, so I think this fallback isn't needed.

asv/environment.py
pv: Probably needs to be `self.repo.get_hash_from_name(self.repo.get_branch_range_spec(conf.branches[0]))`

johntyree:
Tests went down in flames last time. I don't see a way to cancel them myself so we'll have to wait to see if they are resolved with your suggested changes.

asv/environment.py
johntyree: removed

asv/environment.py
johntyree: :+1:

pv:
repo.py:NoRepository needs to stub get_branch_range_spec

pv:
Done in #430

PR #393, opened 2016-03-21
Fix Hg.get_hash_from_name()

philpep:
The function returned the local revision number which is not a correct
commit hash.


pv:
Thanks.
Consider adding also tests when making bug fixes.

PR #394, opened 2016-03-23
Fix hg branches

philpep:
This is a series of fix to make the mercurial plugin behave like the git one.

  * Mercurial default branch is "default"
  * Only follow first parent on merges changesets

I added a test with merges that work for both mercurial & git backends.

### DIFF
--- a/test/tools.py
+++ b/test/tools.py
@@ -113,8 +113,17 @@ class Git(object):
     def add(self, filename):
         self._run_git(['add', relpath(filename, self.path)])
 
-    def create_branch(self, start_commit, branch_name):
-        self._run_git(['checkout', '-b', branch_name, start_commit])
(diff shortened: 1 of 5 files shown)


test/tools.py
```diff
@@ -269,6 +291,39 @@ def generate_test_repo(tmpdir, values=[0], dvcs_type='git',
     return dvcs
 
 
+def make_test_repo(tmpdir, dvcs_type, operations):
```
philpep: Maybe I could adapt my `make_test_repo()` and refactor the test code to replace `generate_test_repo()`. Let me known.

pv: I'd rename the function `make_test_repo -> generate_test_repo_from_ops`.

pv:
thanks, merged

PR #395, opened 2016-04-08
mercurial: close subrepo after checkout

philpep:
hglib repo must be closed explicitely to stop the underlying "hg serve"
subprocess.

Also remove what looks like an outdated TODO

### DIFF
--- a/asv/plugins/mercurial.py
+++ b/asv/plugins/mercurial.py
@@ -104,9 +104,9 @@ class Hg(Repo):
         # the repository data is not shared
 
         def checkout_existing():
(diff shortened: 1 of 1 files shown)


pv:
The TODO is not outdated --- this is supposed to also remove all 
untracked files (`git clean -fdx`), but it doesn't.

philpep:
Ok I've updated my commit & commit message. To remove untracked files it seems we need to call `subrepo.rawcommand('purge')`, I'll test this and a make a separate PR.

pv:
Thanks, merged.

PR #398, opened 2016-04-13
Auto zoom the y axis when zooming in

### LINKED ISSUE #396: Feature request: Automatically scale Y axis when zooming in on X range
It would be nice one could scale the Y axis to fit the range of the currently shown data. For example, if I select an X range to focus on and the benchmark had a very large Y range for the commits not shown in the zoomed range, it is difficult to see the performance changes since the Y axis doesn't scale.

Example:
selection_001
selection_003

mdboom:
Fix #396.

Cc: @moorepants

### DIFF
--- a/asv/www/graphdisplay.js
+++ b/asv/www/graphdisplay.js
@@ -888,9 +888,22 @@ $(document).ready(function() {
         /* Find the minimum and maximum values */
         var min = Infinity;
         var max = -Infinity;
+        var left = options.xaxis.min || 0;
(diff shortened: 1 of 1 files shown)


moorepants:
I'll try this branch out tomorrow. Thanks!

moorepants:
I finally checked this out. It seems to be doing what I'd expect, although I don't have access to our larger sympy asv database at the moment. I just ran a few commits and the y axis max seems to scale to the data on the screen. Thanks!

pv:
LGTM

PR #400, opened 2016-04-13
Save run times

mdboom:


pv:
What should the behavior be when not all benchmarks in the file were not run at the same time?
E.g. some benchmarks added later on, and re-run for old commits.

pv:
Addition to docs/source/dev.rst could be useful.

pv:
Superceded by #428

PR #401, opened 2016-04-13
Fix tags and zoom in even commit spacing

philpep:
transform() and inverseTransform() now return a value even when
date/index are not in the known values (commits dates).

This fixes tags display (if the tag is not associated with a result
commit) and zooming for both main and overview graphs.


pv:
Thanks, looks correct to me.

PR #402, opened 2016-04-14
mercurial: clean untracked files

philpep:


asv/plugins/mercurial.py
```diff
@@ -106,7 +106,7 @@ def checkout_existing():
             with hglib.open(path) as subrepo:
                 subrepo.pull()
                 subrepo.update(commit_hash, clean=True)
```
mdboom: Ah -- I see.  `clean=True` here only discards uncommitted changes, not untracked files.  This makes sense, then, and brings the behavior inline with the `git` plugin.

mdboom:
It looks like this functionality needs a mercurial plugin.  Will need to add that to `.travis.yml` and update requirement docs etc.

pv:
It would be better to somehow tell Mercurial to enable the plugin here.
Users generally may not have the `purge` plugin enabled.

philpep:
I added explicit enabling of the purge extension here.

pv:
Thanks

PR #407, opened 2016-05-04
Make Environments not store their own Repo objects

pv:
Environment only needs to know about a Repo when it is building a
project. Pass the repository in as an argument when needed.

The actual issue this fixes is to avoid cloning the repository when it
is not necessary, for example in 'asv run --python=same'.
(The cloning in this case is a regression introduced in gh-352)

- [x] WIP: may still need some testing


pv:
Merging, should be OK.

PR #410, opened 2016-05-08
Run pip self-upgrade on Windows via python -m pip, so that it works

pv:
On windows, pip cannot upgrade itself, except if run via `python -m pip install ...`.
Change the virtualenv upgrade command to match.


pv:
Merging (tested).

PR #392, opened 2016-03-18
GH391

hcontrast:
Fix attempt GH391

PR #412, opened 2016-05-08
Change install_timeout default value back to 600s

pv:
This was changed from 600s to 120s in gh-392 without comment, so change it back (travis-ci has 600).

I suspect the config variable is not fully necessary, however, as the issue may have been due to https://github.com/pypa/pip/issues/3486


pv:
Merging, straightforward.

PR #420, opened 2016-05-26
Fix config file path

mdboom:
Since 92373df310b838b12b10a00be7a6b38be3e17ee2 it wasn't possible to specify a config file outside of the current directory.


pv:
LGTM

PR #422, opened 2016-06-03
Fix default branch name for mercurial

philpep:
This avoid potential error when using graph.params['branch'] as input of
Repo methods.

### DIFF
--- a/asv/repo.py
+++ b/asv/repo.py
@@ -14,6 +14,10 @@ class Repo(object):
 
     There are no commands that modify the source repository.
     """
+
+    # The default branch name when no branch is configured in asv.conf.json
+    _default_branch = None
+
     def __init__(self, url, mirror_path):
         """
         Create a mirror of the repository at `url`, without a working tree.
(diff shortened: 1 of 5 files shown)


pv:
Nitpick: I would avoid the `branch or xxx` idiom as it conflates None 
and all other false values -- e.g. `branch if branch is not None else xxx`.

philpep:
Ok, implemented a `get_branch_name()` method for this purpose.

asv/repo.py
```diff
@@ -152,6 +166,7 @@ class NoRepository(Repo):
     """
 
     dvcs = "none"
+    _default_branch = "master"
```
philpep: Not sure about these changes on NoRepository. Let me known if you have doubts.

pv: NoRepository should be OK; it shouldn't need the branch value, since 
it's not checking out code or saving any results.

pv:
Thanks, lgtm

PR #425, opened 2016-06-03
Implement "pretty_name" attribute on benchmark func

### LINKED ISSUE #424: Display custom name instead of benchmark function name in web ui
Hello,

Our use case is to benchmark mercurial revsets, a revset is a expression like ``roots((0::) - (0::tip))`` and we want to display the revset as it in the web ui instead of the benchmark function name which seems unrelated because it cannot contains special characters (eg. ``time_root_0_minus_0_tip``).

I guess there can be other similar uses cases of such feature for projects using asv.

philpep:
This optional parameter is printed in web UI instead of function name
allowing to use all characters set that would be avoided for a python
function name.

Closes #424

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -266,6 +266,7 @@ class Benchmark(object):
         name = name.split('.', 1)[1]
         self.name = name
         self.func = func
+        self.pretty_name = getattr(func, "pretty_name", name)
         self._attr_sources = attr_sources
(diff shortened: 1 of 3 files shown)


pv:
Thanks

pv:
Note that it's possible to test also the Javascript part, cf. `test/test_web.py`.

PR #428, opened 2016-06-05
Add benchmark-specific start/end timestamps in the result files

pv:
Record the time when a benchmark was run, on a per-benchmark basis. Working on a per-benchmark basis addresses issues arising if specific benchmarks have been rerun, in which case having a single date range for the whole file may not be indicative.

Alternative to gh-400

### DIFF
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -13,6 +13,7 @@ import shutil
 import sys
 import tempfile
 import itertools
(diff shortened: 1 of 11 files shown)


philpep:
I think you should also remove benchmark start/end time in the ``asv rm benchmark=glob`` command.

philpep:
LGTM

PR #429, opened 2016-06-06
Handle benchmarks by commit revision instead of date

### LINKED ISSUE #390: Order by commit log instead of date
Hello,

Commits are ordered by date for both graph and regression detection. In a workflow with merges (rebase + fast forward merge for instance) or cherry-picked commits, the commit date is not monotonic along the commit log.

I'm considering fixing the publish command to reorder commits as they appear in the commit log, but this will be a big breaking change. I wonder if you have others ideas to fix or workaround this issue ?

Thanks!

philpep:
Commits date are not necessary monotonic along the commit log,
especially if working with rebase workflow (instead of merge), this
generate false positive or missed regression detection and unwanted
peaks in the web UI.

This patch introduce a revision number as a drop-in replacement of
commit date. This revision number is an incremental number representing
the number of ancestors commits.

In the web UI the distance between commits is the number of commits
between them. Commits can still be displayed by date using the date
selector.

Closes #390

### DIFF
--- a/asv/commands/publish.py
+++ b/asv/commands/publish.py
@@ -71,8 +71,6 @@ class Publish(Command):
     def run(cls, conf, env_spec=None):
         params = {}
         graphs = GraphSet()
-        date_to_hash = {}
-        hash_to_date = {}
         machines = {}
         benchmark_names = set()
 
@@ -97,23 +95,12 @@ class Publish(Command):
                 machines[d['machine']] = d
 
         log.step()
-        log.info("Getting tags and branches")
+        log.info("Getting params, commits, tags and branches")
         with log.indent():
-            repo.pull()
-            tags = {}
-            for tag in repo.get_tags():
-                log.dot()
-                tags[tag] = repo.get_date_from_name(tag)
-
-            branches = dict(
-                (branch, repo.get_branch_commits(branch))
-                for branch in conf.branches)
-
-        log.step()
-        log.info("Loading results")
-        with log.indent():
-            # Determine first the set of all parameters
+            # Determine first the set of all parameters and all commits
+            hash_to_date = {}
             for results in iter_results(conf.results_dir):
+                hash_to_date[results.commit_hash] = results.date
                 for key, val in six.iteritems(results.params):
                     if val is None:
                         # Backward compatibility -- null means ''
@@ -122,11 +109,27 @@ class Publish(Command):
                     params.setdefault(key, set())
                     params[key].add(val)
 
+            repo.pull()
+            tags = repo.get_tags()
+            revisions = repo.get_revisions(set(hash_to_date.keys()) | set(tags.values()))
+
+            for tag, commit_hash in list(tags.items()):
+                # Map to revision number instead of commit hash and add tags to hash_to_date
+                tags[tag] = revisions[tags[tag]]
+                hash_to_date[commit_hash] = repo.get_date_from_name(commit_hash)
+
+            revision_to_date = dict((r, hash_to_date[h]) for h, r in six.iteritems(revisions))
+
+            branches = dict(
+                (branch, repo.get_branch_commits(branch))
+                for branch in conf.branches)
+
+        log.step()
+        log.info("Loading results")
+        with log.indent():
             # Generate all graphs
             for results in iter_results(conf.results_dir):
                 log.dot()
-                date_to_hash[results.date] = results.commit_hash
-                hash_to_date[results.commit_hash] = results.date
 
                 for key, val in six.iteritems(results.results):
                     b = benchmarks.get(key)
@@ -154,7 +157,7 @@ class Publish(Command):
 
                         # Create graph
                         graph = graphs.get_graph(key, cur_params)
-                        graph.add_data_point(results.date, result)
+                        graph.add_data_point(revisions[results.commit_hash], result)
 
         log.step()
         log.info("Generating graphs")
@@ -169,7 +172,7 @@ class Publish(Command):
             log.step()
             log.info("Generating output for {0}".format(cls.name))
             with log.indent():
-                cls.publish(conf, repo, benchmarks, graphs, hash_to_date)
+                cls.publish(conf, repo, benchmarks, graphs, revisions)
                 extra_pages.append([cls.name, cls.button_label, cls.description])
 
         log.step()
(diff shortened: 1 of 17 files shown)


philpep:
Hello,

Sorry the commit is quite huge but most of the diff is just a replacement of variables names (date -> revision).

My first attempt was to only use (sorted) commits hashes in the python code and to not use this revision number. But it was too complicated mostly due to summary graphs that need to known the "distance" between commits and the javascript code was ugly... So I just replaced the commits date by the revision number.

You can see a real demo with non-monotonic commits dates on https://hg.logilab.org/users/ppepiot/hgperf/raw-file/50a52a1968de/index.html#others.time_tags?branch=default&rev=28265&idx=2

philpep:
- [ ] Need to fix the web tests

philpep:
Ok the failed test was due to a missing sort parameter in `git log --all`. I have chosen ``--date-order`` because it seems better than ``--topo-order`` for displaying multiple branch graph (btw, both show children before parents).

pv:
I'd suggest using `git rev-list` instead of the UI-level command `git log`.

philpep:
@pv indeed, commit updated.

asv/www/graphdisplay.js
```diff
@@ -1017,28 +1047,43 @@ $(document).ready(function() {
...
+                graphs = $.extend(true, [], orig_graphs);
+                date_to_revision = {};
+                $.each(graphs, function(i, graph) {
+                    $.each(graph.data, function(j, point) {
+                        var date = $.asv.master_json.revision_to_date[point[0]];
+                        date_to_revision[date] = point[0];
+                        point[0] = date;
+                    });
+                    graph.data.sort(function(a, b) { return a[0] - b[0]; });
+                });
```
philpep: Some hints to understand this `orig_graphs` purpose.

I was planning to use flot `transform` and `inverseTransform` functions build the `date scale` graph using `revision_to_date`. But flot assume the transform function does not reorder values (eg. is monotone), so it cannot be used here unfortunately.

So this build a new graphs and keep a reference to the old graphs for switching from one to another (but maybe you have a better idea ?).

asv/commands/publish.py
```diff
@@ -122,11 +109,28 @@ def run(cls, conf, env_spec=None):
                     params.setdefault(key, set())
                     params[key].add(val)
 
+            repo = get_repo(conf)
```
pv: Probably not necessary to call `get_repo` again here?

philpep: Indeed, fixed in 53388ea

pv:
I think the overall approach looks fine, and is the minimum perturbation. Refactoring can be thought of later, if needed.

On `orig_graphs`: looks fine. Another way would be to just call `replace_graphs()` instead of `update_graphs()` in the `#date-scale` callback function --- performance is probably not that much different, since flot is iirc the slowest part (the data is cached on js side, so no network traffic).

pv:
There's a small rebase needed.

asv/graph.py
```diff
@@ -303,21 +301,22 @@ def resample_data(val):
...
 
     if step_size == 0:
-        step_size = max_time - min_time + 1
+        step_size = max_revision - min_revision + 1
 
     # This loop cannot use xrange, because xrange on Python2 on
     # 32-bit systems can only deal with 32-bit integers, and
     # Javascript timestamps (1000*unix_timestamp) handled here
     # overflow this range
+    # FIXME: Use xrange
```
pv: `for i in xrange(min_revision, max_revision, step_size)`  (iow just revert cc6accaee3, d88095 since it's not needed any more)

pv:
Also, I guess that if you get new commits in different branches, the revision numbers will change?

This will invalidate the links to the regression pages --- so maybe it would be better to use commit hashes as the `rev=` keys in those, rather than revision numbers? Otherwise, the revision numbers don't seem to be visible outside asv afaics.

philpep:
Hi, thanks for you feedback.

I rebased the commit and amended a revert of cc6acca (also added a test for summary graphs) and use commit hash instead of revision number in the regression url (indeed "fresh" commits can have their revision number changed if using multiple branches).

pv:
Thanks. Note that the web test currently fails.

philpep:
Yes, should be fixed now.

asv/www/regressions.js
```diff
@@ -181,13 +181,15 @@ $(document).ready(function() {
...
             });
 
-            url_params.time = [];
-            $.each(dates, function(i, date) {
-                if (date[0]) {
-                    url_params.time.push(date[0] + '-' + date[1]);
+            url_params.commits = [];
+            $.each(revisions, function(i, revs) {
+                var commit_b = $.asv.get_commit_hash(revs[1]);
+                if (revs[0]) {
```
pv: `if (revs[0] === null)`? The first revision is `0` which is also false...

philpep: Yes fixed (I don't known how the revision "0" could be part of a regression, but the code is more clear now).

asv/www/regressions.js
```diff
@@ -364,11 +366,11 @@ $(document).ready(function() {
...
             return null;
         }
-        var dates = regression[0];
+        var revisions = regression[0];
         var ignore_payload = benchmark_name;
 
-        $.each(dates, function (i, date) {
-            ignore_payload = ignore_payload + ',' + date[0] + ',' + date[1];
+        $.each(revisions, function (i, revs) {
+            ignore_payload = ignore_payload + ',' + revs[0] + ',' + revs[1];
```
philpep: I think I should use commit hash here too for the html storage.

Btw this is breaking current storage based on timestamps (is this acceptable ?).

pv: It's fine to change the hashes, as things become just un-ignored --- this is a per-user convenience feature (doesn't transfer between different browsers etc.).

pv:
Thanks, merged.

pv:
@philpep: did you notice the regressions here look a bit weird regarding c89b55c8: https://jenkins.philpep.org/hgperf/index.html#regressions?sort=1&dir=desc --- maybe a history flattening artefact?

philpep:
@pv yes I changed the hg reference repositories to benchmark against, so the results before and after c89b55c8 aren't comparables.

PR #430, opened 2016-06-08
Use configured branches when no commit/range is specified

### LINKED ISSUE #364: Plain "asv run" always using "master" is inconsistent.
I was expecting `asv run` to run the benchmarks using the branches listed in `asv.conf.json`. Poking around in docs and source indicates that [this is not the case](https://github.com/spacetelescope/asv/blob/master/asv/commands/run.py#L151).

Why does the default behavior differ from `ALL` and `NEW`? Particularly since the default value of `conf.branches` is `["master"]` anyway.

philpep:
Fix behavior of run, profiling and continuous commands that were using
get_hash_from_master() instead of configured branches when no commit was
specified.

In particular when only one branch was configured and was not "master",
asv tried to use the "master" branch.

Add (and refactor) specific test.

Closes #364 
Alternative to #387

### DIFF
--- a/test/test_workflow.py
+++ b/test/test_workflow.py
@@ -182,40 +182,7 @@ def test_find(capfd, basic_conf):
     assert "Greatest regression found: {0}".format(regression_hash[:8]) in output
 
 
-def _test_run_branches(tmpdir, dvcs, conf, machine_file, range_spec,
-                       branches, initial_commit):
-    # Find the current head commits for each branch
-    commits = [initial_commit]
-    for branch in branches:
-        commits.append(dvcs.get_hash(branch))
-
-    # Run tests
-    tools.run_asv_with_conf(conf, 'run', range_spec, '--quick',
-                            '--bench=time_secondary.track_value',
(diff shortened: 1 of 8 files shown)


test/test_workflow.py
```diff
@@ -224,33 +193,51 @@ def test_run_new_all(basic_conf):
...
+        for commit in expected_commits:
+            for psver in ['0.3.6', '0.3.7']:
+                expected.add('{0}-{1}-py{2[0]}.{2[1]}-colorama{3}-six.json'.format(
+                    commit[:8], tool_name, sys.version_info, psver))
+
+        result_files = os.listdir(join(tmpdir, 'results_workflow', 'orangutan'))
+
+        assert set(result_files) == expected
+
+    for branches, expected_commits in (
```
pv: Style nitpick: I'd prefer to write multiline for assignments as
```
branches_and_commits = (
    ...
)
for branches, expected_commits in branches_and_commits:
...
```

pv:
LGTM, merged.

PR #432, opened 2016-06-12
In web ui ignore hashing, use commit hashes instead of revision numbers

pv:
Leftovers from #429.

Straightforward, so will merge soon.


philpep:
LGTM, thanks.

PR #437, opened 2016-06-15
Revised/additional summary page

pv:
Add a second summary page, showing a list of benchmarks (instead of a grid).

The list shows benchmarks only for one choice of environment parameters (machine, cpu, packages, ...) at once.

Also show how the benchmark performance has evolved "recently", some ideas borrowed from http://speed.pypy.org/changes/

Some overlap with the Regressions display, with the exception that regressions shows regressions across all branches, and does not show improved results.

Demo: https://pv.github.io/numpy-bench/#summarylist
https://pv.github.io/scipy-bench/#summarylist

- [x] what information should it actually show?
- [x] what does "recently" mean? 1 month, 10 revisions, ...?
- [x] what to do with benchmarks with no results?
- [x] the param selector should deal sensibly if there are no results for some param combination
- [x] tests, etc. polish
- [x] a selector for the time/revision range that you want to consider (e.g. show the worst regression ever vs. the last one vs. benchmark-specific configuration in the config file)
- [x] maybe add expandable view to show all regressions?
- [x] the popup graphs

### DIFF
--- a/asv/commands/publish.py
+++ b/asv/commands/publish.py
@@ -180,18 +180,18 @@ class Publish(Command):
         log.step()
         log.info("Generating graphs")
         with log.indent():
-            # Add summary graphs
-            graphs.make_summary_graphs(dots=log.dot)
             # Save files
             graphs.save(conf.html_dir, dots=log.dot)
 
-        extra_pages = []
-        for cls in util.iter_subclasses(OutputPublisher):
+        pages = []
+        classes = sorted(util.iter_subclasses(OutputPublisher),
+                         key=lambda cls: cls.order)
+        for cls in classes:
             log.step()
-            log.info("Generating output for {0}".format(cls.name))
+            log.info("Generating output for {0}".format(cls.__name__))
             with log.indent():
                 cls.publish(conf, repo, benchmarks, graphs, revisions)
-                extra_pages.append([cls.name, cls.button_label, cls.description])
+                pages.append([cls.name, cls.button_label, cls.description])
 
         log.step()
         log.info("Writing index")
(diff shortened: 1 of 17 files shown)


philpep:
This looks awesome, thanks !!

About "what does "recently" mean? 1 month, 10 revisions, ...?"

Looking the latest step (regression or improvement) could be enough. It seems this is what you're currently doing with `last_piece = steps[-1]`.

Now we have this "improvement" information, we could provide a "per commit" view that show all regressions and improvement where a particular commit is involved in. This will help for our use case (mercurial revsets) where a single change can impact multiple benchmarks in a good or bad manner, in this case we may want to promote improvement on very used revsets and accept some regression on rarely used ones.

But indeed this new page overlap with the regression page. Maybe we could merging them in a single page and adding some selectors to filter showed data.

I really like the colors and the percentage here. What represent the data after the  `±`, is it the noise ?

pv:
Yes, I think it is relatively straightforward to add different views to the data, and the per-commit one could also be useful. The difficult part is just deciding what to present and how. (I'm a bit time-constrained currently, so don't let this PR/WIP block if you're thinking of something, since this may take long to finish.)

I don't see the overlap with the regression pageas necessarily as a problem in itself, but it can be useful to import the ideas that can be imported. Potentially for the present page: 
- [ ] a selector for the time/revision range that you want to consider (e.g. show the worst regression ever vs. the last one vs. benchmark-specific configuration in the config file)
- [ ] maybe add expandable view to show all regressions?
- [ ] the popup graphs

The +- shows the noise level of the value (= average deviation from the median in the last piecewise step). It maybe should be multiplied by 2 to be more intuitive... Added as inspired by speed.pypy.org.

pv:
I think this is done now. Additions can be added later.

pv:
Merging --- after dogfooding for a while, seems to work ok.

PR #440, opened 2016-06-18
Fix combinatorial explosion in graph selection

### LINKED ISSUE #438: asv html site not working correctly
I am having issues previewing the HTML results site for the following example repo:

https://github.com/astrofrog/asv-debug

If I clone this, then do:

```
asv publish
asv preview
```

then click on one of the preview graphs that appears, the resulting pages are very slow and show either no results or only some of them.

pv:
Previously, the JS UI loaded the graphs by generating all parameter
combinations and then tried to http GET files to see which ones exist.
This is bad for performance.  Instead, save the list of all available
parameter combinations to index.json and filter based on that.

The permutation generation is OK for parameterized benchmarks, since
for those the result set is not sparse.

Fixes #438

### DIFF
--- a/asv/commands/publish.py
+++ b/asv/commands/publish.py
@@ -159,6 +159,13 @@ class Publish(Command):
                         graph = graphs.get_graph(key, cur_params)
                         graph.add_data_point(revisions[results.commit_hash], result)
 
+            # Get the parameter sets for all graphs
+            graph_param_list = []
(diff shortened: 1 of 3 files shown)


astrofrog:
Works for me - thanks @pv!

pv:
Thanks for the confirmation, merging.

PR #441, opened 2016-06-18
Fix two bugs in saving environment parameters in benchmarks

pv:
Fix two bugs in saving environment parameters in benchmarks:

If build failed, the environment parameters were not saved. Also, since dict.update was called, parameters from previous environments could be mixed up in following results.

In practice, the wrong data causes e.g. spurious parameter combinations to appear in the UI.


pv:
Merging, should be straightforward.

PR #444, opened 2016-06-21
Include autocorrelation to the noise model used in step detection

pv:
This is a simple-minded approach for accounting for some autocorrelation
in the benchmark result noise. It improves step detection quality
somewhat in cases where there is a smoothly varying background
superimposed on stepwise changes. This also allows using smaller
degrees-of-freedom counts.


pv:
Merging, seems to be OK based on dogfooding.

Issue #397, opened 2016-04-13
Implement a notification system


philpep:
One possible implementation could be to generate a rss/atom feed of regressions in the static website. What do you think ?

mdboom:
I think that's a great idea!

philpep:
The hard part is what date using as "publication date" in the feed item ? This date must be stable along rebuild of the static website (asv publish).

Here are possibles candidates/solutions:

  * Commit "author date" (``%ad`` in git log format). I'm not in flavor of this cf #390
  * Commit "commit date" (``%cd`` in git log format). No equivalent in mercurial.
  * Benchmark build date cf #400  
  * When publishing, it the target ``regressions.json`` and future ``regressions.xml`` exists, make a diff of regressions and eventually add new items in the feed (instead of overriding). IMHO it's a bit magic, but sould work pretty well.

pv:
I would prefer to keep the html output as write-only, so that you can always throw the old html away.

The benchmark build date is probably the one most closely related to the date when the regression was actually detected. One thing here is that due to the noise in the results, it will lag a few benchmark runs behind, so you get new publications in the past. However, the order of publications is the same as if the publication date was the date of actual publication, so this sounds good to me.

pv:
Initial stab at this: https://github.com/spacetelescope/asv/compare/master...pv:regressions-feed

philpep:
@pv thanks ! I was planning to work on it (but didn't yet). I'll try this on top of #429

pv:
Great, feel free to take over! It doesn't currently organize the feed in 
the same way as the web page, I run out of time yesterday...

pv:
Implemented in #447

philpep:
@pv looks perfect ! Thanks !

PR #447, opened 2016-06-23
Add Atom feed showing a list of regressions

pv:
Implement an Atom feed displaying a list of regressions.

Each feed entry contains the magnitude of change and benchmark name in the title. The body contains benchmark values and information when and on what commit the regression occurred, and the new and old benchmark result values. The publication date of an entry is determined by the date when the benchmark was run (if not recorded in old results, falls back to commit date).

Demo (link on upper right): https://pav.iki.fi/tmp/asv-feed/html-numpy/#regressions
https://pav.iki.fi/tmp/asv-feed/html-scipy/#regressions

Discussed in #397


pv:
Seems to work OK based on dogfooding, and the idea was OK'd in #397, so merging.
Waiting for reported issues to give more feedback if something needs to be improved.

Issue #417, opened 2016-05-21
PyPy support?


njsmith:
For a change, I have a project that actually can be run under PyPy. (I guess this will become much more common within asv's target audience, since they're working on supporting the full scientific stack within the year.) I wanted to try tracking speed under both CPython and PyPy. But this is kind of a mess because (a) AFAICT there's no way to ask asv to use pypy for running stuff, and (b) with the current benchmark runner it wouldn't be useful anyway, because measuring the first few seconds of a PyPy run isn't very interesting; I want to run it for a while to let the JIT warm up and then see how fast it's going.

I guess no-one will spring into action and implement this just because I filed an issue, but I figured it might be useful to at least make a note that this could be an interesting addition and provide a place to talk about what might be needed :-)

pv:
(a) #452 
(b) I'm not sure the current default values are that unreasonable for pypy --- the minimum of 3 runs with repetitions to bring the total duration to 0.1--1 sec --- appears to be enough to allow for jit warmup in practice. As these values are tunable by the user, it's looks ok as is.

PR #452, opened 2016-06-30
Support PyPy

pv:
Add support for PyPy:

- for `pythons=["pypy"]` etc in virtualenv
- for running asv itself

The `mem_` benchmarks don't work, and may not be fixable, so I just documented they don't work.

Addresses #417


pv:
Merging, should be straightforward change.

PR #456, opened 2016-07-10
Add a configurable threshold to regression reporting

pv:
Add a user-configurable threshold for determining whether `asv publish` should report a regression or not. Change the default threshold from 0% relative change to 5%, in order to make the regression reporting less noisy out-of-the-box.

This contains also same refactoring commits as in gh-437.


pv:
Merging, seems to work ok.

Issue #461, opened 2016-07-17
KeyError: u'params' when running asv gh-pages


astrofrog:
When I try and run asv gh-pages for the following benchmarks:

https://github.com//astropy/astropy-benchmarks

I get the following error:

```
$ asv gh-pages
[ 14.29%] · Loading machine info
[ 28.57%] · Getting params, commits, tags and branchesTraceback (most recent call last):
  File "/disk1/data/anaconda/envs/hyperion-latest/bin/asv", line 9, in <module>
    load_entry_point('asv===0.2.dev0-', 'console_scripts', 'asv')()
  File "/disk1/data/anaconda/envs/hyperion-latest/lib/python2.7/site-packages/asv/main.py", line 38, in main
    result = args.func(args)
  File "/disk1/data/anaconda/envs/hyperion-latest/lib/python2.7/site-packages/asv/commands/__init__.py", line 48, in run_from_args
    return cls.run_from_conf_args(conf, args)
  File "/disk1/data/anaconda/envs/hyperion-latest/lib/python2.7/site-packages/asv/plugins/github.py", line 44, in run_from_conf_args
    return cls.run(conf=conf)
  File "/disk1/data/anaconda/envs/hyperion-latest/lib/python2.7/site-packages/asv/plugins/github.py", line 58, in run
    Publish.run(conf)
  File "/disk1/data/anaconda/envs/hyperion-latest/lib/python2.7/site-packages/asv/commands/publish.py", line 103, in run
    for results in iter_results(conf.results_dir):
  File "/disk1/data/anaconda/envs/hyperion-latest/lib/python2.7/site-packages/asv/results.py", line 39, in iter_results
    yield Results.load(os.path.join(root, filename))
  File "/disk1/data/anaconda/envs/hyperion-latest/lib/python2.7/site-packages/asv/results.py", line 337, in load
    d['params'],
KeyError: u'params'
```

My versions are:

```
bash-4.3$ asv --version
asv 0.2.dev0+
bash-4.3$ python --version
Python 2.7.11 :: Anaconda 2.1.0 (64-bit)
```

asv is using the latest developer version at the time of writing

pv:
Does `asv publish` work?

pv:
The reason is the `results/ant/machine-scipy0.17.json` file, which is not named `machine.json` or a valid results file.

pv:
asv should print a warning and ignore a malformed file, rather than 
crashing.

PR #462, opened 2016-07-18
Warn about obviously malformed results files instead of terminating with an error

pv:
Fixes gh-461


pv:
Merging, should be ok.

PR #464, opened 2016-07-23
Fix summarylist urls to respect benchmark parameters

pv:
The summarylist urls didn't include the correct parameter selectors.
Add those.


pv:
Merging, straightforward.

Issue #459, opened 2016-07-12
ExistingEnvironment should ignore `matrix`


pv:
ExistingEnvironment cannot install packages, and should ignore the matrix specified in the configuration.

See #458

PR #466, opened 2016-07-25
Fix usability issues in environment selection

pv:
- Fix bug in environment exclude rules with autodetected environment_type
- When the user explicitly selects an environment (e.g. on command line), use the specified environment/python combination even if exclude rules result to an empty requirement matrix.
-  	Make ExistingEnvironment ignore the requirements matrix. This ensures -E existing or --python=same results to only one environment.

Fixes gh-459


pv:
Merging, should be ok

PR #472, opened 2016-07-31
Allow users to control setup_cache timeout

pv:
The timeout for `setup_cache` cannot currently be controlled by the user. Make this possible.

The cache setup timeout is taken from the .timeout attribute of the `setup_cache` function. If not specified, the default value is taken the maximum of the timeouts of the benchmarks.


pv:
Merging, should be fine

Issue #471, opened 2016-07-28
Set indexing error when running asv from master


moorepants:
On master I am getting this, with:

`SYMPY_CACHE_SIZE=10000 asv --config asv.conda.conf.json run ALL --parallel -k -e -s 500`

```
· Fetching recent changes.Traceback (most recent call last):
  File "/home/moorepants/miniconda/bin/asv", line 9, in <module>
    load_entry_point('asv', 'console_scripts', 'asv')()
  File "/home/moorepants/src/asv/asv/main.py", line 38, in main
    result = args.func(args)
  File "/home/moorepants/src/asv/asv/commands/__init__.py", line 48, in run_from_args
    return cls.run_from_conf_args(conf, args)
  File "/home/moorepants/src/asv/asv/commands/run.py", line 123, in run_from_conf_args
    **kwargs
  File "/home/moorepants/src/asv/asv/commands/run.py", line 166, in run
    commit_hashes = util.pick_n(commit_hashes, steps)
  File "/home/moorepants/src/asv/asv/util.py", line 650, in pick_n
    spaced.append(items[int(i)])
TypeError: 'set' object does not support indexing
```

moorepants:
I think this has to do with the `-s` flag.

pv:
Fix in #473

PR #473, opened 2016-07-31
Fix asv run --steps NEW and behavior of Repo.get_new_branch_commits

pv:
Fixes gh-471


pv:
Merging (straightforward bugfix)

PR #474, opened 2016-08-11
Safer git usage in gh-pages

pv:
Instead of checking out gh-pages in current repository, pull the branch
from a separate repository. Add a test.

`_has_staged_changes` is not needed, since there are now no commands
changing the index.


pv:
Merging, seems to work.

Issue #481, opened 2016-10-28
RFC: enhanced benchmark discovery


philpep:
Currently benchmark discovery (and categorization in web ui) is done by introspecting python files in `benchmark_dir` and looking for `__name__` of functions and classes and module name relative to `benchmark_dir`.

This make hard to generate benchmark function programmatically because we have to take care of `__name__` and create a module for each category and separate the entire benchmark code in a specific directory. Also changing the function name or moving it to another module requires to update all benchmark results.

I wonder if we could use the same logic than https://github.com/celery/celery for discovering tasks eg:

  * `benchmark_load` config could be a list of importables modules
  * autoregister functions ``<import_name_relative_to_pythonpath>.module.Class.function``
  * support named tasks (eg by adding a `name` attribute on the function). Out of scope of this issue but a decorator to handle benchmark attributes and declaration may be more pythonic
  * categorization is handled with dots in benchmark name like the current behavior.

What do you think ?

After reviewing my list, I think just adding named benchmark is a good start (and backward compatible), I try to write a PR for this.

pv:
I would like to keep it simple here, and not offer customization options on the benchmark discovery, unless the use cases are clearly necessary. I think it's fine for asv to have an opinion on how benchmark suites should be structured.

A `benchmark_name` attribute perhaps is motivated, as the name encodes not only the benchmark but also the benchmark results which may be useful to keep for backward compatibility. A question however is whether the presence of this attribute should be taken to mean that the object in question is a benchmark. Probably yes, so the attribute should be `benchmark_name` instead as `name` is likely too ambiguous.

anntzer:
It would be great if asv could rely on pytest to perform benchmark discovery/collection, as this would e.g. also allow using pytest parametrization and fixtures, as pytest-benchmark (http://pytest-benchmark.readthedocs.io/en/stable/usage.html) does.  See e.g. https://github.com/anntzer/mpl_cairo/blob/master/tests/test_speed.py for a case where parametrization makes sense.

(While I think pytest-benchmark is much better at benchmark collection, its reporting/postprocessing facilities are nowhere as good as asv's).

pv:
Current situation: https://asv.readthedocs.io/en/stable/benchmarks.html#general
`benchmark_name` is added, probably sufficient for most benchmark autogeneration, with suitable amount of work.

pytest-benchmark integration is #567, I'm not planning to work on implementing/maintaining it however at the moment.

PR #484, opened 2016-10-28
Support for named benchmarks

philpep:
Related to #481

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -239,13 +239,6 @@ def _get_first_attr(sources, name, default, ignore_case=False):
     return default
 
 
-def get_benchmark_type_from_name(name):
-    for bm_type in benchmark_types:
-        if bm_type.name_regex.match(name):
-            return bm_type
-    return None
-
-
 def get_setup_cache_key(func):
     if func is None:
         return None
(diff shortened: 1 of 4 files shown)


pv:
I guess this is mostly OK. Comments:

- I think the attribute name should be the more specific `benchmark_name` instead of `name`
- The benchmark selection now needs to import the whole benchmark suite for each benchmark run, which may be expensive. Would it make sense in `get_benchmark_from_name` to first try importing the benchmark function based on its name, and only if that fails fall back to scrounging with `disc_benchmarks`?

philpep:
Sorry for the delay.

Pushed new commits with ``benchmark_name`` attribute instead of ``name``.

The current implementation of benchmark discovery by name use a generator, so it doesn't always import the whole suite. Nevertheless I re-implemented direct import of benchmark in https://github.com/spacetelescope/asv/pull/484/commits/9fb848ed0796a2cccd0924e4dff2f2e8ff85f40a
I'd prefer to write it shorter but didn't find a way. My opinion is that this is not worth to have this complexity since import of the benchmark suite should be insignificant comparing to the benchmark run. Let me know if you prefer to drop it and I'll update the PR.

Thanks !

pv:
The benchmark suite import time can be similar to the benchmark runtime. 
For example, for the scipy benchmark suite, it's 0.5 sec vs. ~0.1 sec 
per-file. For pandas, it's ~ 1.0 sec total vs. 0.3 sec per-file (both 
with hot disk cache here).

pv:
Note that the 'benchmarks' package name should be replaced by dirname(root).

philpep:
Good catch ! I updated the commit.

PR #489, opened 2016-11-16
ENH: Terminate process group on timeout under Windows.

wrwrwr:
Consider a subsubprocess created within a benchmark that times out. The current behavior is:
* POSIX: the benchmark subprocess group is terminated (including the subsubprocess);
* Windows: just the benchmark subprocess is terminated.

Under Windows this can lead to the main process hanging on [`stderr_reader.join()`](https://github.com/spacetelescope/asv/blob/a14d8dc70442faef6e8b4eacb60ffc9c3d488fa6/asv/util.py#L413) (as the subsubprocess may inherit file descriptors).

This change should ensure that subsubprocesses are terminated on timeout also under Windows. The implementation uses console events as that seems to be more reliable than `taskkill` or `psutil.children()` based solutions. This is a bit specific, so some additional testing under Windows would be appreciated.

Some references:
* a nice write-up on [Python and signals in Windows](https://bugs.python.org/issue26350#msg260201);
* MSDN on [`CREATE_NEW_PROCESS_GROUP`](https://msdn.microsoft.com/pl-pl/library/windows/desktop/ms684863(v=vs.85).aspx#CREATE_NEW_PROCESS_GROUP) and [`CTRL_BREAK_EVENT`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682541(v=vs.85).aspx);
* [the commit](https://github.com/spacetelescope/asv/commit/32db52d954bb06921ba171aee60f900d9914ab88) in which the POSIX counterpart was introduced;
* the AppVeyor failure in #487 is most likely due to this issue.

### DIFF
--- a/asv/util.py
+++ b/asv/util.py
@@ -341,26 +341,23 @@ def check_output(args, valid_return_codes=(0,), timeout=600, dots=True,
 
     log.debug("Running '{0}'".format(' '.join(args)))
 
-    posix = getattr(os, 'setpgid', None)
-    if posix:
-        # Run the subprocess in a separate process group, so that we
-        # can kill it and all child processes it spawns e.g. on
-        # timeouts. Note that subprocess.Popen will wait until exec()
-        # before returning in parent process, so there is no race
-        # condition in setting the process group vs. calls to os.killpg
-        preexec_fn = lambda: os.setpgid(0, 0)
+    kwargs = dict(shell=shell, env=env, cwd=cwd,
+                  stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    if WIN:
+        kwargs['close_fds'] = False
+        kwargs['creationflags'] = subprocess.CREATE_NEW_PROCESS_GROUP
     else:
-        preexec_fn = None
+        kwargs['close_fds'] = True
+        posix = getattr(os, 'setpgid', None)
+        if posix:
+            # Run the subprocess in a separate process group, so that we
+            # can kill it and all child processes it spawns e.g. on
+            # timeouts. Note that subprocess.Popen will wait until exec()
+            # before returning in parent process, so there is no race
+            # condition in setting the process group vs. calls to os.killpg
+            kwargs['preexec_fn'] = lambda: os.setpgid(0, 0)
 
-    proc = subprocess.Popen(
-        args,
-        close_fds=(not WIN),
-        env=env,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-        shell=shell,
-        preexec_fn=preexec_fn,
-        cwd=cwd)
+    proc = subprocess.Popen(args, **kwargs)
 
     last_dot_time = time.time()
     stdout_chunks = []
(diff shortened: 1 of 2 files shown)


asv/util.py
```diff
@@ -341,26 +341,28 @@ def get_content(header=None):
...
 
-    posix = getattr(os, 'setpgid', None)
-    if posix:
-        # Run the subprocess in a separate process group, so that we
-        # can kill it and all child processes it spawns e.g. on
-        # timeouts. Note that subprocess.Popen will wait until exec()
-        # before returning in parent process, so there is no race
-        # condition in setting the process group vs. calls to os.killpg
-        preexec_fn = lambda: os.setpgid(0, 0)
+    kwargs = {
```
pv: I'd put this in the `kwargs = dict(env=env, ...)` syntax.

asv/util.py
```diff
@@ -392,7 +394,7 @@ def watcher_run():
                 time.sleep(0.1)
                 if time.time() - start_time[0] > timeout:
                     was_timeout[0] = True
-                    proc.terminate()
+                    proc.send_signal(signal.CTRL_BREAK_EVENT)
```
pv: Can processes ignore that signal? If yes, is there a way to force termination (like sigkill on posix)?

wrwrwr: Processes *can* ignore `CTRL_BREAK_EVENT` by registering a handler for `SIGBREAK` (albeit -- at this moment -- this isn't documented neither on [MSDN](https://msdn.microsoft.com/en-us/library/xdkz3x12.aspx#Anchor_2) nor for [Python](https://docs.python.org/3.6/library/signal.html#signal.signal)).

However, as far as I can tell, there isn't much more that we could do:
* [`CTRL_CLOSE_EVENT`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683242(v=vs.85).aspx#CTRL_CLOSE_EVENT) would be more like `SIGTERM` followed by `SIGKILL` (forced termination after a delay of a few seconds), but there doesn't seem to be a way to trigger it programmaticaly;
* [`TerminateProcess`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686714(v=vs.85).aspx) is a nearly instant (possibly even overly brutal) way to end a process, but the only way to enumerate processes in a group seems to be through a [snapshot](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682489(v=vs.85).aspx).

Moreover, termination can always be escaped by creating a new group (also under Linux).

I've been thinking about adding a note under [benchmark attributes](http://asv.readthedocs.io/en/latest/writing_benchmarks.html#benchmark-attributes), something like:

> We do our best to terminate the whole benchmark on timeout, but if you create process groups within the benchmark, trap `SIGBREAK` or handle `CTRL_BREAK_EVENT` under Windows, your subprocesses may survive.

Or would this be too detailed?

pv:
Thanks, merged!

I think it's best to not worry too much about the edge case here (processes ignoring signals), which is probably going to be rare.

PR #493, opened 2016-12-09
Benchmark statistics

pv:
Record more benchmark samples, and compute, display, and use statistics based on them.

- Change default settings to record more (but shorter) samples. <s>Changes meaning of `goal_time`</s> Makes determination of `goal_time` more accurate; however, no big changes to methodology --- there's room to improve here.
- Do warmup before benchmark --- this seems to matter also on CPython (although possibly not due to CPython itself but some OS/CPU effects)
- Display spread of measurements (median +/- half of interquartile range)
- Estimate confidence interval and use that to decide what is not significant in `asv compare`.
- Optionally, save samples to the json files.
- <s>Switch to gzipped files.</s>

The statistical confidence estimates are a somewhat tricky point, because timing samples usually have strong autocorrelation (multimodality, stepwise changes in location, etc.), which makes simple stuff often misleading. There's some alleviation for this currently there in that it tries to regress the timing sample time series looking for steps, and adds those to the CI. Not rigorous, but probably better than nothing.

The problem is that there's low-frequency noise in the measurement, so measuring for a couple of second does not give a good idea of the full distribution.

todo:

- [x] more tests, e.g. printing and formatting is mostly untested
- [x] do we really need to gzip?
- [x] update documentation
- [x] tuning for pypy

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -431,9 +431,12 @@ class TimeBenchmark(Benchmark):
 
         start_time = time.time()
 
+        max_time = start_time + min(warmup_time + 1.3 * repeat * goal_time,
+                                    self.timeout - 1.3 * goal_time)
+
         def too_slow():
             # too slow, don't take more samples
-            return time.time() > start_time + warmup_time + 2 * repeat * goal_time
+            return time.time() > max_time
 
         if number == 0:
             # Select number & warmup.
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -16,7 +16,8 @@ its runtime to a file.
 # there although it's not part of the package, and Python puts it to
 # sys.path[0] on start which can shadow other modules
 import sys
-sys.path.pop(0)
+if __name__ == "__main__":
+    sys.path.pop(0)
 
 import copy
 try:
@@ -35,6 +36,7 @@ import pickle
 import re
 import textwrap
 import timeit
+import time
 
 # The best timer we can use is time.process_time, but it is not
 # available in the Python stdlib until Python 3.3.  This is a ctypes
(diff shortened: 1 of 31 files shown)


wrwrwr:
Changing `goal_time` to mark a single repeat time is nice, but do I read it right that setting `repeat = 5` will take 5 samples if `number` is fixed and 50 samples if it is left at zero? Why not make the number of samples equal to `repeat` in the adaptive case too (with a single setup per sample and a higher default)?

pv:
Yes, I've got something like that in the works. This is partly 
orthogonal to this PR, however.
.
The benchmark methodology may still need tuning. One issue faced here 
not faced by perf is that the benchmarks that run later, run under 
different CPU/systemload/heating conditions than the benchmarks that run 
earlier. So the sampling probably should use multiple processes, and be 
done in A B C ... A' B' C' ... A'' B'' C'' ... order.

wrwrwr:
Interleaving benchmarks would surely be good, comparability across separate import benchmarks was something I was aiming for. The whole samples/repeats logic would probably need to be moved up to realize that though (with a "sample" rather than a "benchmark" script?).

pv:
I think this is more or less finished now. Some fine-tuning may be useful to do later on. 

The changes in benchmarking parameter default values may be annoying, but these should give better accuracy. The default runtime also becomes shorter. Big improvements on the accuracy probably will need stuff such as splitting the benchmark runs into several parts, to be run in an interleaved order to ensure long sampling time spans to capture the low-frequency noise properly. https://github.com/pv/asv/tree/many-proc

pv:
After some dogfooding, I think this is ready to go.

PR #495, opened 2016-12-22
Replace forward slash with underscore in env name.

### LINKED ISSUE #494: Allow dependency to be installed from git
Would be great to also support setting a dependency to install pip install from git like the below.  Some projects are built in tandem and only work with specific dependencies that may only be available as commits.

`pip install git+https://github.com/repo/project.git`

I found that this can already be done by setting the `matrix` config in asv.conf.json to:

```
{
  'pip+git+https://github.com/repo/project.git': []
}
```

but the directory structure for results gets messed up.  It causes the results to get put in nested directories presumably because of a split on `/` is done on the key.  This results in asv not able to find `machine.json` when doing `asv publish`.

postelrich:
Replace forward slash in env name to prevent subdirectories getting created.

Resolves #494

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -179,12 +179,11 @@ def get_env_name(tool_name, python, requirements):
     reqs = list(six.iteritems(requirements))
     reqs.sort()
     for key, val in reqs:
-        key = key.replace('/', '_')
         if val:
             name.append(''.join([key, val]))
         else:
             name.append(key)
-    return '-'.join(name)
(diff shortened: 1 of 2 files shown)


pv:
Syntax errors.

pv:
Still syntax errors. Did you check if this change fixes the problem you had?

pv:
Thanks, done in gh-498 in a more complete way.

PR #504, opened 2017-02-19
In web regression listing, show the date of the last (not first) regression

pv:
Ordering by most recent regression is likely the most sensible one.
This also makes the date ordering of items displayed on web more closely match the RSS feed.

Also fix a bug producing completely incorrect date formatting (Date.getDay vs Date.getDate...).


pv:
Seems to work fine.

PR #509, opened 2017-04-15
Benchmark result versioning

pv:
Add a version attribute to benchmarks, and when benchmarks are run,
tag the results produced with it. The default value is based on hash of
source code lines, but can be overridden by user.

In run, consider results with different benchmark version as missing.
In compare, mark incompatible results.
In publish, ignore results with different benchmark versions.


pv:
Seems to work ok in dogfooding, so planning to merge.

PR #512, opened 2017-05-10
DOC: comment out not implemented but logical to have teardown_cache

### LINKED ISSUE #511: teardown_cache is "exampled" but not implemented
#277 added support for `setup_cache` and I see example also of the matching `teardown_cache` in test/benchmark/cache_examples.py but unfortunately support for it is actually not implemented. So if I am creating some reusable by all benchmarks resource (e.g. temporary file), the only way I could get rid of it whenever I am done I guess is to use atexit callback

yarikoptic:
Closes #511

### DIFF
--- a/test/benchmark/cache_examples.py
+++ b/test/benchmark/cache_examples.py
@@ -11,9 +11,6 @@ class ClassLevelSetup:
     def setup_cache(self):
         return [0] * 500
 
-    def teardown_cache(self):
-        pass
-
     def track_example(self, big_list):
         return len(big_list)
 


pv:
I think you can probably remove the function instead, it is probaly 
leftover from an earlier design.

yarikoptic:
it was more of a hint that may be it is worth implementing it eventually ? ;)  e.g. to fulfill the cases beyond temp file in cwd (establish DB, start some docker container, etc)

pv:
Yes, it could possibly be useful for such external state control. 
Doctrine however forbids commented-out code :)

yarikoptic:
ok - here you go

pv:
thanks, merged (busy times...).

PR #521, opened 2017-06-20
chmod html_dir to ensure writability

Warbo:
asv's HTML output is made by copying a "template" directory, which lives
alongside asv's code (e.g. in /usr or the like), to the destination, which is
then modified to contain the benchmark data.

If the template is read-only, e.g. due to the filesystem it's stored on, then we
can copy it across, but don't have permission to modify it (e.g. getting OSError
13 "permission denied" when trying to make the graph/ subdir).

This change runs chmod to grant write permission to the owner (the current user)
immediately after the copy, so we avoid such "permission denied" errors.


pv:
Should use a for loop, not generator expression.

PR #525, opened 2017-06-24
BUG: util.which should also recognize .bat

pv:
For example conda can in sub-environments become a bat file.


pv:
superceded by gh-535

Issue #523, opened 2017-06-24
Support venv on python>=3.3


anntzer:
I know that the docs state a dependency on
> virtualenv, 1.10 or later (this is true even with Python 3.3, where virtualenv is included as venv, since venv is not compatible with other versions of Python).

but would it be possible to just replace
```
import virtualenv
```
by
```
try: import venv as virtualenv
except ImportError: import virtualenv
```
everywhere, so that virtualenv becomes unneeded on Python>=3.3?

(Basically, I don't know if the two have APIs compatible enough for asv.)

pv:
Want to try if it works? Patches are welcome in principle. All relevant
code is in asv/plugins/virtualenv.py

anntzer:
Ah, I think that won't work because you want to be able to use the *current* python to create virtualenvs for *other* pythons, which venv doesn't support.

pv:
I think that's probably not so crucial, it would probably also be fine to shell out to a subprocess.

pv:
The workaround to the nested virtualenv/venv case apparently requires finding the "real" Python executable: from https://github.com/berdario/pew/blob/c25f19e9bb0bc94f696407378fc104706b5dd3b5/pew/_venv.py#L68-L104
```
def find_real_python():
    # We need to run venv on the Python we want to use. sys.executable is not
    # enough because it might come from a venv itself. venv can be easily
    # confused if it is nested inside another venv.
    # https://bugs.python.org/issue30811
    python = ('python.exe' if windows else 'python')


    # If we're in a venv, excellent! The config file has this information.
    pyvenv_cfg = pathlib.Path(sys.prefix, 'pyvenv.cfg')
    if pyvenv_cfg.exists():
        return Cfg(pyvenv_cfg).bindir.joinpath(python)


    # Or we can try looking this up from the build configuration. This is
    # usually good enough, unless we're in a virtualenv (i.e. sys.real_prefix
    # is set), and sysconfig reports the wrong Python (i.e. in the virtualenv).
    import sysconfig
    bindir = sysconfig.get_config_var('BINDIR')
    try:
        real_prefix = sys.real_prefix
    except AttributeError:
        return pathlib.Path(bindir, python)
    if not os.path.realpath(bindir).startswith(real_prefix):
        return pathlib.Path(bindir, python)


    # OK, so we're in a virtualenv, and sysconfig lied. At this point there's
    # no definitive way to tell where the real Python is, so let's make an
[...]

PR #526, opened 2017-06-24
Use venv instead of virtualenv when available.

anntzer:
Attempt at fixing #523 as suggested by @pv.

### DIFF
--- a/asv/plugins/virtualenv.py
+++ b/asv/plugins/virtualenv.py
@@ -17,7 +17,8 @@ from ..console import log
 from .. import util
 
 
-WIN = (os.name == "nt")
+# We always run `python -mpip` instead of `pip` because the latter may fail to
+# get installed into nested environments (... sometimes).
 
 
 class Virtualenv(environment.Environment):
@@ -148,9 +149,9 @@ class Virtualenv(environment.Environment):
 
     def _install_requirements(self):
         if sys.version_info[:2] == (3, 2):
-            pip_args = ['install', '-v', 'wheel<0.29.0', 'pip<8']
+            pip_args = ['-m', 'pip', 'install', '-v', 'wheel<0.29.0', 'pip<8']
         else:
-            pip_args = ['install', '-v', 'wheel', 'pip>=8']
+            pip_args = ['-m', 'pip', 'install', '-v', 'wheel', 'pip>=8']
 
         if 'COV_CORE_SOURCE' in os.environ:
             # To measure coverage of ASV parts run in a subprocess from
@@ -158,11 +159,7 @@ class Virtualenv(environment.Environment):
             # to be installed for that environment.
             pip_args.append('pytest-cov')
 
-        if not WIN:
-            self.run_executable('pip', pip_args)
-        else:
-            # Run pip self-upgrade via python -m pip, so that it works on Windows
-            self.run_executable('python', ['-m', 'pip'] + pip_args)
+        self.run_executable('python', pip_args)
 
         if self._requirements:
             args = ['install', '-v', '--upgrade']
--- a/asv/plugins/virtualenv.py
+++ b/asv/plugins/virtualenv.py
@@ -52,17 +52,6 @@ class Virtualenv(environment.Environment):
         self._requirements = requirements
         super(Virtualenv, self).__init__(conf, python, requirements)
 
-        try:
-            import virtualenv
-        except ImportError:
-            raise environment.EnvironmentUnavailable(
-                "virtualenv package not installed")
-
-        # Can't use `virtualenv.__file__` here, because that will refer to a
-        # .pyc file which can't be used on another version of Python
-        self._virtualenv_path = os.path.abspath(
-            inspect.getsourcefile(virtualenv))
-
     @staticmethod
     def _find_python(python):
         """Find Python executable for the given Python version"""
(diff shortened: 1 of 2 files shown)


asv/plugins/virtualenv.py
```diff
@@ -108,24 +97,30 @@ def name(self):
...
             return False
+        executable = Virtualenv._find_python(python)
 
-        try:
-            import virtualenv
-        except ImportError:
-            return False
+        if LooseVersion(match.group(1)) < LooseVersion("3.3"):
+            self._virtualenv_argv = [
+                executable, "-mvirtualenv", "--no-site-packages"]
```
pv: Maybe it should use current python, when using virtualenv?
venv is bundled, but Virtualenv is not necessarily installed in the target Pythons...

pv:
Also, many tests seem to break.

asv/plugins/virtualenv.py
anntzer: The point I was making in https://github.com/spacetelescope/asv/issues/523#issuecomment-310868051 is that venv is unable to create virtualenvs for pythons others than the current interpreter, so if we want the master python to use venv, then any other python in the test matrix either needs to be >=3.3 (i.e., have venv itself), or have virtualenv installed.

As I only work with Py3 that restriction seems acceptable to me, but I can also understand if you disagree, in which case I don't think such an approach can ever be made to work (unless builtin venv picks up the ability to create venvs for other pythons, but why whould they?).  So, please let me know if I should try to fix the tests, or if you simply won't accept such a design (again, it's up to you, of course).

pv:
What I mean:
.
- When using venv, use it as you do. It's guaranteed to be built-in on 
new enough Python versions.
- When using virtualenv, the old approach is probably preferable.

pv:
Namely, it is not clear to me why it is necessary to the way the 
*virtualenv* package is used. The *venv* package needs to be called with 
the Python corresponding to the version it is used for, and that's fine 
as it is built-in.

pv:
Edit: "... necessary to change the way the virtualenv package ..."

anntzer:
Ah, I see, your suggestion is to add a third environment type, "venv", in addition to the two existing ones ("conda" and "virtualenv"); it would only work if the entire matrix uses recent enough Pythons.  Is my understanding correct?  (If yes that looks reasonable to me.)

pv:
What I mean that it would be possible to do it here like this:
```
if target_python_version >= 3.3:
     call target_python -mvenv ...
else:
     # Do it the way it's currently done in asv
     import virtualenv
     call current_python self._virtualenv_path -p target_python ...
```
Currently, in this PR it is done like this:
```
if target_python_version >= 3.3:
     call target_python -mvenv ...
else:
     call target_python virtualenv ...
```

anntzer:
Ah yes, thanks for the clarification.  Will work on that.

asv/plugins/virtualenv.py
```diff
@@ -108,24 +97,30 @@ def name(self):
...
             return False
+        executable = Virtualenv._find_python(python)
 
-        try:
-            import virtualenv
-        except ImportError:
-            return False
+        if LooseVersion(match.group(1)) < LooseVersion("3.3"):
+            self._virtualenv_argv = [
+                sys.executable, "-mvirtualenv", "--no-site-packages"]
```
pv: This is a classmethod, so the initialization cannot be done here.

anntzer: Even if I rename "self" to "cls" everywhere in the class (which I'm happy to do) it should not change the semantics of the method (basically even now I'm just setting the attribute on the class rather than on the attribute, which should not matter).

I cannot reproduce the failure of test_run_build_failure or of test_run_spec locally (looks like pip is not getting installed for a reason I can't fathom), they pass for me locally.  (test_web_regressions needs the selenium test driver which I don't have installed but given that it's essentially the same failure I don't think it's relevant).

pv:
Of course renaming self->cls does not change the semantics, but makes it 
less misleading.
.
Setting the attribute on the class is wrong, because many environment 
objects can exist at the same time.
.
Moreover, you are not using the correct target python here, in 
particular self._executable is not used anywhere.
.
So the test failures are all real, and should be resolved.

pv:
To maybe unpack the logic here:
.
The Virtualenv.matches classmethod is used for autodetecting which of 
the available environment classes supports a given Python version. If 
the user specified the environment type, this method will not be called.

anntzer:
I apparently can't make this work and cannot even reproduce the test failures locally, so I am giving up on this attempt.
Feel free to close the PR.

pv:
There's a "return True" missing in the virtualenv if branch.
.
The "pip not found" issue on the other hand could be ubuntu/debian 
specific. Maybe their venv does not for some reason install pip by default?
.
The python distribution platforms are quite heterogenous, so it's not so 
easy to build solutions that work everywhere.

pv:
Also, "def matches()" should return False on failed virtualenv import, 
not raise an exception.

anntzer:
Thanks for the careful suggestions.
But I still hit the same "pip not found" issue...

pv:
I think the problem is that (1) travis python is actually in a 
virtualenv, and (2) if you do
```
virtualenv -p /usr/bin/python3 fooenv
source fooenv/bin/activate
python -mvenv fooenv2
```
Then the created environment doesn't get pip installed for some reason. 
Virtualenv breaks venv. Maybe virtualenv bug, perhaps worth reporting.

anntzer:
Wow, nice find.

It looks like this actually is a problem with ensurepip(?).  venv installs pip via ensurepip while it is set up (https://github.com/python/cpython/blob/3.6/Lib/venv/__init__.py#L238), but the call to ensurepip (while nested-env is active) results in
```
Requirement already up-to-date: setuptools in ./outer-env/lib/python3.6/site-packages
Requirement already up-to-date: pip in ./outer-env/lib/python3.6/site-packages
```

pv:
Maybe can be fixed by replacing all calls to `pip` by 
`run_executable(['python', '-mpip', ...])`? Needs also corresponding 
change in Environment.check_presence (which maybe should be called after 
also after env creation in Environment.create, to check if the env 
creation was actually successful.)

anntzer:
No, pip is just not installed in the inner env (despite believing the contrary).

pv:
But it seems to work?

anntzer:
So it looks like that the outer-env pip leaks into the inner env.
```
(inner-env) $ python -minspect -d pip
Target: pip
Origin: /tmp/outer-env/lib/python3.6/site-packages/pip/__init__.py
Cached: /tmp/outer-env/lib/python3.6/site-packages/pip/__pycache__/__init__.cpython-36.pyc
Loader: <_frozen_importlib_external.SourceFileLoader object at 0x7f14adc063c8>
Submodule search path: ['/tmp/outer-env/lib/python3.6/site-packages/pip']
```
I guess that's good enough...

pv:
But it seems the venv is not actually isolated from the parent 
virtualenv, pip installs stuff in the parent env, so it's really just 
broken :(

pv:
So I guess at minimum, here it would be necessary to check if the target 
Python is comes from a virtualenv and if yes, never use venv. Or wait 
until CPython/virtualenv fixes their stuff. Sounds messy.

anntzer:
http://bugs.python.org/issue30811

I think accepting that virtualenv is a dependency even on recent Pythons (if not using conda) is a good enough compromise for me...

pv:
Yes, I think the simplest solution is to not use venv for the present. 
The CPython release cycle probably means it will take some time until 
that is fixed.

pv:
Note that your bug report seems inconsistent --- e.g. uses python2 
-mvirtualenv without -p flag to create py3 env, and the -minspect output 
shows the import paths are in inner-env

anntzer:
That was a very sloppy report, I updated it.

pv:
No activity: closing.

PR #534, opened 2017-07-21
[WIP] ENH: Allow asv continuous to use a branch not in JSON `branches`

tylerjereddy:
### Background / Motivation
For some projects it would be useful to be able to run `asv continuous` between a (commit hash on) standard `master` branch that can be listed in the JSON `branches` parameter (i.e., the branch the project mainly uses on github) and a (commit hash on) new feature branch from a given developer (in i.e., a github PR) from their **new** branch that is **not** included in the `branches` parameter of the JSON file.

If I'm not mistaken, this isn't currently possible if you want to check the performance between i.e., the tip of master and the most recent commit hash of the new feature branch not listed in the JSON file. 

It might be possible to manually edit the JSON file for each new feature branch (add its name), but that seems rather awkward, right?

### Implementation in this PR
I've just hacked something together to support the above functionality based on a few hours of looking at the asv source code, so there is almost certainly a far better way to do this, if there's even general interest in doing this at all.

If there might be interest in supporting this, I'm hoping that I can at least catalyze some criticism / feedback on doing this more appropriately based on this initial hack that seems to work in my early / crude tests.

I currently run a command as below using this 'new feature:'
`asv continuous --bench GROReadBench 175aba4b 3e16993 -e --extrabranch 3e16993`

Where the new `extrabranch` provides a commit hash from the new feature branch that is not listed in the JSON file (ok, even if you like this feature / idea, the API is almost certainly going to need to change).

Suggestions? Even if `asv` itself doesn't want to support this, one project I work on will likely want to maintain a fork that does because of the development workflow, so any suggestions are most welcome.

### DIFF
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -309,7 +309,7 @@ class Benchmarks(dict):
     """
     api_version = 1
 
-    def __init__(self, conf, repo, environments, benchmarks=None, regex=None):
+    def __init__(self, conf, repo, environments, commit_hash, benchmarks=None, regex=None):
         """
         Discover benchmarks in the given `benchmark_dir`.
 
@@ -333,7 +333,7 @@ class Benchmarks(dict):
         self._benchmark_dir = conf.benchmark_dir
 
         if benchmarks is None:
-            benchmarks = self.disc_benchmarks(conf, repo, environments)
+            benchmarks = self.disc_benchmarks(conf, repo, environments, commit_hash)
         else:
             benchmarks = six.itervalues(benchmarks)
 
@@ -349,7 +349,8 @@ class Benchmarks(dict):
                 self[benchmark['name']] = benchmark
 
     @classmethod
-    def disc_benchmarks(cls, conf, repo, environments):
+    def disc_benchmarks(cls, conf, repo, environments,
+                        commit_hash):
         """
         Discover all benchmarks in a directory tree.
         """
@@ -374,7 +375,7 @@ class Benchmarks(dict):
         log.info("Discovering benchmarks")
         with log.indent():
             env.create()
-            env.install_project(conf, repo)
+            env.install_project(conf, repo, commit_hash)
 
             result_file = tempfile.NamedTemporaryFile(delete=False)
             try:
--- a/asv/commands/continuous.py
+++ b/asv/commands/continuous.py
@@ -43,6 +43,9 @@ class Continuous(Command):
             run only once.  This is useful to find basic errors in the
             benchmark functions faster.  The results are unlikely to
             be useful, and thus are not saved.""")
+        parser.add_argument(
+            "--extrabranch", default=None,
+            help="""Name of an extra branch to consider if not in JSON config file.""")
         common_args.add_factor(parser)
         common_args.add_show_stderr(parser)
         common_args.add_bench(parser)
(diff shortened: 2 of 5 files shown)


pv:
"asv continuous" can take two commit arguments, so if you know the 
commit hashes --- or even branch names -- you should be able to compare 
them. These commits do not need to be on any branch --- indeed the 
branches listed in the JSON config should not play any role here. So 
before commenting on your suggested change, can you clarify what goes 
wrong? Why are changes necessary?

pv:
From the code changes, it appears you want to control the version 
installed when benchmark *discovery* is run. This should not matter for 
a correctly structured benchmark suite, as the no benchmark will be run 
using this version.
.
However, the assumption in asv is that the benchmark suite must be 
importable for any version you wish to benchmark against. This 
requirement is largely not negotiable, so imports need be protected with 
try-except blocks.

tylerjereddy:
The main branch of the project (the only one the JSON file knows about) has two issues: `setup.py` is in the wrong place and it has no benchmark suite yet. The feature branch has a few commits -- one moves `setup.py` to the place `asv` expects and another adds the `asv` benchmark suite (JSON file, a simple benchmark to run, etc.) to the repo.

If I run `asv continuous` on two commits in the feature branch, asv will still apparently use a commit hash that precedes the feature branch even though I've only specified comparison of commit hashes in the feature branch; so asv continuous is perhaps discovering the benchmarks to run from a parent of the commit hashes (apparently the tip of the branch specified in the JSON file). I was trying to get benchmark discovery to happen from the first chronological commit hash provided to asv continuous because the 'master' branch is not compatible with asv yet.

However, if the JSON file is edited to contain the name of the feature branch instead of the master branch, it will use the feature branch that is compatible with asv and that contains the newly-included suite / benchmarks & everything works as expected. It doesn't seem to have complete isolation to the commit hashes that are provided to asv continuous (discovery happens beyond those provided commit hashes and 'master' was never intended for use with asv).

So while introduction of the asv suite to a project might seem like an extreme example, doesn't this issue also exist if you write a new benchmark that doesn't exist in master on a new feature branch? Can you discover a benchmark test that doesn't exist in master (introduced in feature branch with new name that is not listed in the JSON file), and run that?

So: 
master branch commit hash a57381 -- doesn't have the new benchmark test
new feature branch commit hash a57382 -- has the new benchmark test written in
new feature branch commit hash a57383 -- makes a potential performance improvement to code being benchmarked

Then run `asv continuous` with JSON file only knowing about the master branch -- it seems master gets checked out for discovery and you can't run the benchmark introduced in the feature branch, right?

If my assumptions / descriptions above sound consistent with your expectations, perhaps this can simply be summarized as: 
1. Changes (i.e., new tests) to the benchmarking suite should be merged to the 'discovery' / master branch (listed in JSON file) **first**.
2. After a new benchmark test has been merged to the 'discovery' / master branch, it is then safe to open a new feature branch that attempts to alter the target benchmark code and check performance there, because the discovery can safely happen from (the tip of) the JSON-listed master.

Perhaps supporting flexible test discovery (from i.e., a feature branch not listed in JSON file) is more confusing than it is worth & the above workflow is the intended approach anyway? If so, I guess this should be closed and I can perhaps focus more on the issue of retroactive testing when the repo had setup.py in the wrong place for years worth of commits (which has more to do with supporting a custom build / make process for `asv` as I understand it).

pv:
The issue does not affect adding new benchmarks --- asv always uses the 
benchmark suite is present when running "asv continuous" etc.
.
To avoid confusion --- asv installs different versions of the *project*. 
It checks out the specified repository to a temporary directory and 
builds it, installing to a virtualenv. It never changes anything in the 
current working tree.
.
The benchmark suite is completely separate object, and its versioning is 
**not** managed by asv. The benchmark suite does not need to be in the 
same repository as the project. It also does not even need to be version 
controlled.
.
If you are working on a feature branch, you can check out the version of 
the benchmark suite you want to use.
.
It is correct that the benchmark discovery will not work if the branches 
defined in the JSON file are not installable. However, I'm not so sure 
if this is a common occurrence that needs a workaround.

pv:
More flexible project installation would be welcome addition, e.g. 
finding setup.py in subdirectory or using a non-Python build system.
.
The problem with this is how to do it so that it does not bring in a 
large amount of complexity. One simple alternative could be to add 
"shell_install", "shell_uninstall", and "shell_setup_env" options 
containing command lines to execute, to be run instead of "pip install 
.", "pip uninstall $PROJECT", and "python -mvirtualenv ..." for 
installing/uninstalling/setting up the project. This then would be an 
additional setup target in addition to Virtualenv and Conda.

tylerjereddy:
Thanks for the clarifications.

> It is correct that the benchmark discovery will not work if the branches defined in the JSON file are not installable. 

Since the branch defined in the JSON has nothing to do with the commit hashes I'm feeding in to `asv continuous`, I guess I'm just not understanding why `asv` is even using it--why is `asv` trying to install a commit hash of the project that I don't want to benchmark during the "discovery" process? 

If you can put the benchmarks outside of the project, why can't `asv` just read those & only touch the commit hashes that are specified?

pv:
asv wants to make sure *some* version of the project is installed in the virtualenv, when it imports the benchmark suite to crawl through the available tests. It should not matter what version gets installed, so it just picks something --- the versions specified in the `branches` are usually available, so it uses them by default.

It could be better design to make it so that the benchmark suite must be importable without the project being installed in the environment --- but this would be a backward compatibility break.

tylerjereddy:
@pv Thanks, closing now that I understand what is going on with the third commit hash.

PR #535, opened 2017-07-24
[WIP] ENH: Initial changes needed to get asv working with package/setup.py

tylerjereddy:
I managed to make a minimal set of hacks to `asv` that allow for benchmarking with `asv continuous`  when the project `setup.py` is in a subdirectory instead of the project root. This is obviously a work in progress based on a hack for a single project with a single specified / hard-coded subdirectory for the (conda) build.

Perhaps over time & with some guidance I can help make this more general so that it may be suitable for `asv` itself and not just our custom fork.

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -498,7 +498,8 @@ class Environment(object):
     def build_project(self, repo, commit_hash):
         self.checkout_project(repo, commit_hash)
         log.info("Building for {0}".format(self.name))
-        self.run(['setup.py', 'build'], cwd=self._build_root)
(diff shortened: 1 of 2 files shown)


pv:
Do you mind if we close this for the moment (in order to not clog the appveyor queue --- I don't have permissions to cancel jobs there) --- I think the approach to take will need some thinking how to do, and this PR is not yet very close to a solution.

tylerjereddy:
@pv Agreed, this is likely to require some thought & I was a bit too excited opening the PR. Closing myself. The other project I work on will use the crude hack in its own fork for now.

PR #538, opened 2017-08-03
Add asv/__main__.py

asmeurer:
That way python -m asv works.


pv:
Can you copypaste the `from __future__`, coding utf-8, and license blurb from the other files?

asmeurer:
Done.

pv:
thanks, merged

Issue #405, opened 2016-05-03
Specifying conda channels in ``matrix``?


cgranade:
Is it possible to specify a `conda` channel other than the default in the `matrix` key of `asv.conf.json`? I'm benchmarking a project which depends on CVXOPT, which suggests installing using `conda install -c https://conda.anaconda.org/omnia cvxopt`. Thanks!

anirban89:
👍 this would be useful for our project [pyqg](https://github.com/pyqg/pyqg) as well

pv:
Was implemented #539

PR #539, opened 2017-08-03
ENH: enable specifying custom conda channels + create env via environment.yml

tylerjereddy:
- This PR adds the basic functionality to specify a list of custom conda channels using a new key in the JSON config file; the associated feature request issue is #405 
- This actually looks fairly clean to me (and works for me), but there are probably some things that the core devs will know that I haven't thought of
- I suspect the way the channels are added may need slight modification -- installing packages is done in a single call to conda for performance reasons, and we may wish to do the same for adding channels (I currently loop over the input channels as you can see)
- need for tests? -- make sure this works when adding a single vs. multiple custom channels? mock the call to those channels? how much of a pain would this be to do? Maybe there's something simpler that can be done test-wise.

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -416,7 +416,9 @@ class Environment(object):
 
         for executable in ['pip', 'python']:
             exe = self.find_executable(executable)
-            if not os.path.isfile(exe):
+            # do not count a system install of pip or python
+            # as evidence of an installed env
+            if not os.path.isfile(exe) or (self._path not in os.path.dirname(exe)):
                 return False
 
         try:
@@ -536,17 +538,15 @@ class Environment(object):
         """
 
         # Assume standard virtualenv/Conda layout
-        if WIN:
-            executable += ".exe"
+        # account for known posix / Windows subfolders
+        paths = [self._path]
+        possible_subfolders = ['Scripts', 'bin']
+        for subfolder in possible_subfolders:
+            paths.append(os.path.join(self._path, subfolder))
 
-            exe = os.path.join(self._path, 'Scripts', executable)
-            if os.path.isfile(exe):
-                return exe
-            exe = os.path.join(self._path, executable)
-            if os.path.isfile(exe):
-                return exe
+        executable = util.which(executable, paths)
 
-        return os.path.join(self._path, 'bin', executable)
+        return executable
 
     def run_executable(self, executable, args, **kwargs):
         """
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -415,10 +415,9 @@ class Environment(object):
             return False
 
         for executable in ['pip', 'python']:
-            exe = self.find_executable(executable)
-            # do not count a system install of pip or python
-            # as evidence of an installed env
-            if not os.path.isfile(exe) or (self._path not in os.path.dirname(exe)):
+            try:
+                self.find_executable(executable)
+            except IOError:
                 return False
 
         try:
@@ -535,18 +534,19 @@ class Environment(object):
     def find_executable(self, executable):
         """
         Find an executable (eg. python, pip) in the environment.
+
+        If not found, raises IOError
         """
 
         # Assume standard virtualenv/Conda layout
-        # account for known posix / Windows subfolders
-        paths = [self._path]
-        possible_subfolders = ['Scripts', 'bin']
-        for subfolder in possible_subfolders:
-            paths.append(os.path.join(self._path, subfolder))
+        if WIN:
+            paths = [self._path,
+                     os.path.join(self._path, 'Scripts'),
+                     os.path.join(self._path, 'bin')]
+        else:
+            paths = [os.path.join(self._path, 'bin')]
 
-        executable = util.which(executable, paths)
-
-        return executable
+        return util.which(executable, paths)
 
     def run_executable(self, executable, args, **kwargs):
         """
--- a/asv/plugins/conda.py
+++ b/asv/plugins/conda.py
@@ -69,6 +69,7 @@ class Conda(environment.Environment):
             # doing a dry run below.  All it needs to be is something
             # that doesn't already exist.
             path = os.path.join(tempfile.gettempdir(), 'check')
+
             # Check that the version number is valid
             try:
                 util.check_call([
@@ -86,15 +87,16 @@ class Conda(environment.Environment):
 
     def _setup(self):
         try:
-            conda = self.find_executable('conda')
+            conda = util.which('conda')
         except IOError as e:
             raise util.UserError(str(e))
 
         log.info("Creating conda environment for {0}".format(self.name))
+
         # create a temporary environment.yml file
         # and use that to generate the env for benchmarking
-
-        with tempfile.NamedTemporaryFile(mode='w', delete=False) as env_file:
+        env_file = tempfile.NamedTemporaryFile(mode='w', delete=False)
+        try:
             env_file.write('name: {0}\n'
                            'channels:\n'.format(self.name))
             env_file.writelines(('   - %s\n' % ch for ch in self._conda_channels))
@@ -102,15 +104,9 @@ class Conda(environment.Environment):
                            '   - python={0}\n'
                            '   - wheel\n'
                            '   - pip\n'.format(self._python))
+
             # categorize & write dependencies based on pip vs. conda
-            reqs = self._get_requirements(conda)
-            # handle the case where there are multiple dependencies
-            # that are not actually installed (this is mostly
-            # related to tests that probe issue #169)
-            if reqs == None:
-                conda_args, pip_args = [], []
-            else:
-                conda_args, pip_args = reqs
+            conda_args, pip_args = self._get_requirements(conda)
             env_file.writelines(('   - %s\n' % s for s in conda_args))
             if pip_args:
                 # and now specify the packages that are to be installed in
@@ -118,8 +114,12 @@ class Conda(environment.Environment):
                 env_file.write('   - pip:\n')
                 env_file.writelines(('     - %s\n' % s for s in pip_args))
 
-        util.check_output([conda] + ['env', 'create', '-f', env_file.name,
-                                 '-p', self._path, '--force'])
+            env_file.close()
+
+            util.check_output([conda] + ['env', 'create', '-f', env_file.name,
+                                         '-p', self._path, '--force'])
+        finally:
+            os.unlink(env_file.name)
 
     def _get_requirements(self, conda):
         if self._requirements:
@@ -141,7 +141,7 @@ class Conda(environment.Environment):
 
             return conda_args, pip_args
         else:
-            return [],[]
+            return [], []
 
     def install(self, package):
         log.info("Installing into {0}".format(self.name))
--- a/asv/plugins/conda.py
+++ b/asv/plugins/conda.py
@@ -44,6 +44,7 @@ class Conda(environment.Environment):
         """
         self._python = python
         self._requirements = requirements
+        self._conda_channels = conf.conda_channels
         super(Conda, self).__init__(conf, python, requirements)
 
     @classmethod
@@ -85,30 +86,44 @@ class Conda(environment.Environment):
 
     def _setup(self):
         try:
-            conda = util.which('conda')
+            conda = self.find_executable('conda')
         except IOError as e:
             raise util.UserError(str(e))
 
         log.info("Creating conda environment for {0}".format(self.name))
-        util.check_call([
-            conda,
-            'create',
-            '--yes',
-            '-p',
-            self._path,
-            '--use-index-cache',
-            'python={0}'.format(self._python),
-            'pip'])
+        # create a temporary environment.yml file
+        # and use that to generate the env for benchmarking
 
-        log.info("Installing requirements for {0}".format(self.name))
-        self._install_requirements(conda)
+        with tempfile.NamedTemporaryFile(mode='w', delete=False) as env_file:
+            env_file.write('name: {0}\n'
+                           'channels:\n'.format(self.name))
+            env_file.writelines(('   - %s\n' % ch for ch in self._conda_channels))
+            env_file.write('dependencies:\n'
+                           '   - python={0}\n'
+                           '   - wheel\n'
+                           '   - pip\n'.format(self._python))
+            # categorize & write dependencies based on pip vs. conda
+            reqs = self._get_requirements(conda)
+            # handle the case where there are multiple dependencies
+            # that are not actually installed (this is mostly
+            # related to tests that probe issue #169)
+            if reqs == None:
+                conda_args, pip_args = [], []
+            else:
+                conda_args, pip_args = reqs
+            env_file.writelines(('   - %s\n' % s for s in conda_args))
+            if pip_args:
+                # and now specify the packages that are to be installed in
+                # the pip subsection
+                env_file.write('   - pip:\n')
+                env_file.writelines(('     - %s\n' % s for s in pip_args))
 
-    def _install_requirements(self, conda):
-        self.install('wheel')
+        util.check_output([conda] + ['env', 'create', '-f', env_file.name,
+                                 '-p', self._path, '--force'])
+
+    def _get_requirements(self, conda):
         if self._requirements:
-            # Install all the dependencies with a single conda command.
-            # This ensures we get the versions requested, or an error
-            # otherwise. It's also quicker than doing it one by one.
+            # retrieve and return all conda / pip dependencies
             conda_args = []
             pip_args = []
 
@@ -124,16 +139,9 @@ class Conda(environment.Environment):
                     else:
                         conda_args.append(key)
 
-            conda_cmd = ['install', '-p', self._path, '--yes']
-            pip_cmd = ['install', '-v', '--upgrade']
-
-            # install conda packages
-            if conda_args:
-                util.check_output([conda] + conda_cmd + conda_args)
-            # install packages only available with pip
-            if pip_args:
-                self.run_executable('pip', pip_cmd + pip_args,
-                                    timeout=self._install_timeout)
+            return conda_args, pip_args
+        else:
+            return [],[]
 
     def install(self, package):
         log.info("Installing into {0}".format(self.name))
--- a/asv/util.py
+++ b/asv/util.py
@@ -272,20 +272,18 @@ def human_value(value, unit, err=None):
     return display
 
 
-def which(filename):
+def which(filename, paths=None):
     """
     Emulates the UNIX `which` command in Python.
 
     Raises an IOError if no result is found.
     """
-    if WIN:
-        if not filename.endswith('.exe'):
-            filename = filename + '.exe'
-
     if os.path.sep in filename:
         locations = ['']
     else:
         locations = os.environ.get("PATH", "").split(os.pathsep)
+        if paths is not None:
+            locations = paths + locations
 
         if WIN:
             # On windows, an entry in %PATH% may be quoted
@@ -294,9 +292,17 @@ def which(filename):
 
     candidates = []
     for location in locations:
+        for alternate_ext in ['.exe', '.bat']:
+            alternate_filename = filename + alternate_ext
+            alternate_candidate = os.path.join(location, alternate_filename)
+            if os.path.isfile(alternate_candidate) or os.path.islink(alternate_candidate):
+                candidates.append(alternate_candidate)
+
+        # check for the extension-based filenames above first
         candidate = os.path.join(location, filename)
         if os.path.isfile(candidate) or os.path.islink(candidate):
             candidates.append(candidate)
+
     if len(candidates) == 0:
         raise IOError("Could not find '{0}' in PATH".format(filename))
     return candidates[0]
--- a/asv/util.py
+++ b/asv/util.py
@@ -280,31 +280,34 @@ def which(filename, paths=None):
     """
     if os.path.sep in filename:
         locations = ['']
+    elif paths is not None:
+        locations = paths
     else:
         locations = os.environ.get("PATH", "").split(os.pathsep)
-        if paths is not None:
-            locations = paths + locations
-
         if WIN:
             # On windows, an entry in %PATH% may be quoted
             locations = [path[1:-1] if len(path) > 2 and path[0] == path[-1] == '"' else path
                          for path in locations]
 
+    if WIN:
+        filenames = [filename + ext for ext in ('.exe', '.bat', '.com', '')]
+    else:
+        filenames = [filename]
+
     candidates = []
     for location in locations:
-        for alternate_ext in ['.exe', '.bat']:
-            alternate_filename = filename + alternate_ext
-            alternate_candidate = os.path.join(location, alternate_filename)
-            if os.path.isfile(alternate_candidate) or os.path.islink(alternate_candidate):
-                candidates.append(alternate_candidate)
-
-        # check for the extension-based filenames above first
-        candidate = os.path.join(location, filename)
-        if os.path.isfile(candidate) or os.path.islink(candidate):
-            candidates.append(candidate)
+        for filename in filenames:
+            candidate = os.path.join(location, filename)
+            if os.path.isfile(candidate) or os.path.islink(candidate):
+                candidates.append(candidate)
 
     if len(candidates) == 0:
-        raise IOError("Could not find '{0}' in PATH".format(filename))
+        if paths is None:
+            loc_info = 'PATH'
+        else:
+            loc_info = os.pathsep.join(locations)
+        raise IOError("Could not find '{0}' in {1}".format(filename, loc_info))
+
     return candidates[0]
 
 
--- a/asv/config.py
+++ b/asv/config.py
@@ -39,6 +39,7 @@ class Config(object):
         self.regressions_first_commits = {}
         self.regressions_thresholds = {}
         self.plugins = []
+        self.conda_channels = []
 
     @classmethod
     def load(cls, path=None):
--- a/asv/template/asv.conf.json
+++ b/asv/template/asv.conf.json
@@ -42,6 +42,10 @@
     // to the current version of Python used to run `asv`.
     // "pythons": ["2.7", "3.3"],
 
+    // The list of conda channel names to be searched for benchmark
+    // dependency packages in the specified order
+    // "conda_channels": ["conda-forge", "defaults"]
+
     // The matrix of dependencies to test.  Each key is the name of a
     // package (in PyPI) and the values are version numbers.  An empty
     // list or empty string indicates to just test against the default
--- a/docs/source/asv.conf.json.rst
+++ b/docs/source/asv.conf.json.rst
@@ -78,7 +78,8 @@ following:
 - a Python version string, e.g. ``"2.7"``, in which case:
 
   - if ``conda`` is found, ``conda`` will be used to create an
-    environment for that version of Python
+    environment for that version of Python via a temporary
+    environment.yml file
 
   - if ``virtualenv`` is installed, ``asv`` will search for that
     version of Python on the ``PATH`` and create a new virtual
@@ -93,6 +94,17 @@ following:
   fully loaded and read-only.  Thus, the benchmarked project must
   already be installed, and it will not be possible to benchmark
   multiple revisions of the project.
+
+``conda_channels``
+------------------
+A list of ``conda`` channel names (strings) to use in the provided
+order as the source channels for the dependencies. For example::
+
+    "conda_channels": ["conda-forge", "defaults"]
+
+The channels will be parsed by ``asv`` to populate the ``channels``
+section of a temporary environment.yml file used to build the
+benchmarking environment.
 
 ``matrix``
 ----------
--- a/test/test_environment.py
+++ b/test/test_environment.py
@@ -8,6 +8,7 @@ import os
 import sys
 import six
 import pytest
+import json
 
 from asv import config
 from asv import environment
@@ -106,6 +107,9 @@ def test_large_environment_matrix(tmpdir):
         # this test run a long time, we only set up the environment,
         # but don't actually install dependencies into it.  This is
         # enough to trigger the bug in #169.
+        env._get_requirements = lambda *a: None
+        # pip / virtualenv setup still uses
+        # _install_requirements
         env._install_requirements = lambda *a: None
         env.create()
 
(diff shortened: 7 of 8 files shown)


tylerjereddy:
AppVeyor succeeded; all travis tests pass except for Python 3.3, which I don't understand.

kain88-de:
> need for tests? -- make sure this works when adding a single vs. multiple custom channels? mock the call to those channels? how much of a pain would this be to do? Maybe there's something simpler that can be done test-wise.

You can check the created `~/.condarc` yaml file. As long channels are added you are fine. This also saves you from writing a mock.



> I suspect the way the channels are added may need slight modification -- installing packages is done in a single call to conda for performance reasons, and we may wish to do the same for adding channels (I currently loop over the input channels as you can see)

You can also directly modify `~/.condarc` but the calls to add a channel are usually fast and I do not expect that  people want to add 100 different channels.


Otherwise I like this addition since conda-forge packages a lot of dependencies for projects that are not available on the default channel.

kain88-de:
@tylerjereddy this will always keep the default channel. I never removed that one myself but for others it might be useful to allow removing it and only use custom channels

tylerjereddy:
@kain88-de The latest commit includes a draft unit test for verifying the addition of channels via probing the output of `conda info` -- which I think just probes the `~/.condarc` data anyway.

One major drawback of this approach is that it touches the actual `~/.condarc` on a users' machine if they run the unit test suite. Can we avoid this?

I think I also overestimated the sophistication of what is going on--although my unit test creates two separate environments for two versions of Python, all that really seems to happen is that global conda settings on a given machine are modified by asv and then used for those environments. 

I think I was originally envisioning more isolation for the environments, but if my interpretation is correct, I'm not even sure we need to create the environments for this unit test (though it may not hurt) -- it seems the main thing is just to check that the machine conda settings have been adjusted appropriately.

kain88-de:
> One major drawback of this approach is that it touches the actual ~/.condarc on a users' machine if they run the unit test suite. Can we avoid this?

You can be specifying which  channel a package should be installed from. Then you can call `conda install -c custom_channel package`. But that requires to change the way conda packages are installed and how to specify a package.

pv:
The config setting looks OK on a quick look. Add the empty list value in 
https://github.com/spacetelescope/asv/blob/master/asv/config.py#L21 and 
remove the try-except check.

Changing ~/.condarc cannot be done, so some other approach should be 
used to inform conda to what to use.

kain88-de:
Conda has a feature to [share environments ](https://conda.io/docs/using/envs.html#share-an-environment). What would prohibit us from using that feature directly? It allows to specify the channels I'm using as well as the packages with optional version tracking.  Installation can still be done in a single step here.


Here is the environment file for my own mdanalysis environment.

```
name: mda
channels:
- kain88-de
- bioconda
- conda-forge
- defaults
dependencies:
- biopython=1.70=np112py36_0
- joblib=0.9.3=py36_0
- typing=3.5.2.2=py36_0
- alabaster=0.7.10=py36_0
- apipkg=1.4=py36_0
- asn1crypto=0.22.0=py36_0
- babel=2.3.4=py36_0
- blas=1.1=openblas
- ca-certificates=2017.7.27.1=0
- certifi=2017.7.27.1=py36_0
- cffi=1.10.0=py36_0
- chardet=3.0.2=py36_1
- coverage=4.4.1=py36_0
- cryptography=1.9=py36_0
- curl=7.54.1=0
- cycler=0.10.0=py36_0
- cython=0.26=py36_0
- dbus=1.10.10=3
- decorator=4.1.2=py36_0
- docutils=0.13.1=py36_0
- execnet=1.4.1=py36_0
- expat=2.2.1=0
- fontconfig=2.12.1=4
- freetype=2.7=1
- gettext=0.19.7=1
- glib=2.51.4=0
- griddataformats=0.3.3=py36_0
- gst-plugins-base=1.8.0=0
- gstreamer=1.8.0=2
- hdf4=4.2.12=0
- hdf5=1.8.18=0
- hypothesis=3.11.6=py36_0
- icu=58.1=1
- idna=2.5=py36_0
- imagesize=0.7.1=py36_0
- ipython=6.1.0=py36_0
- ipython_genutils=0.2.0=py36_0
- jedi=0.10.2=py36_0
- jinja2=2.9.5=py36_0
- jpeg=9b=0
- krb5=1.14.2=0
- libffi=3.2.1=3
- libiconv=1.14=4
- libnetcdf=4.4.1.1=5
- libpng=1.6.28=0
- libssh2=1.8.0=1
- libtiff=4.0.6=7
- libxcb=1.12=1
- libxml2=2.9.4=4
- markupsafe=1.0=py36_0
- matplotlib=2.0.2=py36_1
- mmtf-python=1.0.7=py36_0
- msgpack-python=0.4.8=py36_0
- ncurses=5.9=10
- netcdf4=1.2.9=np112py36_0
- networkx=1.11=py36_0
- nose=1.3.7=py36_2
- numpy=1.12.1=py36_blas_openblas_200
- olefile=0.44=py36_0
- openblas=0.2.19=2
- openssl=1.0.2l=0
- pandas=0.20.3=py36_1
- patsy=0.4.1=py36_0
- pcre=8.39=0
- pexpect=4.2.1=py36_0
- pickleshare=0.7.3=py36_0
- pillow=4.2.1=py36_0
- pip=9.0.1=py36_0
- prompt_toolkit=1.0.15=py36_0
- ptyprocess=0.5.2=py36_0
- py=1.4.34=py36_0
- pycparser=2.18=py36_0
- pygments=2.2.0=py36_0
- pyopenssl=16.2.0=py36_0
- pyparsing=2.2.0=py36_0
- pyqt=5.6.0=py36_4
- pysocks=1.6.7=py36_0
- pytest=3.2.0=py36_0
- pytest-cov=2.5.1=py36_0
- pytest-xdist=1.18.2=py36_0
- python=3.6.2=0
- python-dateutil=2.6.1=py36_0
- pytz=2017.2=py36_0
- qt=5.6.2=3
- readline=6.2=0
- reportlab=3.4.0=py36_0
- requests=2.18.3=py36_0
- scikit-learn=0.18.2=np112py36_blas_openblas_200
- scipy=0.19.1=py36_blas_openblas_201
- seaborn=0.8.0=py36_0
- setuptools=36.2.2=py36_0
- simplegeneric=0.8.1=py36_0
- sip=4.18=py36_1
- six=1.10.0=py36_1
- snowballstemmer=1.2.1=py36_0
- sphinx=1.6.3=py36_0
- sphinxcontrib-websupport=1.0.1=py36_0
- sqlite=3.13.0=1
- statsmodels=0.8.0=np112py36_0
- tk=8.5.19=2
- tornado=4.5.1=py36_0
- traitlets=4.3.2=py36_0
- urllib3=1.21.1=py36_1
- wcwidth=0.1.7=py36_0
- wheel=0.29.0=py36_0
- xorg-libxau=1.0.8=3
- xorg-libxdmcp=1.1.2=3
- xz=5.2.2=0
- zlib=1.2.11=0
- libgfortran=3.0.0=1
- llvmlite=0.19.0=py36_0
- numba=0.34.0=np112py36_0
- pip:
  - ipython-genutils==0.2.0
  - prompt-toolkit==1.0.15
prefix: /home/max/anaconda3/envs/mda
```

tylerjereddy:
It looks like sharing environments requires access to an `environment.yml` file and the docs even suggest emailing that file to another person that wants to use the same environment, which doesn't seem like a suitable automated / general approach for environment generation in `asv`.

If `conda install -c channel_1 channel_2 channel_3 package`  were possible, that might be a clean approach, since you could just use the custom channels in the specified priority order without modifying condarc. 

However, it seems that you can only specify a single channel this way (i.e., the one you want to use for the specified package), so the `-c` flag approach (as I think @kain88-de suggests) would appear to require specification of a more complex dependency matrix that associates every required package with its cognate channel; the asv JSON file may get pretty gruesome with a dictionary-style association like this.

kain88-de:
> It looks like sharing environments requires access to an environment.yml file and the docs even suggest emailing that file to another person that wants to use the same environment, which doesn't seem like a suitable automated / general approach for environment generation in asv.

It should be possible to place this into the asv config file. I would also expect I can generate such a file from the current asv config files. Also with your suggested additions to specify a channel.

pv:
You should be able generate a temporary environment.yml that is used in 
the environment creation.

tylerjereddy:
Ok, so I think I understand the proposed route forward -- programmatically generate `environment.yml` from the asv config, including channels in the specified priority order.

A user could specify `nodefaults` as a channel to exclude the default channel without us having to add another JSON flag for that functionality, as documented for the `.yml` conda file.

kain88-de:
The suggested solution might break existing workflows for people who implicitly depend on the conda channels they define in their own `.condarc`. Personally I like using a environment file better because it leads to better reproducible  environments. So this change should be mentioned more explicit in the release notes.

tylerjereddy:
I started hacking away at the usage of an `environment.yml` file in the latest commit. I only get about three test failures locally, but there is definitely still a little more design / thought needed I think.

A common thing I see is this: `"CondaValueError: Value error: prefix already exists: /Users/treddy/miniconda2/envs/conda-py2.7-colorama0.3.6-six"`

And indeed, `conda env list` does get polluted with the new environments that are created during the tests. I think there's a flag to force the creation if the env already exists, but probably not the way to go.

Should we try to clean up the environment names after the tests /benchmarks run? I'm slightly concerned about the fragility of this approach--i.e., that the cleanup may not happen if an end-user has a test failure before it can happen & indeed the simple fact of modifying a users' `conda env list` is perhaps problematic in general (is there a nice trick to avoid this completely?).

Perhaps specifying the full path to the environment prefix for `conda env create` & using a temp dir of some sort could prevent polluting i.e. the local env list?

pv:
It's looks like it's creating the envs to the wrong place, they should go
to self._path. Conda create had a -p option. Can that be used with
environment.html?

29.8.2017 1.47 "Tyler Reddy" <notifications@github.com> kirjoitti:

I started hacking away at the usage of an environment.yml file in the
latest commit. I only get about three test failures locally, but there is
definitely still a little more design / thought needed I think.

A common thing I see is this: "CondaValueError: Value error: prefix already
exists: /Users/treddy/miniconda2/envs/conda-py2.7-colorama0.3.6-six"

And indeed, conda env list does get polluted with the new environments that
are created during the tests. I think there's a flag to force the creation
if the env already exists, but probably not the way to go.

Should we try to clean up the environment names after the tests /benchmarks
run? I'm slightly concerned about the fragility of this approach--i.e.,
that the cleanup may not happen if an end-user has a test failure before it
can happen & indeed the simple fact of modifying a users' conda env list is
perhaps problematic in general (is there a nice trick to avoid this
completely?).

Perhaps specifying the full path to the environment prefix for conda env
create & using a temp dir of some sort could prevent polluting i.e. the
local env list?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/spacetelescope/asv/pull/539#issuecomment-325504635>,
or mute
the thread

tylerjereddy:
Looks like it can. I'm down to two (local) test failures now: test_conda_pip_install & test_continuous. Hopefully I can take a closer look at those two issues soon.

`conda env list` no longer gets polluted by running the test suite now

tylerjereddy:
I was able to resolve one test failure related to an [open conda bug](https://github.com/conda/conda/issues/4985) where pip needs to be specified as a dependency within `environment.yml`. Actually, the latter conda bug is labelled as an easy fix, but might be a bit much at this stage to go even farther upstream fixing issues.

Anyway, that's good, we're a tiny bit closer. I note that the total number of failures in the test suite does differ a bit for the same commit hash on different operating systems sometimes, at least for this PR branch.

tylerjereddy:
I've patched all 7 test failures on a linux machine with the latest changes -- let's see how the continuous integration works out now.

tylerjereddy:
Both Travis and AppVeyor builds are passing. At this stage, I think at the very least we need:

- cleanup of affected function names and comments / docs -- in a few places I've repurposed a function without changing its name to reflect this, and some comments probably need adjustment as well
- the original unit test I added is probably out of date now -- an appropriate / updated version should be written and then the test can be reactivated
- we might want to think about one or two more test cases; indeed, there is currently a modest drop in test coverage

Still, I think the basic functionality is here now.

tylerjereddy:
Although the tests pass, as I try to write a specific test for channel addition, I'm noticing that `environment.yml` does not show up in `self._path` when I check on my machine after local tests. I can see `asv-env-info.json` there & the code to write the two files looks relatively similar to me, but there must be something tricky about writing files to the environment.

Furthermore, the environment creation command must be working for the tests to pass, and I can open the `environment.yml` file from within `plugins/conda.py` after creating it to print the contents, but not from within the test suite or when manually cd'ing to the directory in /tmp.

tylerjereddy:
I've updated the new tests to reflect the desired / enforced behavior. I note that `environment.yml` is actually obliterated from the env by the `--force` flag I used in `conda env create` to avoid certain test error scenarios where a conda env already existed. This is mostly a side-effect of the behavior that allows the unit tests to all pass & removing the `yml` file from the env is not necessarily something I was aiming to do. We could possibly take measures to restore it after if this is considered important (it may be useful residual info to have in the env folder).

One of the new unit tests verifies that the order of channels that is specified for `environment.yml` is respected when installing packages. I note that actually parsing which channels are added by `asv` to the env was rather tricky. `conda info` doesn't seem applicable (indeed I think we actually want to avoid modifying this data just as we want to avoid touching `condarc`), and I actually ended up parsing the json data from `conda list` within the env to find the source channels used to install those packages. That's the best approach I've found so far.

Let's see what the CI says now.

pv:
Note that the method names need to be changed --- 
`_install_requirements` does not actually install requirements, and 
`_add_channels` does not add channels. I'd include the `_add_channels` 
in the body of `_setup` without a separate method. What 
`_install_requirements` does sounds more like `_get_requirements`.

`mkdir_p` should not be needed, the path is guaranteed to already exist 
when `_setup` is called. If it's really needed, it should go in util.py

You can also write some of the file generation stuff simpler, e.g.,
```
+            env_file.write('dependencies:\n'
+                           '   - python={0}\n'
+                           '   - wheel\n'
+                           '   - pip\n'.format(self._python))
```
Note how strings can be split to multiple lines and that parenthesis is 
not needed before .format.

Instead of using `create --force` wouldn't it be better to create the 
environment.yml as a temporary file? Like so:
```
with tempfile.NamedTemporaryFile() as env_file:
     env_file.write(...)
     ...
     env_file.flush()
     util.check_output([conda, 'env', 'create', '-f', env_file.name] + ...)

```

tylerjereddy:
Ok, I think I've addressed most of the comments. I suspect there will still be an issue with the AppVeyor test--not exactly sure what the problem there is yet, but perhaps something quite simple involving windows paths.

tylerjereddy:
- Travis stuff mostly looks ok -- one strange failure for Python 3.5 only -- is that real?
- I adopted the proposed `NamedTemporaryFile`  solution, but I don't understand how you are planning to overcome the platform-specific limitation here -- `conda env create` won't be able to access the file name for opening a second time on Windows [as the std lib docs indicate](https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile)

There's an [open bug report in for CPython](https://bugs.python.org/issue14243) for this portability issue; looks like it is targeted for Python 3.7 maybe. Basically, you can open the tempfile a second time in the context manager on unix but not on Windows.

@pv my suggestion is that we use `NamedTemporaryFile(mode='w', delete=False)` so that the tempfile is not actually deleted after the context manager control flow exits, and then we move the `conda env create` call outside of the context manager. This is far less Pythonic, and kind of defeats the purpose of using a tempfile, but it does circumvent the issue, and eventually we can use the more Pythonic approach when this gets patched in CPython and platform portability is restored.

tylerjereddy:
We may also have to switch back to using `--force` with conda as well, because we won't be able to leverage the cleanup of the tempfile if we use delete=False.

tylerjereddy:
I've pushed the attempted windows compatibility changes despite the issues discussed above -- let's see what happens with the CI now.

tylerjereddy:
So, less issues now:
- All Windows builds have only two failures, both for the same unit test because of `conda list -p env._path --json` failing on that platform for whatever reason (file path handling needs to be done more carefully in that unit test maybe?)
- only the Python 2.6 Travis test suite has a single failure; all others pass -- looks like an issue with pip wheel in Python 2.6

tylerjereddy:
`util.check_output([conda, 'list'])` is sufficient to trigger the missing file failure on Windows. Now testing to see what path for `conda` is being used on Windows, but whatever the path, it is already clear that `os.path.normpath()` is not sufficient to make it work on that platform.

tylerjereddy:
So, `env._path` exists as expected on the Windows / AppVeyor builds, but the `bin` subdirectory does not. I'm not sure if a symlink to the `conda` executable is being created within (another subdir of) `env._path` on Windows, but I'm pretty sure that the unix builds do create a symlink to conda executable somewhere (I remember seeing it when debugging).

There is a `conda-meta` entry in `env._path` on windows. I can take a peek in there.

My current suspicion is that `env.find_executable()` may need to be modified to point to the proper Windows executable for `conda`, because what it gives me now doesn't exist. There is a comment in the latter function referencing a standard conda layout, which is perhaps now no longer the case with my modifications, but it could also be something more confusing.

tylerjereddy:
My debug work indeed indicates that there's a `conda` file (probably a symlink type thing) in the `Scripts` dir on Windows / AppVeyor when building an env. `asv` assumes that the executable for conda on windows will have an `exe` extension in the env and reside in either the env root or the `Scripts` subdirectory, based on the code in `find_executable`. Since this is not the case in the current AppVeyor builds, `asv` falls back to assuming that we have a unix configuration with a `bin` path instead, but it is in fact a Windows build that appears to have a `conda` executable without an `exe` extension.

So, I'll aim to try to patch `find_executable` to handle this scenario. I'm not clear on whether this is genuinely related to anything I've changed rather than simply the current tests don't probe this issue, but I think the logical fix is becoming clearer now.

pv:
Maybe related: https://github.com/spacetelescope/asv/pull/525

pv:
Maybe the minimal thing would be to add a `paths=None` argument that 
takes a list of paths [with default value 
os.environ['PATH'].split(os.pathsep)] to util.which so that the thing is 
done only in a single place.

tylerjereddy:
So you're suggesting it would be more appropriate to use `util.which` than `find_executable` in my unit test to probe for conda, and to modify util.which for increased flexibility if needed? `find_executable` doesn't seem to directly interact with `util.which`, so I'm guessing that changing the latter won't affect the current test.

pv:
find_executable should be implemented using util.which(executable, 
paths=[...]), so that the logic of appending exe, bat, etc. on Windows 
is not duplicated in many places.

pv:
Note that it may be faster to grab a windows VM and debug locally, than triaging it via appveyor:
https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/
https://developer.microsoft.com/en-us/windows/downloads/virtual-machines

tylerjereddy:
I noticed that on a Windows machine `test_matrix_environments()` can be forced to use the `python.exe` within `self._path` with some minor changes. However, even with this adjustment, the libraries installed match the system rather than the environment if probed using the above `python.exe`. Further evidence that the system (or root / other conda) python is effectively being used by this test include variabilities in the library causing the version check failure when comparing AppVeyor test builds with builds on my Windows 10 machine / powershell.

That said, there is evidence that the original `conda env create` was done correctly because I can manually go into the pertinent `tmp` path on Windows and probe the expected library versions using `conda list -p (self._path)`.

What does this mean? Should there be a separate `python.exe` in the `Scripts` subfolder on windows? I'm only seeing one in the env `self._path`, and while the *version* of Python that is used does match the matrix, the version of the conda-installed libraries does not reflect the dependency matrix of the env (and probing the paths of these python libraries also doesn't match the env).

I think I'm slightly confused by the raw `python -c` calls in this unit test. I'm assuming we don't need to actually 'activate env' via conda  (there is no such call in the source anywhere) on the basis that the Python executable that is found is the one that would normally be 'activated' in any case, but I must be missing a little something here. This unit test is also rather prone to accidental passing since the system install of python could just happen to have dependency version matches to the env.

tylerjereddy:
- all the tests are passing on all platforms except for a single fail on Python 2.6 -- that failure I don't really understand, but it is an unsupported version of Python anyway
- there are maybe one or two debug-related lines of code that should be removed, but the rest should be looking pretty solid now
- we could maybe add the unit tests from the issue @pv linked above and ensure that we handle those file types as well, although that could also be considered a separate issue which is only partly addressed here to obtain the desired functionality, which was really only supposed to be custom conda channel support

asv/util.py
```diff
@@ -294,9 +292,17 @@ def which(filename):
 
     candidates = []
     for location in locations:
+        for alternate_ext in ['.exe', '.bat']:
```
pv: This must be contained inside `if WIN` --- the stuff does not exist on other platforms.

asv/environment.py
```diff
@@ -416,7 +416,9 @@ def check_presence(self):
 
         for executable in ['pip', 'python']:
             exe = self.find_executable(executable)
-            if not os.path.isfile(exe):
+            # do not count a system install of pip or python
+            # as evidence of an installed env
+            if not os.path.isfile(exe) or (self._path not in os.path.dirname(exe)):
```
pv: This check should be removed: the correct paths must be specified in find_executable.

asv/util.py
```diff
@@ -272,20 +272,18 @@ def human_value(value, unit, err=None):
...
-    if WIN:
-        if not filename.endswith('.exe'):
-            filename = filename + '.exe'
-
     if os.path.sep in filename:
         locations = ['']
     else:
         locations = os.environ.get("PATH", "").split(os.pathsep)
+        if paths is not None:
+            locations = paths + locations
```
pv: This should be:
```
if paths is not None:
    locations = paths
else:
    locations = os.environ ...
```

asv/plugins/conda.py
```diff
@@ -85,30 +86,44 @@ def _matches(cls, python):
...
-            '--use-index-cache',
-            'python={0}'.format(self._python),
-            'pip'])
-
-        log.info("Installing requirements for {0}".format(self.name))
-        self._install_requirements(conda)
-
-    def _install_requirements(self, conda):
-        self.install('wheel')
+        # create a temporary environment.yml file
```
pv: Empty line before comment block, not after it.

asv/plugins/conda.py
```diff
@@ -85,30 +86,44 @@ def _matches(cls, python):
...
+            env_file.write('name: {0}\n'
+                           'channels:\n'.format(self.name))
+            env_file.writelines(('   - %s\n' % ch for ch in self._conda_channels))
+            env_file.write('dependencies:\n'
+                           '   - python={0}\n'
+                           '   - wheel\n'
+                           '   - pip\n'.format(self._python))
+            # categorize & write dependencies based on pip vs. conda
+            reqs = self._get_requirements(conda)
+            # handle the case where there are multiple dependencies
```
pv: Empty line before comments block when at the same intendation as the preceding line.

asv/environment.py
```diff
@@ -416,7 +416,9 @@ def check_presence(self):
 
         for executable in ['pip', 'python']:
             exe = self.find_executable(executable)
-            if not os.path.isfile(exe):
+            # do not count a system install of pip or python
```
pv: Empty line before comment blocks, when at the same intendation level as the preceding line.

asv/plugins/conda.py
```diff
@@ -85,30 +86,44 @@ def _matches(cls, python):
 
     def _setup(self):
         try:
-            conda = util.which('conda')
+            conda = self.find_executable('conda')
```
pv: This must use `util.which`.
The environment is not initialized yet at this stage, so it should not try looking for the conda executable inside it.

asv/util.py
```diff
@@ -272,20 +272,18 @@ def human_value(value, unit, err=None):
...
 
         if WIN:
             # On windows, an entry in %PATH% may be quoted
```
pv: Also, this part can be run only when `paths is not None` because the `paths` list does not contain quoted paths.

asv/util.py
pv: Also, this is simpler to rewrite so that you build the list of possible filenames in a separate `if WIN: ... else: ...` block above: 
```
if WIN:
    ...
    filenames = [filename + ext for ext in ('.exe', '.bat', '')]
else:
    ...
    filenames = [filename]
...
for location in locations:
    for fn in filenames:
        candidate = os.path.join(location, fn)
        ...

**Comment by pv** [comment]
ok, fixed those myself now that I'm at it.

**Comment by pv** [comment]
Still needs:

- adding documentation to `docs/source/asv.conf.json.rst`
- adding the option to `asv/template/asv.conf.json`

**Comment by tylerjereddy** [comment]
Thanks for taking care of those! I pushed first draft documentation changes for the files you requested.

**Comment by pv** [comment]
Rebased and squashed, will still have to test on windows (separately, since appveyor doesn't seem to be on a cooperative mood)

**Comment by tylerjereddy** [comment]
The AppVeyor tests all pass now too. I note that on my Windows 10 machine I still get the 1 unit test failure I describe above involving `test_matrix_environments`.

I think my basic conclusion was that the conda env site package libs weren't always being used by the python executable in the env created by this test even though conda list -p self._path listed the correct dependency versions for the env.

Maybe it is ok since AppVeyor isn't reproducing this. For completeness, pytest traceback:

```
================================== FAILURES ===================================
__________________________ test_matrix_environments ___________________________

tmpdir = local('c:\\users\\treddy\\appdata\\local\\temp\\pytest-of-treddy\\pytest-45\\test_matrix_environments0')

    @pytest.mark.skipif(not ((HAS_PYTHON_27 and HAS_PYTHON_34) or HAS_CONDA),
                        reason="Requires Python 2.7 and 3.4")
    def test_matrix_environments(tmpdir):
        conf = config.Config()

        conf.env_dir = six.text_type(tmpdir.join("env"))

        conf.pythons = ["2.7", "3.4"]
        conf.matrix = {
            "six": ["1.10", None],
            "colorama": ["0.3.6", "0.3.7"]
        }
        environments = list(environment.get_environments(conf, None))

        assert len(environments) == 2 * 2 * 2

        # Only test the first two environments, since this is so time
        # consuming
        for env in environments[:2]:
            env.create()

            output = env.run(
                ['-c', 'import six, sys; sys.stdout.write(six.__version__)'],
                valid_return_codes=None)
            if 'six' in env._requirements:
>               assert output.startswith(six.text_type(env._requirements['six']))
E               AssertionError: assert False
E                +  where False = <built-in method startswith of unicode object at 0x000000000419F060>('1.10')
E                +    where <built-in method startswith of unicode object at 0x000000000419F060> = '1.11.0'.startswith
E                +    and   '1.10' = <type 'unicode'>('1.10')
E                +      where <type 'unicode'> = six.text_type

```

If I manually check:
```
PS C:\users\treddy\appdata\local\temp\pytest-of-treddy\pytest-45\test_matrix_environments0\env\d26ca092d5e1c69bb963927c9c813a5e> .\python.exe
Python 2.7.13 |Continuum Analytics, Inc.| (default, May 11 2017, 13:17:26) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> import six
>>> six.__path__
[]
>>> import colorama
>>> colorama.__path__
['C:\\users\\treddy\\appdata\\local\\temp\\pytest-of-treddy\\pytest-45\\test_matrix_environments0\\env\\d26ca092d5e1c69bb963927c9c813a5e\\lib\\site-packages\\colorama']
>>> six.__version__
'1.11.0'
>>> colorama.__version__
'0.3.6'
>>> help(six)

Help on package six:

NAME
    six - Utilities for writing code that runs on Python 2 and 3

FILE
    c:\users\treddy\appdata\roaming\python\python27\site-packages\six-1.11.0-py2.7.egg\six.py
```

Looks like the env version of colorama is used, but not the case for six. The env packages are as expected though (including the version of `six` the unit test is looking for but doesn't find):
```
PS C:\users\treddy\appdata\local\temp\pytest-of-treddy\pytest-45\test_matrix_environments0\env\d26ca092d5e1c69bb963927c9c813a5e> conda list -p .
# packages in environment at C:\users\treddy\appdata\local\temp\pytest-of-treddy\pytest-45\test_matrix_environments0\env\d26ca092d5e1c69bb963927c9c813a5e:
#
certifi                   2016.2.28                py27_0
colorama                  0.3.6                    py27_0
pip                       9.0.1                    py27_1
python                    2.7.13                        1
setuptools                36.4.0                   py27_1
six                       1.10.0                   py27_0
vs2008_runtime            9.00.30729.5054               0
wheel                     0.29.0                   py27_0
wincertstore              0.2                      py27_0
```

Is Windows 10 newer than what is used in the current AppVeyor matrix? I suspect the actual issue is my usage of powershell instead of CMD --  I'm using a solution desribed in [an open conda issue](https://github.com/conda/conda/issues/626) for environment handling in this currently problematic scenario (powershell + Windows 10 + conda). So, perhaps we can consider that a conda issue or not critical.

In short, it seems that official conda versions don't support environment activation / control in powershell + Windows 10 at the moment, but there are some work-arounds described above. I'm not sure if that is really the source of the issue (I suppose if I test on CMD instead of powershell and everything passes that would be enlightening).

**Comment by pv** [comment]
The six on your machine looks like it comes from a `--user` installed 
package.

**Comment by pv** [comment]
This actually looks like a conda bug (would maybe be useful to report to 
them):
```
$ conda create -n foo python=3.6
...
$ source activate foo
$ python
Python 3.6.2 |Continuum Analytics, Inc.| (default, Jul 20 2017, 13:51:32)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
 >>> import sys
 >>> sys.path
['', '/home/pauli/.local/lib/conda/envs/foo/lib/python36.zip', 
'/home/pauli/.local/lib/conda/envs/foo/lib/python3.6', 
'/home/pauli/.local/lib/conda/envs/foo/lib/python3.6/lib-dynload', 
'/home/pauli/.local/lib/python3.6/site-packages', 
'/home/pauli/.local/lib/conda/envs/foo/lib/python3.6/site-packages']
#
# Note that /home/pauli/.local/lib/python3.6/site-packages
# appears in sys.path. This seems to only occur on Python 3.6.
#
```

tylerjereddy:
Ok, I've opened the issue on the conda github (cross-ref'd above) after confirming on my machine that this doesn't happen with Python 2.7 or 3.5, so does indeed appear to be 3.6 only as you suggest. I offered to help them patch it if they want to guide me.

I didn't observe any other issues running the full `asv` testsuite on Win 10 / powershell.

pv:
Thanks, merged!

tylerjereddy:
Note that I can reproduce the issue on Python 2.7 if I install with --local packages before creating an env there as well. Anyway, I'll carry on that discussion with the conda team I guess.

tylerjereddy:
Ok, I've closed the conda issue -- note that the usage of `export PYTHONNOUSERSITE=True` solves our problem and the default behavior of Python is to include the site packages in the manner we have both observed (see discussion in the conda issue & links therein).

There are probably other ways to exclude the site packages as well so that conda behaves more like virtualenv, which is probably what we want in asv. Without such a setting / suppression of local site packages, there is some risk that running the conda components of the asv testsuite locally will produce different results than in containerized CI tests where there are no local site packages that may pollute the path.

Is this sufficiently worrying to open another issue for asv so that we have robust site package suppression with conda (i.e., matching isolation of the envs for conda and virtualenv)?

SylvainCorlay:
It does not seem that other conda channels than `default`  are taken into account with the version at the tip of master at the moment.

pv:
The tests show otherwise.

pv:
That is, it may be then a difference between the environment at hand,
and what the tests use.

SylvainCorlay:
The tests appear to be setting the `conda_channels` from the Python API (and not getting it from the `conf.json` file).

SylvainCorlay:
Opened #657 to track this specific issue.

pv:
That should not matter. Indeed, it works correctly also in practice for
me when e.g. setting "conda_channels": ["conda-forge"] in the config
file.

Perhaps you have old environments left over --- it doesn't change the
conda_channels automatically?

PR #540, opened 2017-08-07
Add the possibility for asv machine to use defaults

Lothiraldan:
Using asv in a full-automated way is difficult because asv run needs asv
machine to run first. Make asv machine accept a '--yes' that accept the
default value. This way automated scripts can put 'asv machine --yes' before
running 'asv run'.

### DIFF
--- a/asv/console.py
+++ b/asv/console.py
@@ -230,8 +230,12 @@ def color_print(*args, **kwargs):
         write(end)
 
 
-def get_answer_default(prompt, default):
+def get_answer_default(prompt, default, use_defaults=False):
     color_print("{0} [{1}]: ".format(prompt, default), end='')
+
+    if use_defaults:
+        return default
+
     x = input()
     if x.strip() == '':
         return default
--- a/asv/machine.py
+++ b/asv/machine.py
@@ -147,8 +147,8 @@ class Machine(object):
             }
 
     @staticmethod
-    def generate_machine_file():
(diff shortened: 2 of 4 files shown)


asv/console.py
```diff
@@ -230,8 +230,12 @@ def color_print(*args, **kwargs):
         write(end)
 
 
-def get_answer_default(prompt, default):
+def get_answer_default(prompt, default, accept_default=False):
     color_print("{0} [{1}]: ".format(prompt, default), end='')
+
+    if accept_default is True:
```
pv: `if accept_default` according to pep8

asv/machine.py
```diff
@@ -147,8 +147,8 @@ def get_defaults():
             }
 
     @staticmethod
-    def generate_machine_file():
-        if not sys.stdout.isatty():
+    def generate_machine_file(accept_default=False):
+        if not sys.stdout.isatty() and accept_default is False:
```
pv: `not accept_default`

pv:
lgtm, aside from style nitpicks

pv:
Also, adding a test in `test/test_machine.py` for this would be useful.
You can use `run_asv` from `test/tools.py` to run asv in a test.

Lothiraldan:
@pv Sorry for the delay, I added a unit-test at the Machine object level. Do you think it's sufficient ?

pv:
thanks, merged

PR #545, opened 2017-09-21
Allow to overwrite 'html_dir' and to filter commits considered in 'publish'

philpep:
Our usecase is to have multiple website giving multiple views of our workflow development. Eg. One website for published commits and one for WIP, also having a limited website that only show tagged commits.


pv:
thanks, lgtm

PR #555, opened 2017-09-26
Use a less hacky way to launch virtualenv

pv:


asv/plugins/virtualenv.py
```diff
@@ -58,11 +58,6 @@ def __init__(self, conf, python, requirements):
             raise environment.EnvironmentUnavailable(
                 "virtualenv package not installed")
 
-        # Can't use `virtualenv.__file__` here, because that will refer to a
-        # .pyc file which can't be used on another version of Python
-        self._virtualenv_path = os.path.abspath(
-            inspect.getsourcefile(virtualenv))
-
```
philpep: Seems that `import inspect` is now unused in this file.

LGTM otherwise.

PR #556, opened 2017-10-04
Display last part of benchmark name in left pannel

philpep:
On left pannel, navigation accross benchmark is generated using
benchmark name partitioned by '.', only display last part of the name
since all other parts are already displayed.

Previous code using "bm.name || last_part" was inconsistent since
bm.name is never empty and last_part is generated using bm.name

### DIFF
--- a/asv/www/graphdisplay.js
+++ b/asv/www/graphdisplay.js
@@ -185,7 +185,7 @@ $(document).ready(function() {
                 });
             }
 
-            var name = bm.pretty_name || bm.name || parts[parts.length - 1];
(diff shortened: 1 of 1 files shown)


pv:
Thanks. There's little space in the panel anyway, so this is better.

pv:
I think this doesn't actually work because `self.pretty_name = getattr(func, "pretty_name", name)`

Should have checked better before merging.

philpep:
Yes, indeed. I was mislead by my test suite which set pretty_name to None by default. I'll try to make this more consistent. Good catch, thanks!

PR #557, opened 2017-10-04
Group benchmarks in summary grid

philpep:
Use the first component of the benchmark name as group name and use it
as separation between benchmarks. So we can remove group name from
benchmark name and it become shorted.
This ease quickly reading the summary grid.

Screenshot https://framapic.org/Bo0MeTRUrTWX/RSzZevQISzuu.png


asv/www/summarygrid.js
```diff
@@ -18,6 +18,89 @@ $(document).ready(function() {
         $(window).on('scroll', handler);
     }
 
+    function get_benchmarks_by_groups() {
```
pv: Moved this back to summarygrid.js since it's not used by anything else.

Issue #561, opened 2017-10-19
Would like a way to run commits only on merge commits


moorepants:
Hi, we use asv for sympy and we do not have strict rules about combining all commits in a PR into one before merging, so SymPy has lots of commits that don't' really need to be benchmarked. It would be nice if there was an easy way to just run the benchmarks on the merge commits so we could essentially benchmark the result of each pull request. Is this something that is easy to do? If not, it'd be a nice feature to have.

pv:
A work-around is to list commits one by one, along the lines of
"for c in `git log --merges --oneline --format=%H`; do asv run "$c^!";
done"

moorepants:
Cool, that works. It probably wouldn't be too hard to make this work:

```
git log --merges --oneline --format=%H | asv run
```

pv:
`asv run "--merges master"` works

PR #562, opened 2017-10-19
Enabled stdin for run range_spec.

moorepants:
This assumes that you supply a chunk of text in which a commit hash is
on each line of the text. For example, this now works:

git log --oneline --format=%H | asv run

Fixes issue 561.

### DIFF
--- a/asv/commands/run.py
+++ b/asv/commands/run.py
@@ -6,6 +6,7 @@ from __future__ import (absolute_import, division, print_function,
 
 import six
 
+import sys
 import logging
 import traceback
 
@@ -53,7 +54,7 @@ class Run(Command):
             description="Run a benchmark suite.")
 
         parser.add_argument(
-            'range', nargs='?', default=None,
+            'range', nargs='?', default=sys.stdin,
             help="""Range of commits to benchmark.  For a git
             repository, this is passed as the first argument to ``git
             log``.  See 'specifying ranges' section of the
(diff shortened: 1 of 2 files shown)


pv:
I don't like that much reading stuff from stdin for this command. How about instead enabling it to take multiple arguments on command line, so you could run `| xargs asv run` instead?

pv:
OTOH, the usual unix "read from stdin" could be to do `asv run -` although not sure if argparse likes that.

pv:
Also, the final concern is that `asv run` may ask interactive questions when the machine file is missing --- does that still work ok in this scenario?

moorepants:
I'm happy to change this to whatever you all prefer, stdin or xargs approach. I like the former because I never remember xargs. The interactivity works:

```
(asv-dev) moorepants@garuda:sympy_benchmarks(issue4)$ git --git-dir /home/moorepants/src/sympy/.git log --merges --oneline --format=%H | asv run
· Config file /home/moorepants/src/sympy_benchmarks/asv.conf.json not found.
```

pv:
I'm fine with `asv run -`, although it could be nice to also allow multiple arguments.

basnijholt:
Let me just mention https://github.com/airspeed-velocity/asv/issues/561 such that the issue and PR are linked.

I would like to  have this feature too,
`"for c in `git log --merges --oneline --format=%H`; do asv run "$c^!";
done"`
is just a bit unreadable.

PR #568, opened 2017-10-29
Perform benchmark discovery only when running benchmarks

pv:
Discover benchmarks only when the project is going to be built and the
commit hashes to be checked are known, so the benchmark discovery can be
performed using those commits.  This also adjusts the Benchmarks API so
that the discovery is explicit.

Previously, benchmark discovery was done installing a version of the
project at the tip of the branch or from an existing wheel. This leads
to problems when the build fails or the project is not in an importable
state.

This also changes "asv publish" to emit an error message rather than
running benchmark discovery, if benchmarks.json is damaged or missing.
That is an unusual error condition, so it is reasonable to handle it as
such (the error message explains how to resolve it).

### DIFF
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -309,7 +309,40 @@ class Benchmarks(dict):
     """
     api_version = 1
 
-    def __init__(self, conf, repo, environments, benchmarks=None, regex=None):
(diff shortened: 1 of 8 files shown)


pv:
I think this has been dogfooded for long enough now.

PR #569, opened 2017-10-29
virtualenv: run pip via python -mpip to avoid shebang limits

pv:
The script files in a virtualenv are not necessarily executable if the
path is too deeply nested so that shebang limit overflows.


pv:
Merging (fixes travis failures).

PR #580, opened 2017-11-14
Generate shareable urls

philpep:
This is a first attempt to generate shareable urls (i.e. copying url and pasting to a new tab should display the same page in the same state).

TODO (might be in separate PR):

- [x] What about possible conflicting param names (commits, branch, idx, x-axis, y-axis) ? EDIT: fixed by prefixing params with `p-`
- [x] Replace "idx" (index of params permutation) in regression / summarylist by flat params
- [ ] store actual zoom state (EDIT: might be done in dedicated PR)

### DIFF
--- a/asv/www/graphdisplay.js
+++ b/asv/www/graphdisplay.js
@@ -385,9 +385,19 @@ $(document).ready(function() {
                 var max_curves = 8;
 
                 for (var k = 0; k < params.length; ++k) {
+                    var param_name = param_names[k]
+                    var param_values = params[k]
                     var item = [];
-                    for (var j = 0; j < params[k].length && (j+1)*count <= max_curves; ++j) {
-                        item.push(j);
+                    if (state_selection['p-'+param_name] !== undefined) {
+                        for (var j = 0; j < param_values.length; ++j) {
+                            if (state_selection['p-'+param_name].includes(param_values[j])) {
+                                item.push(j);
+                            }
+                        }
+                    } else {
+                        for (var j = 0; j < param_values.length && (j+1)*count <= max_curves; ++j) {
+                            item.push(j);
+                        }
                     }
                     count = count * item.length;
                     benchmark_param_selection.push(item);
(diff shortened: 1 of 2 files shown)


philpep:
I fixed the two ongoing issues:

- Fix possible param name conflict by prefixing benchmark params with `p-`
- Replace the `idx` by flat param selection for generating regressions graph links.

I think the PR is ready for review & merge. Thansk!

pv:
The links generated from "Regressions" page don't seem to work for me.

philpep:
Thanks for testing this. I was able to reproduce a similar error message with a small screen height, do you still have the issue with #596 ?

pv:
That fixed the JS exception, but there's still no plot shown.

asv/www/graphdisplay.js
```diff
@@ -385,9 +385,19 @@ $(document).ready(function() {
...
 
                 for (var k = 0; k < params.length; ++k) {
+                    var param_name = param_names[k]
+                    var param_values = params[k]
                     var item = [];
-                    for (var j = 0; j < params[k].length && (j+1)*count <= max_curves; ++j) {
-                        item.push(j);
+                    if (state_selection['p-'+param_name] !== undefined) {
+                        for (var j = 0; j < param_values.length; ++j) {
+                            if (state_selection['p-'+param_name].includes($.asv.convert_benchmark_param_value(param_values[j]))) {
```
pv: `$.asv.convert_benchmark_param_value` does not always return a string, so this construct fails. The function probably should be changed to always return a string.

pv: Maybe the parameters should not use `$.asv.convert_benchmark_param_value` at all, and instead just use the raw string value.

philpep: Good catch, this was broken with non string parameters. Drop the use `$.asv.convert_benchmark_param_value` fix it.
I pushed new version of the fist patch, tested against numpy and mercurial suite.

pv:
I now notice the Atom feed URL generation also needs updates to match 
the new URL scheme.

The `idx` parameter probably should be removed from the location hash if 
present, because its presence interferes with the parameter selector UI.

PR #590, opened 2017-12-07
flake8 fixups

jbrockmendel:
As long as I'm going over the code to get a feel for it, might as well...

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,3 +3,9 @@ python-file-with-version = asv/__init__.py
 
 [tool:pytest]
 addopts=-p no:logging
+
+[flake8]
+exclude=extern/
+max-line-length = 80
+ignore=E402
+# E402 module level import not at top of file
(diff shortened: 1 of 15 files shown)


philpep:
Hello, I think it could be great if flake8 is enforced by a new travis job. ALso please note that there is already a pyflakes check in travis, maybe flake8 could replace it.

setup.cfg
```diff
@@ -3,3 +3,9 @@ python-file-with-version = asv/__init__.py
 
 [tool:pytest]
 addopts=-p no:logging
+
+[flake8]
+exclude=extern/
+max-line-length = 80
```
pv: This could be bigger, maybe 120.

jbrockmendel: It's not up to me, but the default is 79 and that's pretty standard.  I only bumped it to 80 here because there are a lot of lines in asv that are exactly 80.

pv: I recommend you change it to a larger number. 80 may be a limit, but there's no particular reasoning for that particular number. Indeed, asv codebase at several places has longer lines, which I don't think is problematic for readability.

PR #593, opened 2017-12-08
Add possibility to filter param selection to run

philpep:
By extending --bench command to match against benchmark_name('param1', 'param2', ...)

Skipped param combination are displayed as "skipped" in asv run output
and results are merged with eventual previous results (skipped results
keep their old previous value).

This allow for instance to run only newly added parameters set.

### DIFF
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -33,7 +33,7 @@ BENCHMARK_RUN_SCRIPT = os.path.join(
 
 
 def run_benchmark(benchmark, root, env, show_stderr=False,
-                  quick=False, profile=False, cwd=None):
+                  quick=False, profile=False, cwd=None, selected_idx=None):
     """
     Run a benchmark in different process in the given environment.
 
@@ -59,6 +59,10 @@ def run_benchmark(benchmark, root, env, show_stderr=False,
     cwd : str, optional
         The path to the current working directory to use when running
         the benchmark process.
+
+    selected_idx : str, optional
+        The list of parameters combination to run benchmark on. By default
+        run all combinations.
 
     Returns
     -------
@@ -120,6 +124,13 @@ def run_benchmark(benchmark, root, env, show_stderr=False,
         result['started_at'] = datetime.datetime.utcnow()
 
         for param_idx, params in param_iter:
+            if (selected_idx is not None and benchmark['params']
+                    and param_idx not in selected_idx):
+                # Use NaN to mark the result as skipped
+                bench_results.append(dict(samples=None, number=None,
+                                          result=float('nan'), stats=None))
+                bench_profiles.append(None)
+                continue
             success, data, profile_data, err, out, errcode = \
                 _run_benchmark_single(
                     benchmark, root, env, param_idx,
@@ -326,6 +337,9 @@ class Benchmarks(dict):
             `regex` is a list of regular expressions matching the
             benchmarks to run.  If none are provided, all benchmarks
             are run.
+            For parameterized benchmarks, the regex match against
+            `funcname(param0, param1, ...)` to include the parameter
+            combination in regex filtering.
         """
         self._conf = conf
         self._benchmark_dir = conf.benchmark_dir
@@ -336,10 +350,23 @@ class Benchmarks(dict):
             regex = [regex]
 
         self._all_benchmarks = {}
+        self._benchmark_selection = {}
         for benchmark in benchmarks:
             self._all_benchmarks[benchmark['name']] = benchmark
-            if not regex or any(re.search(reg, benchmark['name']) for reg in regex):
-                self[benchmark['name']] = benchmark
+            if benchmark['params']:
+                self._benchmark_selection[benchmark['name']] = []
+                for idx, param_set in enumerate(
+                        itertools.product(*benchmark['params'])):
+                    name = '%s(%s)' % (
+                        benchmark['name'],
+                        ', '.join(param_set))
+                    if not regex or any(re.search(reg, name) for reg in regex):
+                        self[benchmark['name']] = benchmark
+                        self._benchmark_selection[benchmark['name']].append(idx)
+            else:
+                self._benchmark_selection[benchmark['name']] = None
+                if not regex or any(re.search(reg, benchmark['name']) for reg in regex):
+                    self[benchmark['name']] = benchmark
 
     @classmethod
     def discover(cls, conf, repo, environments, commit_hash, regex=None):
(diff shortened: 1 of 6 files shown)


pv:
Appveyor failures may be real

pv:
NB. you may save some time by debugging in a real Windows environment,
e.g. in VM:
https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

pv:
Ah right. Maybe repo.py should set
os.environ['GIT_CEILING_DIRECTORIES'] += dirname(repo), 
or maybe use --git-dir instead of -C,
so that it won't try to climb up?

pv:
iiuc, hglib doesn't have this problem?

philpep:
We probably have the issue with hglib too (but since asv is using git there's no bug in appveyor tests).

I think a good fix is to test for `.git` or `.hg` directories prior running commands.

philpep:
Well maybe not, since `git` will also look for parent directory even if `.git` exist (but is empty). I'll try with `--git-dir` instead.

pv:
--git-dir also needs --work-tree used at the same time, not sure about
bare repos

philpep:
Yes `git-dir` doesn't work in the same way when using bare repos. Seems that GIT_CEILING_DIRECTORIES is the simplest way to ensure a git root repository.

Made it in https://github.com/airspeed-velocity/asv/pull/636

asv/benchmarks.py
```diff
@@ -120,6 +120,10 @@ def log_result(msg):
         result['started_at'] = datetime.datetime.utcnow()
 
         for param_idx, params in param_iter:
+            if benchmark['params'] and param_idx not in benchmark['_selected_idx']:
+                bench_results.append(dict(samples=None, number=None, result='skipped', stats=None))
```
pv: `result=nan` is used for skipped values elsewhere, and can be used also here so that it's not necessary postprocess it out afterward or change printing.

asv/benchmarks.py
```diff
@@ -145,7 +149,7 @@ def log_result(msg):
                     bench_profiles.append(profile_data)
             else:
                 failure_count += 1
-                bench_results.append(dict(samples=None, number=None, result=None, stats=None))
+                bench_results.append(dict(samples=None, number=None, result='failed', stats=None))
```
pv: This can be left as is.

asv/benchmarks.py
```diff
@@ -230,6 +234,8 @@ def log_result(msg):
             with log.indent():
                 log.error(result['stderr'])
 
+        result['result'] = [None if x in ('skipped', 'failed') else x
+                            for x in result['result']]
```
pv: This does not need to be added.

asv/commands/profiling.py
```diff
@@ -166,7 +166,7 @@ def run(cls, conf, benchmark, revision=None, gui=None, output=None,
 
             benchmarks = Benchmarks.discover(conf, repo, environments,
                                              [commit_hash],
-                                             regex='^{0}$'.format(benchmark))
+                                             regex='^{0}\('.format(benchmark))
```
pv: But does it now match to non-parametrized benchmarks any more?

asv/results.py
```diff
@@ -559,19 +559,41 @@ def load(cls, path, machine_name=None):
 
         return obj
 
+    def _merge_results(self, old):
```
pv: Move this method below `add_existing_results` so it's easier to read?

asv/util.py
```diff
@@ -266,7 +266,9 @@ def human_value(value, unit, err=None):
             if err is not None:
                 display += "±{:.2g}".format(err)
     elif value is None:
-        display = "failed"
+        display = 'failed'
+    elif value in ('failed', 'skipped'):
```
pv: Doesn't need to be added.

asv/results.py
```diff
@@ -559,19 +559,41 @@ def load(cls, path, machine_name=None):
...
+                new_results[key] = val
+            elif self._benchmark_params[key]:
+                old_benchmark_results = {}
+                for idx, param_set in enumerate(itertools.product(
+                        *old._benchmark_params[key])):
+                    old_benchmark_results[param_set] = val[idx]
+                for idx, param_set in enumerate(itertools.product(
+                        *self._benchmark_params[key])):
+                    if (new_results[key][idx] is None and
+                            old_benchmark_results.get(param_set) is not None):
```
pv: I think this logic should use the mask stored in `self._benchmark_selection`, not look at the result values.

Benchmarks may be skipped also due to the benchmark suite itself, which is a somewhat different situation than here.

asv/benchmarks.py
```diff
@@ -338,8 +344,15 @@ def __init__(self, conf, benchmarks, regex=None):
...
-                self[benchmark['name']] = benchmark
+            benchmark['_selected_idx'] = []
+            for idx, param_set in enumerate(
+                    itertools.product(*benchmark['params'])):
+                name = '%s(%s)' % (
+                    benchmark['name'],
+                    ', '.join(param_set))
+                if not regex or any(re.search(reg, name) for reg in regex):
+                    self[benchmark['name']] = benchmark
+                    benchmark['_selected_idx'].append(idx)
```
pv: Storing it in `benchmark` I think makes it end up in the json files when saving which is a bit ugly.

Maybe 
`self._benchmark_selection[benchmark['name']] = []`
would be nicer?

asv/benchmarks.py
```diff
@@ -120,6 +120,10 @@ def log_result(msg):
         result['started_at'] = datetime.datetime.utcnow()
 
         for param_idx, params in param_iter:
+            if benchmark['params'] and param_idx not in benchmark['_selected_idx']:
```
pv: I'd maybe prefer a new `selected_idx` keyword argument to `run_benchmark` instead of storing the selection mask in the benchmark dicts.

Then it's just a matter of adding `selected_idx=self._benchmark_selection[name]` here
https://github.com/philpep/asv/blob/2df016ddbd6236c1e6ddbfd94c366198702f2b80/asv/benchmarks.py#L626-L630

asv/results.py
pv: Ok, sorry, that information is not available here.

Then I guess the next best is to replace this if statement with `if util.is_nan(new_results[key][idx]):`, which will fill in skipped results but leave failures in.

philpep:
Thank your for reviewing this! I'll submit a V2 to address your comments.

asv/commands/profiling.py
philpep: With current implementation it will match because we regex search against `benchmark_name()`.
But I think we should keep matching against `benchmark_name` for non-parameterized benchmarks (at least for backward compatibility).
For parameterized benchmark `asv profile` will require a fully qualified benchmark name with param selection. I'll update the `--help` command to reflect this.

philpep:
I updated the patch. Thanks again for your effective review, the code looks a lot better than before!

Summary of changes:

* use NaN for skipped results instead 'skipped'/'failed' hack
* use Benchmark._benchmark_selection to store the idx selection for each benchmark and pass it as `selected_idx` to `run_benchmarks()`
* move `_merge_results()` into `add_existing_results()`
* match regex against `benchmark_name` without trailing `()` for non-parameterized benchmarks

pv:
thanks, merged

PR #608, opened 2018-01-17
Add branch state in url on regression page

philpep:
Add a 'branch' parameter in the url so the branch selection can be
shared through the url.


asv/www/regressions.js
```diff
@@ -113,11 +122,19 @@ $(document).ready(function() {
             });
         });
 
+        var branch_index = 0;
         if (branches && branches.length > 1) {
-            current_title = "Regressions in " + branches[0] + " branch";
+            var info = $.asv.parse_hash_string(window.location.hash);
+            if (info.params.branch) {
```
pv: Probably no need to re-parse the location hash.
`params` is passed to this function (see line 45, 53)

pv: Probably no need to re-parse the location hash.
`params` is passed to this function (see line 45, 53)

philpep: Indeed, changed to use `params` directly.

asv/www/regressions.js
```diff
@@ -113,11 +122,18 @@ $(document).ready(function() {
             });
         });
 
+        var branch_index = 0;
         if (branches && branches.length > 1) {
-            current_title = "Regressions in " + branches[0] + " branch";
+            if (params.branch) {
+                branch_index = branches.indexOf(info.params.branch[0]);
```
pv: `info.params` -> `params`

PR #611, opened 2018-01-31
Allow specifying project subdir in the repo

### LINKED ISSUE #547: Cannot specify an install command (or no install)
I'm writing benchmarks for a package that's not its own git repo, but is located in a subdir of a larger repo (I know, not optimal - can't control that in this case). There seems to be no way to specify the location of ``setup.py`` or a custom install command in ``asv.conf.json``, it assumes that the ``setup.py`` (or ``pyproject.toml`` etc.) is located in the root of the repo.

A more limited alternative that would also be useful is to just run the benchmarks in-place, without trying to clone, build a wheel, etc. Just execute benchmarks as is, giving responsibility for installing the package to the developer. 

The most similar issue I could find is gh-26 (custom install command), but either a ``--no-install`` or a ``--setup-file-location`` option seems more limited in scope and easier to implement than gh-26.

pitrou:
Some Python projects are not located at the root of their repository. For example, the Arrow project has its Python bindings in a `python` subdirectory (see https://github.com/apache/arrow/).

This allows configuring ASV so that it is able to run "setup.py build" and "pip install" from the right directory.

This PR has been checked to work with `asv run` and `asv dev` at least.

Fixes #547

### DIFF
--- a/asv/config.py
+++ b/asv/config.py
@@ -22,6 +22,7 @@ class Config(object):
         self.project = "project"
         self.project_url = "#"
         self.repo = None
+        self.repo_subdir = ""
         self.branches = [None]
         self.pythons = ["{0[0]}.{0[1]}".format(sys.version_info)]
         self.matrix = {}
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -354,6 +354,7 @@ class Environment(object):
 
         """
         self._env_dir = conf.env_dir
+        self._repo_subdir = conf.repo_subdir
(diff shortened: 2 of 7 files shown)


pv:
I guess this is OK as a minimal solution. Fully customizable build setup could come later if needed. 

Test (c.f. eg `test_workflow:test_run_build_failure`)?

pitrou:
Ok, I added some tests and a piece of doc.

PR #625, opened 2018-02-20
Fix #624: Provide ASV-specific environment variables to subprocesses

pitrou:
This follows a general pattern in development and integration tools.

Issue #78, opened 2014-07-10
Support Autotools build


cdeil:
I'd like to use `asv` for [GammaLib](http://gammalib.sourceforge.net/), which is a C++ library with a SWIG Python wrapper.
The installation after git checkout is basically

```
./autogen.sh
./configure --prefix=...
make
make install
```

Then a setup script must be sourced and then it's possible to `import gammalib`.

How much work would it be to support such an install with `asv` and what needs to be done?
Is there any change that the memory measurements are meaningful if pretty much all of the objects are going to be SWIG-wrapped C++ objects?

PR #626, opened 2018-02-20
[EXP] Issue #78: Customize build and install commands

pitrou:
Proof-of-concept based on PR #625.

This is meant to trigger discussion. One slight issue with customizing build and install commands is that it has to disable the wheel cache (which seems disabled by default, though). Another possibility would be to have separate pre-build and pre-install hooks, but I think forcing "setup.py build" and "pip install" on everyone isn't very convenient.


pv:
I think the simpler the solution from asv's side, the better, so from that POV I think this approach looks good. (Sorry for the late reply...)

One thing however is to decide how this plays together with the "environment" stuff.
(i) Should it fully override the virtualenv/conda config, or just the package installation part?
(ii) If the latter, do the install commands need to be env-specific?

I would perhaps prefer (i) just package installation, and (ii) not env-specific. I think the user can always write a "shim build system" e.g. a shell script that does whatever is needed (based on environment variables etc. that asv sets).

If there is variation in the necessary build commands during version history, then I think the user can factor this in their shell script.

pitrou:
Here is a working example using this PR:
https://github.com/apache/arrow/pull/1775

As you can see, in the current form neither the conda config nor the package installation part is overriden, but the build script just builds on that.

pitrou:
I've merged from master and fixed conflicts.

pv:
Using this as a basis, #699 --- any comments welcome!

PR #636, opened 2018-04-30
git: always set GIT_CEILING_DIRECTORIES

philpep:
By default when git don't find a repository in current working dir, it
search for git repository in parent directories.

This may lead to unwanted behavior when the benchmarked repository is
nested inside other repo since in `checkout()` we expect to given path
to be the repository and re-clone it if it's not. When the given path is
not a repository we may run git commands (like a checkout) on unrelated
repo.

This bug occur in appveyor tests where `--basetemp` is set to a
directory inside asv checkout, and `test_repo.test_repo_git()` run a
checkout() after `rm -rf .git` causing tests to run a `git checkout -f
master` on asv repo.

We don't have similar issue with hglib since it use the `-R` option to
explicitly set the repository root.

### DIFF
--- a/asv/plugins/git.py
+++ b/asv/plugins/git.py
@@ -66,9 +66,10 @@ class Git(Repo):
         kwargs['cwd'] = cwd
         env = dict(kwargs.pop('env', os.environ))
         if cwd is not None:
-            env['GIT_CEILING_DIRECTORIES'] = ':'.join([
-                os.path.join(cwd, os.pardir),
-                env.get('GIT_CEILING_DIRECTORIES', '')])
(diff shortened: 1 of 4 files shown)


pv:
`dict(kwargs.pop('env', os.environ))` to avoid modifying input args.

philpep:
Ok fixed in 019644837ebb06999825d4584ca24ee59b531068

pv:
Needs some tuning and a test (I'll take a look now)

pv:
GIT_CEILING_DIRECTORIES is colon-separated, not nice on Windows :(

PR #643, opened 2018-05-23
Fix support for mercurial >= 4.5

philpep:
The behavior or 'hg log --follow-first -r <revset>' has changed in
mercurial 4.5: https://www.mercurial-scm.org/repo/hg/rev/a177c6aa055a

This cause 'asv run' with a single revision to resolve to all ancestors
of revision and following first parent instead of only the given single
revision.

Only use --follow-first only for get_branch_commits() which is needed to
draw the graph correctly in case of merges.

This fixes running tests with hg >= 4.5

### DIFF
--- a/asv/plugins/mercurial.py
+++ b/asv/plugins/mercurial.py
@@ -135,10 +135,9 @@ class Hg(Repo):
         rev = self._repo.log(self._encode(hash))[0]
         return int(rev.date.strftime("%s")) * 1000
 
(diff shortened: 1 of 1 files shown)


philpep:
Tests are breaking on old mercurial versions now. I'll fix this.

philpep:
I dropped the second commit which was breaking tests, I'll submit in a dedicated PR if needed.

PR #646, opened 2018-05-31
[WIP] Attempted fix for #645 

### LINKED ISSUE #645: asv cannot use conda environments for conda >= 4.4
Since conda 4.4, conda is no longer an executable script, but rather a bash function that shadows the script. (This is to enable the syntax `conda activate env-name`.)

As a consequence, when calling `asv run`, I get the error "Could not find 'conda' in PATH".

The conda executable still exists, but it is shadowed by this function:

```
$ type conda
conda is a function
conda () 
{ 
    if [ "$#" -lt 1 ]; then
        $_CONDA_EXE;
    else
        \local cmd="$1";
        shift;
        case "$cmd" in 
            activate)
                _conda_activate "$@"
            ;;
            deactivate)
                _conda_deactivate "$@"
            ;;
            install | update | uninstall | remove)
                $_CONDA_EXE "$cmd" "$@" && _conda_reactivate
            ;;
            *)
                $_CONDA_EXE "$cmd" "$@"
            ;;
        esac;
    fi
}
```

So it seems we can grab the conda executable from the `_CONDA_EXE` environment variable instead of the `PATH`. I'll have a look at this shortly, but pointers are welcome.

jni:
Fixes #645 

In conda >= 4.4, conda shipped with a conda bash function to enable the
syntax "conda activate env-name". This shadows the conda binary and
prevents it from being found in the PATH. However, the path to the conda
executable is found in the environment variable CONDA_EXE in this case

### DIFF
--- a/asv/util.py
+++ b/asv/util.py
@@ -279,6 +279,12 @@ def which(filename, paths=None):
 
     Raises an IOError if no result is found.
     """
+    # shortcut to conda for conda >= 4.4 in POSIX platforms
+    # conda is a bash function, not an executable, since 4.4, breaking
+    # path search. See https://github.com/airspeed-velocity/asv/issues/645
+    if filename == 'conda' and 'CONDA_EXE' in os.environ:
+        return os.environ['CONDA_EXE']
+
     # Hide traceback from expected exceptions in pytest reports
     __tracebackhide__ = operator.methodcaller('errisinstance', IOError)
 
--- a/asv/util.py
+++ b/asv/util.py
@@ -279,12 +279,6 @@ def which(filename, paths=None):
 
     Raises an IOError if no result is found.
     """
-    # shortcut to conda for conda >= 4.4 in POSIX platforms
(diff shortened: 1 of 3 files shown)


asv/util.py
```diff
@@ -279,6 +279,12 @@ def which(filename, paths=None):
 
     Raises an IOError if no result is found.
     """
+    # shortcut to conda for conda >= 4.4 in POSIX platforms
+    # conda is a bash function, not an executable, since 4.4, breaking
+    # path search. See https://github.com/airspeed-velocity/asv/issues/645
+    if filename == 'conda' and 'CONDA_EXE' in os.environ:
+        return os.environ['CONDA_EXE']
```
pv: This should be done in `plugins/conda.py`

jni: @pv of course, thanks for the pointer, I just went to the place that the error was coming from. Hold please!

jni:
This should be ready for another round of review. Thanks @pv!

pv:
thanks, merged

Issue #595, opened 2017-12-11
Stagger benchmarks to smooth background noise


jbrockmendel:
When running `asv continuous` to compare commits A and B, all the benchmarks from A run followed by all the benchmarks from B, e.g. "A.foo, A.bar, A.baz, B.foo, B.bar, B.baz".  Would it be feasible to instead run "A.foo, B.foo, A.bar, B.bar, A.baz, B.baz"?

(In fact because each benchmark is run multiple times, Ideally I would like the staggering to be even finer-grained.)

The thought here is that background processes make noise auto-correlated, so the running comparisons back-to-back may give more informative ratios.  (Based on amateur speculation)

pv:
See here https://github.com/pv/asv/commits/many-proc

Full staggering may not be so sensible, because you need to uninstall/install the project in between.

The general issue is also probably not so much background processes, but hardware causes for CPU performance fluctuations. These occur on laptops, but I guess desktop CPUs are have similar behavior vs. thermal throttling etc.

jbrockmendel:
The hardware issue sounds tough, but also sounds like you've put a lot of time and thought into this, which is reassuring.

What am I looking at with many-proc?  Is the idea to save+aggregate results across multiple runs?    That seems like a reasonable approach (and possibly easier to implement than staggering?).

Is the "not so sensible" an indication that I should give up on this?  The status-quo of running `asv continuous` repeatedly and hunting for which ratios are similar across many runs is... not ideal.

pv:
Staggering can be implemented (in the same way as multiple runs + aggregating results), but it will be slower because between each benchmark you need to uninstall/reinstall the project to the environment.

jbrockmendel:
>  it will be slower because between each benchmark you need to uninstall/reinstall the project to the environment.

Can you expand on that a bit?  Could the parent process spawn two env-specific processes which in turn spawn processes for each benchmark without switching in between?

Even if not, slower and more accurate is a tradeoff I'd be happy with.  I implemented asv for dateutil but the PR stalled because ATM the results are just way too noisy.

pv:
@jbrockmendel: you can perhaps try asv master branch now that the multi-process benchmarking is in there. E.g. `asv run ... -a processes=4` to override benchmark attributes from command line.

jbrockmendel:
I'll give it a try.  How do I enable the feature where it stores results to calculate statistics over multiple runs?  If we can get that working then stability becomes a problem we can throw hardware-hours at.

[...]

PR #647, opened 2018-06-02
Benchmark runner refactoring and interleaved benchmark sampling

pv:
Refactor benchmark runner by decoupling preparation/running/result printing steps from each other.

Also enable interleaved sampling runs, i.e. each benchmark run is split to multiple parts and these are run in an interleaved order. This enables each benchmark to sample over long-time variations in CPU and other background factors.


jbrockmendel:
> This enables each benchmark to sample over long-time variations in CPU and other background factors.

Is this related to the issue discussed in #595?

pv:
@jbrockmendel: yes, although it does not interleave between different commits.

PR #655, opened 2018-07-02
Split large command methods into smaller functions

jbrockmendel:
The most important change here is making `Compare.print_table` into a module function.  Calling `print_function` from `Continuous` instead of `Compare.print_function` makes it much easier to reason about the dependencies in the code.

### DIFF
--- a/asv/commands/continuous.py
+++ b/asv/commands/continuous.py
@@ -9,7 +9,7 @@ import six
 
 from . import Command
 from .run import Run
-from .compare import Compare
+from .compare import print_table
 
 from ..repo import get_repo
 from ..console import color_print, log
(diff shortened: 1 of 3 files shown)


asv/commands/continuous.py
```diff
@@ -123,3 +103,40 @@ def results_iter(commit_hash):
...
+    name : str
+    params : list
+    value : object
+    stats : list
+    version : str
+    """
+    for env in run_objs['environments']:
+        machine_name = run_objs['machine_params']['machine']
+        filename = results.get_filename(machine_name, commit_hash, env.name)
+        filename = os.path.join(conf.results_dir, filename)
```
pv: `conf` is undefined.

I'd prefer to keep this as an inline function in the Continuous method, however, it's short enough and the logic belongs there.

jbrockmendel: OK, I'll revert this part of the PR.

pv:
Closing: please reopen if further work done.

PR #659, opened 2018-07-06
Add a min_repeat benchmark attribute

### LINKED ISSUE #653: min repeat for long benchmarks with number = 1
Hello,

We have some time benchmarks that need `setup()` and `teardown()` called between each call of the time benchmark. So we set `number = 1`. The benchmark is parameterized and depending on the variant used and the hardware used, a single benchmark run can take between 1 minutes to 45 minutes.
To control how many repeat will be done we have to set `repeat` and `sample_time` on a per-variant basis in `setup()`, this works but it's still hardware dependent and can lead to having a result based on a single run (see `too_slow()` in `asv.benchmark.TimeBenchmark.benchmark_timing()`).

So we think about introducing a new `min_repeat` attribute, if set this will ensure the benchmark run at least `min_repeat` without regard to `sample_time`. What do you think ?

philpep:
When sample_time is lower than a single benchmark we run only one
sample. For some benchmark it's hard to set sample time to a proper
value to have at least a sufficient number of samples (depending on
hardware / parameterization etc).
Add a min_repeat benchmark attribute to ensure we return a value
computed on at least a given number of samples.

Closes #653

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -433,6 +433,7 @@ class TimeBenchmark(Benchmark):
 
     def _load_vars(self):
         self.repeat = _get_first_attr(self._attr_sources, 'repeat', 0)
+        self.min_repeat = _get_first_attr(self._attr_sources, 'min_repeat', 0)
         self.number = int(_get_first_attr(self._attr_sources, 'number', 0))
         self.sample_time = _get_first_attr(self._attr_sources, 'sample_time', 0.1)
         self.warmup_time = _get_first_attr(self._attr_sources, 'warmup_time', -1)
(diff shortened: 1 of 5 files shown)


pv:
An alternative to this without introducing new attributes could be to take exactly `repeat` samples if it is specified: gh-665

pv:
Went with the alternative --- if you think the min_repeat would still be useful, please rebase and reopen.

philpep:
Hi (and sorry for delay), 

I think the new behavior with `repeat` is better than before but I don't think it will address our issue, because depending on the commit being benchmarked, the benchmark can take x2 or more time, so it's not easy to pick a fixed value for `repeat` to have the best compromise between quality (stability) of the result and benchmark total time.

But I'm also ok to not add more complexity to the current algorithm, and, in the meantime we're considering an integration of asv with the perf module https://pypi.org/project/perf/ which might handle our complex use-cases. So I'll probably be back with this or another solution later. Thanks!

pv:
Another issue that seems to suggest min_repeat in some form: gh-677

pv:
IIUC, perf avoids this issue by not picking `repeat` automatically.

PR #683, opened 2018-07-21
Ensure a minimum number of runs when `repeat` is automatic 

### LINKED ISSUE #677: Big time difference for benchmarks that require high warmup
I was running a benchmark with latest asv 0.3dev0 and I observed a big time difference from asv 0.2.

* 0.2 https://github.com/poliastro/poliastro-benchmarks/blob/df6a71330f08c0c0ae1369818bc5af8106bba62f/results/ks1/d32f3ab8-conda-py3.6-matplotlib2.1-nomkl-numba-numpy1.14-scipy1.0.json#L31
* 0.3dev0 https://github.com/poliastro/poliastro-benchmarks/blob/f47fb57f2db0d76649ae7337867221b3738f057a/results/ks1/d3238535-conda-py3.6-matplotlib2.1-nomkl-numba-numpy1.14-scipy1.0.json#L34

After doing a git bisect, I arrived to [this commit](https://github.com/airspeed-velocity/asv/commit/db5cc300d39eb890292cf979d90f255b57dc9887) and [pull request](https://github.com/airspeed-velocity/asv/pull/493), which as far as I understand has not been backported to 0.2 yet.

This particular benchmark calls a function accelerated with `@jit(nopython=True)`, so I suspected that warmup time had something to do with it. After playing a little bit with the parameters, I found that this diff brought back the old time measurement with asv 0.3dev0:

```diff
diff --git a/benchmarks/examples.py b/benchmarks/examples.py
index 13e86d8..d034eed 100644
--- a/benchmarks/examples.py
+++ b/benchmarks/examples.py
@@ -3,3 +3,5 @@ from poliastro.examples import iss
 
 def time_propagate_iss_one_period():
     iss.propagate(iss.period)
+
+time_propagate_iss_one_period.repeat = 3
```

So there is ~~a UX issue and~~ a measurement issue:

* ~~The UX issue is that, as far as I know, there's no way to know _a posteriori_ how many repetitions, number of samples, etc did asv compute for a benchmark, as this information is not stored in the results.~~
* The measurement issue is that, with the current benchmark methodology, if a function requires a "big" warmup time (in this case, doing the numba JIT compilation), it will be run only once, which might or might not be representative of the intended result.

pv:
Safeguard the "too_slow" check by a requirement of minimum run count.

Even if the function is very slow, it's not sufficient to run it only
once as the first timing may be a fluke, e.g., due to JIT compilation.

Fixes #677

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -470,10 +470,12 @@ class TimeBenchmark(Benchmark):
         samples = [s/number for s in samples]
         return {'samples': samples, 'number': number}
 
-    def benchmark_timing(self, timer, repeat, warmup_time, number=0):
+    def benchmark_timing(self, timer, repeat, warmup_time, number=0,
+                         min_timeit_count=2):
         sample_time = self.sample_time
 
         start_time = time.time()
+        timeit_count = 0
 
         if repeat == 0:
             # automatic number of samples: 10 is large enough to
@@ -483,6 +485,8 @@ class TimeBenchmark(Benchmark):
 
             def too_slow(timing):
                 # stop taking samples if limits exceeded
+                if timeit_count < min_timeit_count:
+                    return False
                 if default_number:
                     t = 1.3*sample_time
                     max_time = start_time + min(warmup_time + repeat * t,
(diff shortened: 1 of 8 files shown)


pv:
Sure, reopen if it's not sufficient.

Juanlu001:
For my particular case, this worked beautifully.

* Commit: https://github.com/poliastro/poliastro-benchmarks/commit/6101b97895daced9e21348ec0e82624e19834d62
* Results: https://poliastro.github.io/poliastro-benchmarks/#examples.time_propagate_iss_one_period?machine=ks1

PR #699, opened 2018-08-16
Custom build commands

pv:
Allow defining custom install/uninstall/build commands in the configuration file.
See {} for synopsis.

Modify WheelCache -> BuildCache, which deals with generic cache
directories that probably are useful also for custom commands.
Moreover, set nonzero default build_cache_size.

Adjust the default pip/setup.py install commands to work with this system.
Also add new environment variables for build and caching.
Document how the configuration options work in reference docs.

- [ ] tests
- [ ] dogfooding


pitrou:
Thanks a lot @pv. AFAIK, this seems to work fine with Arrow's benchmark setup.

PR #736, opened 2018-09-09
CI: add Python 3.7

pv:


pv:
Will take another look when travis has py3.7

PR #748, opened 2018-10-01
Add an option for bypassing commit filtering for asv publish.

germanjoey:
Hello!

My team's been using asv for our project for the last couple of months and we really love it. However, we've noticed a problem where our graphs would occasionally lose all of their past data despite the results files still remaining present. Some investigation revealed the culprit: `asv publish` was filtering out benchmarks from commits that didn't appear via `git rev-parse --first-parent branch`. Our repo has a structure such that a condition like that doesn't make sense for us, so we'd rather just have all benchmarks shown.

As for the option's name, I went with "no-commit-filtering" to maintain the same naming scheme as "no-pull", but am happy to change it to whatever y'all think makes sense.

### DIFF
--- a/asv/commands/publish.py
+++ b/asv/commands/publish.py
@@ -60,6 +60,9 @@ class Publish(Command):
             '--no-pull', action='store_true', dest='no_pull',
             help="Do not pull the repository")
         parser.add_argument(
+            '--no-commit-filtering', action='store_true', dest='no_commit_filtering',
+            help="Include all commits, instead of filtering. Default is 'False'.")
+        parser.add_argument(
             'range', nargs='?', default=None,
             help="""Optional commit range to consider""")
         parser.add_argument(
@@ -75,7 +78,9 @@ class Publish(Command):
     def run_from_conf_args(cls, conf, args):
         if args.html_dir is not None:
             conf.html_dir = args.html_dir
-        return cls.run(conf=conf, range_spec=args.range, pull=not args.no_pull)
+
+        return cls.run(conf=conf, range_spec=args.range, pull=not args.no_pull,
+                        no_commit_filtering=args.no_commit_filtering)
 
     @staticmethod
     def iter_results(conf, repo, range_spec=None):
@@ -91,7 +96,7 @@ class Publish(Command):
                 yield result
 
     @classmethod
-    def run(cls, conf, range_spec=None, pull=True):
+    def run(cls, conf, range_spec=None, pull=True, no_commit_filtering=False):
         params = {}
         graphs = GraphSet()
         machines = {}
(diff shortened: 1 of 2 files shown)


pv:
The concept here is more about history "linearization" than "filtering". The graphs and analysis applied to them requires a linear ordering of commits, and there's probably no best way to do this with significantly branched history.

I guess what you intend here is having `Git.get_branch_commits` run without the `--first-parent` flag --- instead of putting all commits to all branches as this PR seems to do now? Or is the repository without long-lived branches?

Ideally, I'd like to see a bit more thought out proposal here with some room left for the different ways one might like to do the linearization.

pv:
The appveyor one looks like unrelated maybe temporary conda failure, I don't understand it either.

germanjoey:
> I guess what you intend here is having `Git.get_branch_commits` run without the `--first-parent` flag --- instead of putting all commits to all branches as this PR seems to do now (which doesn't make sense to me)?

Yeah, that's essentially what our goal was. Our thought was that "if we have a benchmark, we want it plotted" because we only had nightly benchmarks run against one particular branch. However, as you say, if a project had benchmarks on more than one branch, our feature certainly wouldn't be very useful!

> Ideally, I'd like to see a bit more thought out proposal here with some room left for the different ways one might like to do the linearization.

Thinking about this issue a bit more after reading your comment, it does sound like the best way to address the problem of "what to plot when a project has various benchmark results in a highly branched repo" is to change the behavior of "get_hashes_from_range" and then git/mercurial deal with it. 

I notice that both `asv/plugins/git.py` and asv/plugins/mercurial.py` have hardcoded default arguments; `--first-parent` for git, and `{0}, -rev` for mercurial. So, perhaps we could change the `--no-commit-filtering` flag to a `--commit-filter` argument that defaults to `None`. For each branch in conf.branches, `asv publish` would then:

a.) if `--commit-filter` is not specified, the respective defaults for each repo are used
b.) otherwise, the value for `--commit-filter` is used instead. 

If the loading results stage was split into two parts, so that commits could be reordered according to what was returned by `get_branch_commits`, this would even address the ordering problem that you mentioned as the user would then be able to change the sort order according to what flags were passed. (e.g. by passing `--date-order` or `--author-date-order` or `--topo-order` to git rev-list`).

What do you think?

pv:
Yes, I'd prefer to do it in `get_branch_commits` or `get_hashes_from_range`.

As for the user interface, I'd prefer a flag along the lines of `--commit-history=first-parent/date-order`, ... --- or a config file option, as I think for a given repository you don't need to change the value of the option very often.

PR #751 seems to be trying to do the same thing as this...

PR #751, opened 2018-10-10
add option for first-parent VCS flag (for publish/gh-pages only)

msarahan:
Conda's repo strategy is to branch off at releases and keep "merging up" so that master always contains all the commits from the past.  The --first-parent flag passed to git by ASV gets confused by this, and only shows stuff that has actually been committed on master (that's what it's for, I guess).  We wanted a way to not pass that flag, so that all of our commits show up on master.

This PR adds an option to disable passing that flag to git, or its equivalent to mercurial.  The default behavior does not change.  This PR only affects the behavior of the "publish" and "gh-pages" commands.

### DIFF
--- a/asv/commands/publish.py
+++ b/asv/commands/publish.py
@@ -66,6 +66,18 @@ class Publish(Command):
             '--html-dir', '-o', default=None, help=(
                 "Optional output directory. Default is 'html_dir' "
                 "from asv config"))
+        parser.add_argument(
+            "--first-parent", action="store_true", default=True, dest="first_parent",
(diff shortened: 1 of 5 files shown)


pv:
Seems to be same as #748 

I'd prefer a flag `--commit-history=first-parent/topo-date-order/...` instead, because first-parent and "commits in topological date order" are not the only options. Alternatively, this could as well be a config file option --- it's probably not likely you'd need to change it that often.

msarahan:
Indeed, sorry for not reviewing other PRs before putting this one up.  Closing in favor of #748 

If I can help on that one at all, please let me know.

PR #761, opened 2018-10-24
ENH add ncores field to asv-machine

pierreglaser:
Simple PR to add the number of CPUs into `.asv-machine.json`

### DIFF
--- a/asv/machine.py
+++ b/asv/machine.py
@@ -7,6 +7,7 @@ from __future__ import (absolute_import, division, print_function,
 import six
 
 import os
+import multiprocessing
 import platform
 import sys
 import textwrap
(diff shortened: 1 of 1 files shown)


asv/machine.py
```diff
@@ -144,6 +149,7 @@ def get_defaults():
         return {
             'machine': node,
             'os': "{0} {1}".format(system, release),
+            'ncores': os.cpu_count(),
```
pv: Should work on Python 2 still.
Also, should be `six.text_type`, not int.

asv/machine.py
```diff
@@ -119,6 +119,11 @@ class Machine(object):
          including its speed and class.  For example, 'Intel(R)
          Core(TM) i5-2520M CPU @ 2.50GHz (4 cores)'."""),
 
+        ("ncores",
```
pv: `num_cpu`?

pierreglaser: Actually I like more num_cpu, but it felt weird to use underscores in command line options.

pv:
thanks, merged (build failures due to conda flaking out again)

PR #762, opened 2018-10-25
Unbundle JSON-minify.

QuLogic:
Not sure if I got all the CI done right.

It appears this was bundled because there were no releases at the time, but there is one now.


pv:
I don't really want to unbundle it, it's basically just one function.

QuLogic:
I don't particularly mind either way; it's mostly to appease my [reviewer](https://bugzilla.redhat.com/show_bug.cgi?id=1515053#c8). I can just patch it downstream otherwise.

pv:
I think I don't see a strong case for unbundling in upstream, this feels bit too close to left-pad, and the dependencies have been purposefully kept minimal (e.g. no numpy etc).

PR #766, opened 2018-10-27
TST: serialize 'conda' command calls

pv:
Don't allow pytest-xdist to run conda-using tests in parallel, as it
appears conda is brittle vs. that.

Disable timeout= from LockFile usage, because 'lockfile' package has
hardcoded poll rate of timeout/10 which is too large here...


pv:
Dropping this --- it's essentially workaround for Conda bug.

PR #768, opened 2018-10-29
Allow `asv run` to consume list of hashes from a file

eteq:
This PR enables the listing of hashes asv should run on by providing a file.  Currently the format is intentionally limited to "one hash per line".  The user says they want to use said file by giving "HASHFILE:xxx.yyy" as the `range` parameter to `asv run`. While not strictly a "range", this seemed ok because "ANY", "EXISTING", etc were already options, so this seems not too much farther from "range" than those.

The use case for this is wanting to compare a specific hand-picked set of commits to so how performance on a particular feature evolves with major known (also probably #561, if I understand the suggestion there correctly). To try to also address the use case #562 is going for, I added "stdin" or "-" as options, in which case stdin is treated as the "hash file", although if that's too hacky I could drop that commit.

I know I could do one of the tricks suggested in #562 instead of this, but those are a lot harder to grok, and also seem to preclude the use of some of the nice multi-commit features of `asv run` like `--parallel` or `--interleave-processes`.

Happy to add docs/etc if need be - just "testing the waters" for now, as I found it quite frustrating to not have an easy way to just run on a list of known hashes.

Closes #562

### DIFF
--- a/asv/commands/run.py
+++ b/asv/commands/run.py
@@ -64,12 +64,15 @@ class Run(Command):
             repository, this is passed as the first argument to ``git
             log``.  See 'specifying ranges' section of the
             `gitrevisions` manpage for more info.  Also accepts the
-            special values 'NEW', 'ALL', and 'EXISTING'. 'NEW' will
-            benchmark all commits since the latest benchmarked on this
-            machine.  'ALL' will benchmark all commits in the project.
-            'EXISTING' will benchmark against all commits for which
-            there are existing benchmarks on any machine. By default,
-            will benchmark the head each configured branches.""")
+            special values 'NEW', 'ALL', 'EXISTING', and 'HASHFILE:xxx'.
+            'NEW' will benchmark all commits since the latest
+            benchmarked on this machine.  'ALL' will benchmark all
+            commits in the project. 'EXISTING' will benchmark against
+            all commits for which there are existing benchmarks on any
+            machine. 'HASHFILE:xxx' will benchmark only a specific set
+            of hashes given in the file named 'xxx', which must have
+            one hash per line. By default, will benchmark the head of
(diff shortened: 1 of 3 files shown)


asv/commands/run.py
```diff
@@ -194,6 +196,17 @@ def run(cls, conf, range_spec=None, steps=None, bench=None, attribute=None, para
         elif range_spec == "ALL":
             # All commits on each configured branches
             commit_hashes = repo.get_new_branch_commits(conf.branches, [])
+        elif isinstance(range_spec, six.string_types) and range_spec.startswith('HASHFILE:'):
+            hashfn = range_spec[9:]
+            if hashfn == '-' or hashfn.lower() == 'stdin':
```
pv: I'd prefer to only have `-`, no need for two ways to do it.

asv/commands/run.py
```diff
@@ -194,6 +196,17 @@ def run(cls, conf, range_spec=None, steps=None, bench=None, attribute=None, para
...
+                hashstr = sys.stdin.read()
+            elif os.path.isfile(hashfn):
+                with open(hashfn, 'r') as f:
+                    hashstr = f.read()
+            else:
+                log.error('Requested commit hash file "{}" is not a file'.format(hashfn))
+                return 1
+            commit_hashes = hashstr.strip().split('\n')
```
pv: `commit_hashes = [h.strip() for h in hashstr.split("\n") if h.strip()]`
To allow empty lines etc.

eteq:
Thanks @pv - I've pushed up a commit that I think addresses your inline review comments and adds to the test you suggested as well as some docs describing this feature in the `range` argument as well as the "using.rst" file.

pv:
thanks, merged

PR #771, opened 2018-11-06
ENH: remove memory addresses from param repr()

qwhelan:
The `pandas` benchmark suite sometimes uses functions as `params`, which largely works fine. The one issue is that the `repr()` of a function includes its memory address and can change over subsequent `asv` invocations (for normal functions) or even within a single `asv` run (for `lambda` functions). This change in `repr()` impacts the ability to track and compare a benchmark over time, leading to broken `asv compare` functionality like below:

```
Benchmarks that have stayed the same:

       before           after         ratio
     [24ab22f7]       [d7cef344]
              n/a          997±3μs      n/a  ctors.SeriesConstructors.time_series_constructor(<function SeriesConstructors.<lambda> at 0x7f0deb0800d0>, False)
              n/a         1.05±0ms      n/a  ctors.SeriesConstructors.time_series_constructor(<function SeriesConstructors.<lambda> at 0x7f0deb0800d0>, True)
      1.04±0.01ms              n/a      n/a  ctors.SeriesConstructors.time_series_constructor(<function SeriesConstructors.<lambda> at 0x7fd9d0bbc598>, False)
      1.06±0.02ms              n/a      n/a  ctors.SeriesConstructors.time_series_constructor(<function SeriesConstructors.<lambda> at 0x7fd9d0bbc598>, True)
```

This PR modifies `repr()` iff it appears to include a memory address, allowing proper comparison over time. The new output looks like this after a re-run:

```
       before           after         ratio
     [24ab22f7]       [d7cef344]
          995±7μs      1.03±0.02ms     1.04  ctors.SeriesConstructors.time_series_constructor(<function <lambda>>, False)
      1.05±0.01ms      1.09±0.05ms     1.03  ctors.SeriesConstructors.time_series_constructor(<function <lambda>>, True)

```

One unaddressed issue is that any suite with multiple `lambda` functions will now have name collisions in `params`. I intend to address this in `pandas` by creating normal named functions in the relevant suites. Alternative solutions would be to append distinct numbers or hashes to the `repr()`s, or to warn or raise if a user passes multiple `lambda` functions.

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -66,6 +66,8 @@ import pkgutil
 import traceback
 import contextlib
 from importlib import import_module
+from collections import Counter
+import warnings
 
 
 # The best timer we can use is time.process_time, but it is not
@@ -402,6 +404,21 @@ def check_num_args(root, benchmark_name, func, min_num_args, max_num_args=None):
     return ok
 
 
+def _repr_no_address(obj):
+    result = repr(obj)
+    address_regex = re.compile('^<.* at (0x[\da-fA-F]*)>$')
+    match = address_regex.match(result)
+    if match:
+        suspected_address = match.group(1)
+        # Double check this is the actual address
+        default_result = object.__repr__(obj)
+        known_address = address_regex.match(default_result).group(1)
+        if known_address == suspected_address:
+            result = re.sub(' at 0x[\da-fA-F]*>$', '>', result)
+
+    return result
+
+
 class Benchmark(object):
     """
     Represents a single benchmark.
@@ -461,7 +478,17 @@ class Benchmark(object):
                                                                    len(self._params))]
 
         # Exported parameter representations
-        self.params = [[repr(item) for item in entry] for entry in self._params]
+        self.params = [[_repr_no_address(item) for item in entry] for entry in self._params]
+        for i, param in enumerate(self.params):
+            if len(param) != len(set(param)):
+                counter = Counter(param)
+                dupe_dict = {name: 0 for name, count in counter.items() if count > 1}
+                for j in range(len(param)):
+                    name = param[j]
+                    if name in dupe_dict:
+                        param[j] = name + ' ({})'.format(dupe_dict[name])
+                        dupe_dict[name] += 1
+                self.params[i] = param
 
     def set_param_idx(self, param_idx):
         try:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,11 @@ Bug Fixes
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+- Uniqueness of ``repr()`` for ``param`` objects is now guaranteed
+  by suffixing unique identifier corresponding to order of
+  appearance. (#771)
+- Memory addresses are now stripped from the ``repr()`` of ``param``
+  elements, allowing comparison across multiple runs. (#771)
 
 
 0.4.1 (2019-05-30)
--- a/docs/source/benchmarks.rst
+++ b/docs/source/benchmarks.rst
@@ -97,6 +97,12 @@ The following attributes are applicable to all benchmark types:
 
   Note that ``setup_cache`` is not parameterized.
 
+  For the purposes of identifying benchmarks in the UI, ``repr()`` is called
+  on the elements of ``params``. In the event these strings contain memory
+  addresses, those adresses are stripped to allow comparison across runs.
+  Additionally, if this results in a non-unique mapping, each duplicated
+  element will be suffixed with a distinct integer identifier corresponding
+  to order of appearance.
 
 Timing benchmarks
 `````````````````
--- a/docs/source/writing_benchmarks.rst
+++ b/docs/source/writing_benchmarks.rst
@@ -229,7 +229,9 @@ as skipped for the parameter values in question.
 
 The parameter values can be any Python objects. However, it is often
 best to use only strings or numbers, because these have simple
-unambiguous text representations.
+unambiguous text representations. In the event the ``repr()`` output
+is non-unique, the representations will be made unique by suffixing
+an integer identifier corresponding to the order of appearance.
 
 When you have multiple parameters, the test is run for all
 of their combinations::
--- a/test/benchmark/params_examples.py
+++ b/test/benchmark/params_examples.py
@@ -41,6 +41,15 @@ class ParamSuite:
         del self.value
 
 
+class FunctionParamSuite:
+    params = [track_param,
+              lambda x: x]
+    param_names = ['func']
+
+    def time_func(self, func):
+        return func('foo')
+
+
 class TuningTest:
     params = [1, 2]
     counter = [0, 0]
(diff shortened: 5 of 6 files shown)


asv/benchmark.py
```diff
@@ -340,6 +340,15 @@ def get_source_code(items):
     return "\n\n".join(sources).rstrip()
 
 
+def _repr_no_address(obj):
+    result = repr(obj)
+    address_regex = re.compile('^<.* at 0x[\da-fA-F]*>$')
+    if address_regex.match(result):
+        result = '<{} {}>'.format(str(obj.__class__.__name__), str(obj.__name__))
```
pv: This needs at least to guard against missing attributes, as `obj` may be any object whatsoever, and the repr may contain anything. I'd actually prefer to guard this with `if isinstance(obj, ...):`, so that the way the transformation is applied is actually controlled.

qwhelan: Fair point, I'll see if I can accomplish the same result solely by modifying the `repr()` string, which should resolve your concern.

pv: I'm a bit worried about applying modifications to the repr string blindly --- the problem with this is that it's quite obscure behavior, and if something goes wrong it will be difficult to understand...

qwhelan:
I went with raising an exception on non-unique `repr()` approach, as it doesn't seem like having `lambda_1` through `lambda_n` will ever be that useful in the UI.

Here's the output on `pandas` without my patch (`numpy` and `scipy` don't need any changes):
```
$ asv dev
· Discovering benchmarks
·· Error running /home/chris/anaconda3/bin/python /home/chris/projects/asv/asv/benchmark.py discover /home/chris/projects/pandas/asv_bench/benchmarks /tmp/tmpom4ey47w/result.json
   STDOUT -------->

   STDERR -------->
     File "/home/chris/projects/asv/asv/benchmark.py", line 1230, in <module>
       main()
     File "/home/chris/projects/asv/asv/benchmark.py", line 1223, in main
       commands[mode](args)
     File "/home/chris/projects/asv/asv/benchmark.py", line 932, in main_discover
       list_benchmarks(benchmark_dir, fp)
     File "/home/chris/projects/asv/asv/benchmark.py", line 917, in list_benchmarks
       for benchmark in disc_benchmarks(root):
     File "/home/chris/projects/asv/asv/benchmark.py", line 825, in disc_benchmarks
       class_attr)
     File "/home/chris/projects/asv/asv/benchmark.py", line 771, in _get_benchmark
       return cls(name, func, sources)
     File "/home/chris/projects/asv/asv/benchmark.py", line 495, in __init__
       Benchmark.__init__(self, name, func, attr_sources)
     File "/home/chris/projects/asv/asv/benchmark.py", line 420, in __init__
       dupe_str))
   ValueError: Benchmark ctors.SeriesConstructors.time_series_constructor has duplicate string representations for param data_fmt:
   <function SeriesConstructors.<lambda>>: 7 copies

```

asv/benchmark.py
qwhelan: Okay, at the extreme end of options is to just use `IPython.lib.pretty.pprint()`, which handles this on a type-by-type basis (and produces identical output for functions as here). However, that still leaks memory addresses in a bunch of cases and adds a dependency to a sensitive file.

Would the following be good enough in your view?
* If the regex triggers, extract the suspected memory address.
* Then compute `object.__repr__(obj)` and extract a known memory address.
* Iff they match, delete the memory address substring.

qwhelan:
@pv Sorry, forgot to update that comment - I ended up switching approaches due to finding existing cases where it didn't make sense to force a `repr()` change. Here's current pandas output:
```
$ asv dev -b ctors.SeriesConstructors
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[  0.00%] ·· Benchmarking existing-py_home_chris_anaconda3_bin_python
[ 50.00%] ··· ctors.SeriesConstructors.time_series_constructor                                                                                                          ok
[ 50.00%] ··· ========================================================== ========== ==========
              --                                                               with_index
              ---------------------------------------------------------- ---------------------
                                       data_fmt                            False       True
              ========================================================== ========== ==========
               <function SeriesConstructors.<lambda> at 0x7fa7f0beba60>   1.49±0ms   1.81±0ms
                                         list                             3.12±0ms   3.33±0ms
               <function SeriesConstructors.<lambda> at 0x7fa7f0bebae8>   2.72±0ms   2.99±0ms
               <function SeriesConstructors.<lambda> at 0x7fa7f0bebb70>   6.91±0ms   7.37±0ms
               <function SeriesConstructors.<lambda> at 0x7fa7f0bebbf8>   2.95±0ms   3.03±0ms
               <function SeriesConstructors.<lambda> at 0x7fa7f0bebc80>   2.86±0ms   3.11±0ms
               <function SeriesConstructors.<lambda> at 0x7fa7f0bebd08>   2.74±0ms   3.08±0ms
               <function SeriesConstructors.<lambda> at 0x7fa7f0bebd90>   2.86±0ms   3.07±0ms
              ========================================================== ========== ==========
```
And here's the new output:
```
$ asv dev -b ctors.SeriesConstructors
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[  0.00%] ·· Benchmarking existing-py_home_chris_anaconda3_bin_python
[ 50.00%] ··· ctors.SeriesConstructors.time_series_constructor                                                                                                          ok
[ 50.00%] ··· ============================================ ========== ==========
              --                                                 with_index
              -------------------------------------------- ---------------------
                                data_fmt                     False       True
              ============================================ ========== ==========
               <function SeriesConstructors.<lambda>> (0)   1.41±0ms   1.81±0ms
                                  list                      3.16±0ms   3.40±0ms
               <function SeriesConstructors.<lambda>> (1)   2.77±0ms   2.96±0ms
               <function SeriesConstructors.<lambda>> (2)   6.90±0ms   7.48±0ms
               <function SeriesConstructors.<lambda>> (3)   2.88±0ms   2.94±0ms
               <function SeriesConstructors.<lambda>> (4)   2.94±0ms   3.13±0ms
               <function SeriesConstructors.<lambda>> (5)   2.88±0ms   3.04±0ms
               <function SeriesConstructors.<lambda>> (6)   4.02±0ms   2.98±0ms
              ============================================ ========== ==========
```

pv:
Can you update the documentation, adding explanation what it now does?

qwhelan:
@pv Added, still getting what appears to be an unrelated conda error.

Apologies for the delay!

qwhelan:
@pv Rebased and all green aside from a travis build that timed out (no reported failures)

pv:
But now it's a different approach again? Was this intended?

qwhelan:
@pv It's the same approach as the previous push, but distinct from the initial one - I ran into cases where distinct parameter cases where `repr()`s collided and could not be easily changed, so I went with the enumerating approach.

pv:
75e836c and 33072d1are not the same, please look at the PR diff.

qwhelan:
My local branch matches 75e836c, so not sure where the difference came from. Just force-pushed again and confirming this is the intended diff.

pv:
Thanks! And sorry for the long delay...

qwhelan:
Thanks for merging!

PR #779, opened 2018-12-17
CLN: fix solve_potts_approx benchmark

qwhelan:
API seems to have changed in `solve_potts_approx()` at some point, leading to a benchmark failure.

### DIFF
--- a/benchmarks/step_detect.py
+++ b/benchmarks/step_detect.py
@@ -17,4 +17,4 @@ class Simple:
             step_detect.detect_regressions(self.y)
 
     def time_solve_potts_approx(self):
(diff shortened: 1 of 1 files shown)


benchmarks/step_detect.py
```diff
@@ -17,4 +17,4 @@ def time_detect_regressions(self):
             step_detect.detect_regressions(self.y)
 
     def time_solve_potts_approx(self):
-        step_detect.solve_potts_approx(self.y, 0.3, p=1)
```
pv: Should be `gamma=0.3`. The option for setting `p` was removed.

qwhelan: This is fixed

PR #780, opened 2018-12-17
 BUG: change default timer to timeit.default_timer to fix resolution on Windows

### LINKED ISSUE #775: Default sample_time < clock resolution on Windows, leading to very noisy results
When running `asv` on Windows, it's quite common to see very noisy results like the following:
```
$ asv dev -b cut_timedelta
· Discovering benchmarks
· Running 4 total benchmarks (1 commits * 1 environments * 4 benchmarks)
[  0.00%] ·· Benchmarking existing-py_home_chris_anaconda3_bin_python
[ 12.50%] ··· reshape.Cut.time_cut_datetime                                                                                                                               ok
[ 12.50%] ··· ====== ==========
               bins
              ------ ----------
                4     15.6±0ms
                10     0±0ns
               1000   46.9±0ms
              ====== ==========
```
It's obvious from the above that we're using `time.time()` on Windows, which has a resolution of `15.6 ms`. This could be fine if `sample_time` were large enough, but the default of `10 ms` is only `~2/3` the clock resolution, leading to the time quantization to be especially pronounced.

Here's a log-log plot of the error ratio (std/mean) on the `pandas` asv benchmark suite when run on Windows:
log_log_hires

You can see that the first benchmark with an error better than 1% has to run for ~4s, or 400x the default `sample_time`!

The avenues I see for fixing `asv` on Windows are:
1. Switch to `timeit.default_timer()`
2. Increase `sample_time` by a couple orders of magnitude
3. Dynamically switch between `time.time()` and `time.clock()` based on expected runtime

I believe option 1 is by far the best option, but it's not clear to me whether there were wall vs cpu timing issues on Windows in the past that lead to the current selection of timer.

cc @topper-123

qwhelan:
This PR closes #775 in two different ways:
- The clock used for determining the `number` of runs to target `sample_time` is `time.time()`, which has a resolution of `15.6ms` on Windows. This leads to a lot of error in the proper scaling and is clearly better off being a wall time measurement
- Additionally, we change the default for `Benchmark.timer` to `timeit.default_timer()`. This fixes the poor benchmark resolution on Windows

To see proof of the latter, we can run the following:
```
asv run HEAD^..HEAD --record-samples
```
When testing this change, be careful to not use `asv run` to iterate through commits as it uses `asv/benchmark.py` from HEAD. Instead, step through with `git` and the above command.

For our baseline, we see the following output in `asv`'s own benchmarks:
```
    "results": {
        "step_detect.Simple.time_detect_regressions": {
            "result": [
                0.125
            ],
            "samples": [
                [
                    0.125,
                    0.125,
                    0.125,
                    0.125,
                    0.125,
                    0.125,
                    0.125,
                    0.125,
                    0.125,
                    0.109375
                ]
            ],
            "stats": [
                {
                    "ci_99": [
                        0.109375,
                        0.125
                    ],
                    "max": 0.125,
                    "mean": 0.1234375,
                    "min": 0.109375,
                    "number": 1,
                    "q_25": 0.125,
                    "q_75": 0.125,
                    "repeat": 10,
                    "std": 0.0046875
                }
            ]
        },
        "step_detect.Simple.time_solve_potts_approx": {
            "result": [
                0.015625
            ],
            "samples": [
                [
                    0.015625,
                    0.0,
                    0.015625,
                    0.015625,
                    0.015625,
                    0.015625,
                    0.015625,
                    0.0,
                    0.015625,
                    0.015625
                ]
            ],
            "stats": [
                {
                    "ci_99": [
                        0.0,
                        0.015625
                    ],
                    "max": 0.015625,
                    "mean": 0.0125,
                    "min": 0.0,
                    "number": 1,
                    "q_25": 0.015625,
                    "q_75": 0.015625,
                    "repeat": 10,
                    "std": 0.006250000000000001
                }
            ]
        }
    },
```
As you can see, the quantization of `process_time()` on Windows is *very* clear.

And now with this PR:
```
    "results": {
        "step_detect.Simple.time_detect_regressions": {
            "result": [
                0.1247326135635376
            ],
            "samples": [
                [
                    0.12278103828430176,
                    0.1226201057434082,
                    0.13060903549194336,
                    0.1316690444946289,
                    0.1291341781616211,
                    0.12431907653808594,
                    0.1271800994873047,
                    0.12331914901733398,
                    0.1220710277557373,
                    0.12514615058898926
                ]
            ],
            "stats": [
                {
                    "ci_99": [
                        0.1220710277557373,
                        0.1316690444946289
                    ],
                    "max": 0.1316690444946289,
                    "mean": 0.12588489055633545,
                    "min": 0.1220710277557373,
                    "number": 1,
                    "q_25": 0.12291556596755981,
                    "q_75": 0.128645658493042,
                    "repeat": 10,
                    "std": 0.003353031858757166
                }
            ]
        },
        "step_detect.Simple.time_solve_potts_approx": {
            "result": [
                0.011478066444396973
            ],
            "samples": [
                [
                    0.011333942413330078,
                    0.011632919311523438,
                    0.011541128158569336,
                    0.011580944061279297,
                    0.011635065078735352,
                    0.01182699203491211,
                    0.011229991912841797,
                    0.011066913604736328,
                    0.011046171188354492,
                    0.01141500473022461
                ]
            ],
            "stats": [
                {
                    "ci_99": [
                        0.011046171188354492,
                        0.01182699203491211
                    ],
                    "max": 0.01182699203491211,
                    "mean": 0.011430907249450683,
                    "min": 0.011046171188354492,
                    "number": 1,
                    "q_25": 0.011255979537963867,
                    "q_75": 0.011619925498962402,
                    "repeat": 10,
                    "std": 0.0002458795224520173
                }
            ]
        }
    },
```

In both cases, we see that the standard deviations have decreased - in the latter case, they're reduced by 25x. This effect would be even more pronounced if the mean didn't happen to be close to the `process_time()` quantum.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ New Features
 
 API Changes
 ^^^^^^^^^^^
+- Default timer changed from ``process_time()`` to
+  ``timeit.default_timer()`` to fix resolution issues on Windows.
+  Old behavior can be restored by setting ``Benchmark.timer = time.process_time``
 
 Bug Fixes
 ^^^^^^^^^
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -152,6 +152,8 @@ except ImportError:  # Python <3.3
         # Fallback to default timer
         process_time = timeit.default_timer
 
+wall_timer = timeit.default_timer
+
 
 def get_maxrss():
     # Fallback function, in case we don't have one that works on the
@@ -487,7 +489,7 @@ class TimeBenchmark(Benchmark):
         self.number = int(_get_first_attr(self._attr_sources, 'number', 0))
         self.sample_time = _get_first_attr(self._attr_sources, 'sample_time', 0.01)
         self.warmup_time = _get_first_attr(self._attr_sources, 'warmup_time', -1)
-        self.timer = _get_first_attr(self._attr_sources, 'timer', process_time)
+        self.timer = _get_first_attr(self._attr_sources, 'timer', wall_timer)
 
     def do_setup(self):
         result = Benchmark.do_setup(self)
@@ -547,7 +549,7 @@ class TimeBenchmark(Benchmark):
                          number, min_run_count):
 
         sample_time = self.sample_time
-        start_time = time.time()
+        start_time = wall_timer()
         run_count = 0
 
         samples = []
@@ -558,7 +560,7 @@ class TimeBenchmark(Benchmark):
                 return False
             if num_samples < min_repeat:
                 return False
-            return time.time() > start_time + warmup_time + max_time
+            return wall_timer() > start_time + warmup_time + max_time
 
         if number == 0:
             # Select number & warmup.
@@ -569,14 +571,14 @@ class TimeBenchmark(Benchmark):
             number = 1
             while True:
                 self._redo_setup_next = False
-                start = time.time()
+                start = wall_timer()
                 timing = timer.timeit(number)
-                wall_time = time.time() - start
+                wall_time = wall_timer() - start
                 actual_timing = max(wall_time, timing)
                 run_count += number
 
                 if actual_timing >= sample_time:
-                    if time.time() > start_time + warmup_time:
+                    if wall_timer() > start_time + warmup_time:
                         break
                 else:
                     try:
@@ -593,7 +595,7 @@ class TimeBenchmark(Benchmark):
                 self._redo_setup_next = False
                 timing = timer.timeit(number)
                 run_count += number
-                if time.time() >= start_time + warmup_time:
+                if wall_timer() >= start_time + warmup_time:
                     break
 
             if too_slow(1):
@@ -1089,14 +1091,14 @@ def main_run_server(args):
 
             # Wait for results
             # (Poll in a loop is simplest --- also used by subprocess.py)
-            start_time = time.time()
+            start_time = wall_timer()
             is_timeout = False
             while True:
                 res, status = os.waitpid(pid, os.WNOHANG)
                 if res != 0:
                     break
 
-                if timeout is not None and time.time() > start_time + timeout:
+                if timeout is not None and wall_timer() > start_time + timeout:
                     # Timeout
                     if is_timeout:
                         os.kill(pid, signal.SIGKILL)
@@ -1142,7 +1144,7 @@ def main_timing(argv):
     parser.add_argument("--number", action="store", type=int, default=0)
     parser.add_argument("--repeat", action="store", type=int, default=0)
     parser.add_argument("--timer", action="store", choices=("process_time", "perf_counter"),
-                        default="process_time")
+                        default="perf_counter")
     parser.add_argument("--json", action="store_true")
     parser.add_argument("statement")
     args = parser.parse_args(argv)
--- a/docs/source/benchmarks.rst
+++ b/docs/source/benchmarks.rst
@@ -138,20 +138,17 @@ Timing benchmarks
 - ``timer``: The timing function to use, which can be any source of
   monotonically increasing numbers, such as `time.clock`, `time.time`
   or ``time.process_time``.  If it's not provided, it defaults to
-  ``time.process_time`` (or a backported version of it for versions of
-  Python prior to 3.3), but other useful values are
-  `timeit.default_timer` to use the default ``timeit`` behavior on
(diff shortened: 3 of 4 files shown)


asv/benchmark.py
```diff
@@ -487,7 +489,7 @@ def _load_vars(self):
         self.number = int(_get_first_attr(self._attr_sources, 'number', 0))
         self.sample_time = _get_first_attr(self._attr_sources, 'sample_time', 0.01)
         self.warmup_time = _get_first_attr(self._attr_sources, 'warmup_time', -1)
-        self.timer = _get_first_attr(self._attr_sources, 'timer', process_time)
+        self.timer = _get_first_attr(self._attr_sources, 'timer', timeit.default_timer)
```
h-vetinari: is there a reason why you're not already using `wall_timer` here?

benchmarks/step_detect.py
```diff
@@ -17,4 +17,4 @@ def time_detect_regressions(self):
             step_detect.detect_regressions(self.y)
 
     def time_solve_potts_approx(self):
-        step_detect.solve_potts_approx(self.y, 0.3, p=1)
+        step_detect.solve_potts_approx(self.y, [0.3] * len(self.y), gamma=1)
```
h-vetinari: I don't understand how that change relates to the PR.

CHANGES.rst
```diff
@@ -6,6 +6,9 @@ New Features
 
 API Changes
 ^^^^^^^^^^^
+- Default timer changed from ``process_time()`` to
+  ``timeit.default_timer()`` to fix resolution issues on Windows.
+  Old behavior can be restored by setting ``Benchmark.timer = 'process_time'``
```
pv: ... = time.process_time
Assuming Python 3 is ok I think...

docs/source/benchmarks.rst
```diff
@@ -135,23 +135,20 @@ Timing benchmarks
 - ``min_run_count``: the function is run at least this many times during
   benchmark. Default: 2
 
-- ``timer``: The timing function to use, which can be any source of
+- ``timer``:
+  ..versionchanged:: 0.4
```
pv: Rst directives need blank lines.

pv:
Looks ok, modulo comments above.
Not sure how I'd write tests for this.
The CI failures seem to be due to some changes in Conda, not going to look fixing that in near future so they should be ignored.

benchmarks/step_detect.py
qwhelan: This is actually https://github.com/airspeed-velocity/asv/pull/779, which came about from my wanting to use `asv`'s own suite to demonstrate this change - but the one benchmark in the vicinity of `15.6ms` was broken due to a typo. It's necessary to reproduce the results above, but I will remove prior to merge.

asv/benchmark.py
qwhelan: Nope, will clean up

CHANGES.rst
qwhelan: Done

docs/source/benchmarks.rst
qwhelan: Thanks, ended up moving this down before the paragraph explaining the change, as that presentation is a bit cleaner

qwhelan:
@pv All changes implemented.

As far as testing, a simple regression test could look at the samples output from a test like:
```python
def test_pass():
    pass
```
For the baseline, we get:
```
            "samples": [
                [
                    1.939644470927057e-07,
                    1.939644470927057e-07,
                    0.0,
                    1.939644470927057e-07,
                    1.939644470927057e-07,
                    1.939644470927057e-07,
                    1.939644470927057e-07,
                    0.0,
                    1.939644470927057e-07,
                    1.939644470927057e-07
                ]
            ],
```
And for this PR:
```
            "samples": [
                [
                    1.2104441769390362e-07,
                    1.1922779948882427e-07,
                    1.2616902625495711e-07,
                    1.2339782410223055e-07,
                    1.2387379393762252e-07,
                    1.2318628195316746e-07,
                    1.2182183509171048e-07,
                    1.2359878914384048e-07,
                    1.2223434228238353e-07,
                    1.2985250392551827e-07
                ]
            ],
```

Simply looking for zeros (or values < clock resolution) seems like it would be sufficient (albeit stochastic, and much less reliable on slower machines like cloud VMs)

PR #783, opened 2019-01-01
Give up offline Conda in tests, seems to get too complicated

pv:


qwhelan:
@pv Just FYI, I added support for offline mode in `conda-build` due to the failures we were seeing in the `asv` builds. Might be worth trying turning this back on once this makes it into a release: https://github.com/conda/conda-build/pull/3328

PR #789, opened 2019-01-23
Asv compare variant

Lothiraldan:
This is a Work In Progress command to be able to compare two variants on the same commit. We are starting to use it right now to validate the UX and use-cases.

I've tried to refactor the `compare` command to reduce the amount of duplicated code to a maximum.

Before putting more effort on the code, having your view if this feature is useful and what form should it takes would be invaluable. I'm hesitating between a new `compare_variant` command or a `compare` flag?


pv:
I'd prefer retaining it all in `asv compare` comand, e.g. syntax `asv compare envname1:xxx envname2:yyy`.

Issue #650, opened 2018-06-20
Special dependency setup requirements.


BtbN:
I'm having some issues with some of our dependencies.
Namely GDAL and rasterio.

About GDAL:
GDAL has to be installed in a version matching the binaries of the system.
So the usual pip invocation looks like:
`pip install --upgrade "gdal==$(gdal-config --version)" --install-option="build_ext" --install-option="$(gdal-config --cflags | sed 's/-I/--include-dirs=/')"`

The usual solution here is to install it manually before installing our package.

About rasterio:
rasterio is a bit odd, as setup.py itself has a dependency on numpy. So numpy has to be installed individually before even trying to install rasterio.

Is there a way to still achive this? Maybe provide an external plugin somehow, that installs those odd dependencies?

pv:
If you need it now, it's possible to use a custom plugins like here:
https://github.com/OGGM/oggm/blob/master/asv_oggm_plugin.py
https://github.com/OGGM/oggm/blob/master/asv.conf.json

BtbN:
Yep, that's in fact written by me on my work account. I just forgot I had opened this issue.
It's working fine like that against asv master for us.

I'll leave it to you if wehter to keep this issue open for future enhancements or to just close it as solved.

PR #793, opened 2019-02-19
allow specifying a conda environment file for dependency management

### LINKED ISSUE #372: Allow conda environment specification from enviroment.yml file(s)
Conda already provides a dependency management option in the environment.yml file. I recently added one for my project in addition to the existing requirements.txt for pip. As I'm starting out with asv, I'm dismayed to find out that I need to specify my dependencies at yet another place.

This feature would enable:
- not having one more place to remember to update when requirements are bumped up.
- allowing PyPI packages in conda environments, which is not currently possible with asv.

mmohrhard:
Fixes #372.

We have a complex project that is completely managed by conda and duplicating the whole environment section in asv's config makes the benchmarks less maintainable.

I'm not sure if it is acceptable that this configuration value goes around all the dependency management currently implemented in asv. However, projects using conda environment files are already most likely keeping track of all dependencies in that one file.

### DIFF
--- a/asv/config.py
+++ b/asv/config.py
@@ -41,6 +41,7 @@ class Config(object):
         self.regressions_thresholds = {}
         self.plugins = []
         self.conda_channels = []
+        self.conda_environment_file = None
         self.build_command = None
         self.install_command = None
         self.uninstall_command = None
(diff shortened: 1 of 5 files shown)


pv:
An alternative could be to make the environment setup commands customizable, similarly as for the install/build/uninstall commands in #699.
This would probably also simplify the situation encountered in #650.

pv:
Thanks, merged.

Changed the behavior so that packages in `matrix` are installed afterward via `conda env update`, if there are any.

PR #794, opened 2019-02-20
ENH: add --commit-label option for 'run' command

### LINKED ISSUE #370: Help comparing benchmarks across virtualenvs
Sorry if this is the wrong place to ask.  Sorry too that I am sure I am missing something obvious.

I am using asv on Windows : https://github.com/numpy/numpy/issues/5479

I'm comparing performance of numpy in two virtualenvs, one with MKL linking, the other with ATLAS linking.

I told `asv machine` that I have two machines one called 'mike-mkl' and the other 'mike-atlas', identical other than the names.

Then I ran the benchmarks in the MKL virtualenv as:

```
cd numpy/benchmarks
asv run --python=same --machine=mike-mkl
```

In the ATLAS virtualenv:

```
cd numpy/benchmarks
asv run --python=same --machine=mike-atlas
```

I was expecting this to store the two sets of benchmarks, but it appears from the `results/benchmarks.json` file size and the `asv preview` output, that they overwrite each other.

What is the right way to record / show these benchmarks relative to one another?

jeremiedbb:
Fixes #370 

Add a `--commit-label` option to be used when running the benchmarks against an existing environment.
It will be used in place of the commit hash (which is `None` in that case) in order to be able to save the results.

### DIFF
--- a/asv/commands/run.py
+++ b/asv/commands/run.py
@@ -92,6 +92,11 @@ class Run(Command):
             benchmark functions faster.  The results are unlikely to
             be useful, and thus are not saved.""")
         common_args.add_environment(parser)
+        parser.add_argument(
+            "--commit-label", "-c", default=None,
+            help="""Commit label to use in place of the commit hash
+            when benchmarking against an existing environment. If
+            provided, the results will be saved.""")
         common_args.add_launch_method(parser)
         parser.add_argument(
             "--dry-run", "-n", action="store_true",
@@ -136,7 +141,7 @@ class Run(Command):
             conf=conf, range_spec=args.range, steps=args.steps,
             bench=args.bench, attribute=args.attribute, parallel=args.parallel,
             show_stderr=args.show_stderr, quick=args.quick,
-            profile=args.profile, env_spec=args.env_spec,
+            profile=args.profile, env_spec=args.env_spec, commit_label=args.commit_label,
             dry_run=args.dry_run, machine=args.machine,
             skip_successful=args.skip_existing_successful or args.skip_existing,
             skip_failed=args.skip_existing_failed or args.skip_existing,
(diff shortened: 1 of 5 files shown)


pv:
Looks ok. Needs test, however.

jeremiedbb:
It was a little bit less straightforward because the commit hash is truncated to it's 8 first characters for the results file name. However, with provided commit labels I think we want to keep their full length, because the risk of collision is quite high (for instance 'test_label_1' and 'test_label_2' would be cut to 'test_lab').

I chosed to add a prefix for the commit label and check for this prefix when dealing with the result file name. Do you think it's reasonable ?

jeremiedbb:
I added a test. I wasn't sure where to put it so I made a new test file for the run command. Is there a better place for it ?

asv/commands/run.py
```diff
@@ -386,11 +391,15 @@ def iter_rounds_commits():
...
-                        skip_save = (dry_run or isinstance(env, environment.ExistingEnvironment))
+                        skip_save = dry_run or (isinstance(env, environment.ExistingEnvironment)
+                                                and commit_label is None)
 
                         skip_list = skipped_benchmarks[(commit_hash, env.name)]
                         benchmark_set = benchmarks.filter_out(skip_list)
 
+                        if isinstance(env, environment.ExistingEnvironment) and commit_label is not None:
+                            commit_hash = 'label-' + commit_label.replace('-', '_')
+
```
jeremiedbb: I replace the `'-'` characters by `'_'` because the result filename is split by `'-'` in several places. It should avoid weird behaviors, but may modify the label provided by a user.

jeremiedbb:
The failing tests seems unrelated and are already failing on master [https://travis-ci.org/airspeed-velocity/asv/jobs/506823174](https://travis-ci.org/airspeed-velocity/asv/jobs/506823174)

pv:
I think the main use here is to specify existing commit hashes, when using an existing environment.

So I changed this to `--set-commit-hash`, and removed the label prefix. The prefix probably prevents the use case of using asv to run benchmarks but not manage the environment.

PR #795, opened 2019-03-06
Add warning on publish if no branch was found.

Alphare:
When the history has been modified (either with draft changesets in hg, or with commit rebasing in git), it is possible that no corresponding branch be found when publishing. There is no real way of automatically fixing the issue, so we at least have to log a warning.

However, it does not seem like there are tests for logs except the console tests for indentation, and I feel like this change warrants one. 
Do you have a preferred method of going about this?

### DIFF
--- a/asv/commands/publish.py
+++ b/asv/commands/publish.py
@@ -171,6 +171,14 @@ class Publish(Command):
             for results in cls.iter_results(conf, repo, range_spec):
                 log.dot()
 
+                branches_for_commit = [branch for branch, commits in branches.items() if
+                                       results.commit_hash in commits]
+
+                # Print a warning message if we couldn't find the branch of a commit
+                if not len(branches_for_commit):
+                    msg = "Couldn't find %s in %s branches"
+                    log.warning(msg % (results.commit_hash, branches.keys()))
+
                 for key in results.get_result_keys(benchmarks):
                     b = benchmarks[key]
                     b_params = b['params']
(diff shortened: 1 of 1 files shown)


Alphare:
I've just amended my commit with an improvement to speed and readability by extracting the list of branches for the current commit out of the loop.

asv/commands/publish.py
```diff
@@ -171,6 +171,14 @@ def copy_ignore(src, names):
             for results in cls.iter_results(conf, repo, range_spec):
                 log.dot()
 
+                branches_for_commit = [branch for branch, commits in branches.items() if
+                            results.commit_hash in commits]
+
+                # Print a warning message if we couldn't find the branch of a commit
+                if not len(branches_for_commit):
```
philpep: `len()` is useless here.

asv/commands/publish.py
```diff
@@ -171,6 +171,14 @@ def copy_ignore(src, names):
             for results in cls.iter_results(conf, repo, range_spec):
                 log.dot()
 
+                branches_for_commit = [branch for branch, commits in branches.items() if
+                            results.commit_hash in commits]
```
philpep: NIT: this introduce new non-pep8 code.

asv/commands/publish.py
Alphare: I figured this was more explicit that way, it's almost a question of code style. If you think it's cleaner, I'll remove it

pv: len() is not recommended

pv:
If you want to add general tests for logging, that's of course welcome.
However, perhaps best to add those in a separate PR.

Otherwise the change here looks ok, modulo the style comment.

pv:
Thanks, merged. Any tests are welcome in a separate PR.

PR #806, opened 2019-04-05
Use pip install --force-reinstall flag

### LINKED ISSUE #805: 'asv run' fails on import
When trying to setup asv on my project, it fails to import the project in any benchmark file when running `asv run`. Here is an example how to reproduce with an open-source project (`requests`):
```
git clone https://github.com/kennethreitz/requests.git
cd requests
asv quickstart
<choose option 1>
<add `import requests` to the top of benchmarks/benchmarks.py>
asv run
```
This ends up with the error `ModuleNotFoundError: No module named 'requests'`. However when launching the created python env in `.asv/env`, `import requests` doesn't lead to any error. What am I missing?

pv:
Under some conditions, pip appears to wrongly think the packages is
available in the environment, so tell it to install it always.

Closes #805

In the problematic case `asv run -v` says
```
·· Running 'xxx/python -mpip install /home/pauli/tmp/xxx/requests/.asv/env/a80b4bcff004e7d21aafa78a5932418b/asv-build-cache/9742da7f91f595796a412a8150aacc00191be039/requests-2.21.0-py2.py3-none-any.whl'
   OUTPUT -------->
   Requirement already satisfied: requests==2.21.0 from file:///xxx/requests/.asv/env/a80b4bcff004e7d21aafa78a5932418b/asv-build-cache/9742da7f91f595796a412a8150aacc00191be039/requests-2.21.0-py2.py3-none-any.whl in xxx/requests/.asv/env/a80b4bcff004e7d21aafa78a5932418b/project (2.21.0)
```
But I don't know precisely why this occurs (hence no test). However, adding the `--force-reinstall` flag appears to fix it.

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -660,8 +660,10 @@ class Environment(object):
         """
         cmd = self._install_command
         if cmd is None:
-            # Run pip via python -m pip, avoids shebang length limit on Linux
-            cmd = ["python -mpip install {wheel_file}"]
+            # Run pip via python -m pip, avoids shebang length limit on Linux.
+            # Ensure --force-reinstall so that package being present e.g. on cwd
+            # does not mess things up.
+            cmd = ["python -mpip install --force-reinstall {wheel_file}"]
 
         if cmd:
             commit_name = repo.get_decorated_hash(commit_hash, 8)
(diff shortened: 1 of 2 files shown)


yashasvi-goel:
I'm facing a similar issue.
I've tried this solution(#805 #806 ). It isn't solving the issue for me. I've tried updating my `pip` as well. 
I've double checked my `asv.conf.json`. Is there any other solution to fix this issue.

PR #807, opened 2019-04-06
Add a 'hide-legend' graph option

Alphare:
Legends of graphs with many lines can take up too much space,
this simple button will hide the graph legend on demand.


pv:
Thanks! (I swapped the bool option to show-legend)

Issue #165, opened 2014-12-27
Matrix for environment variables


snorfalorpagus:
Is there an existing feature for managing environment variables in a similar way to module dependencies?

For example, I'd like to run my benchmarks `WITH_FOO=1` and `WITH_FOO=0`, then have the benchmarks call a function with this as an argument, e.g.:

```
if os.environ['WITH_FOO'] == '1':
    # do something
else:
    # do something else

def time_something():
    # usual benchmark stuff
```

The purpose of this is that the [Shapely](https://github.com/Toblerity/Shapely) module has an optional "speedups" module that uses compiled cython instead of ctypes. The changes are global (i.e. many functions benefit from the speedups module), and I'd like to be able to benchmark with and without them, without having to duplicate a lot of code.

mdboom:
There isn't currently a way to do this, but it's a very good idea.  I'd definitely accept a patch, and if I find the time for this, I'll self-assign first.

yarikoptic:
I would love to see this feature implemented too!  My endless kudos to the one who implements it ;)
In our case we would love to benchmark @datalad under different default versions of git-annex repositories.

marmoute:
We would also be happy to have this kind of feature ☺

PR #809, opened 2019-04-24
Add env variable matrix to pass to tests and build

Alphare:
Hello,

This PR introduces support for environment variables, as askied in #165 .
When discussing the feature during the [Mercurial mini-sprint](https://octobus.net/blog/2019-04-17-mercurial-mini-sprint-2019-04.html) we held in Paris earlier this month, it became apparent that there would be a need for specifying whether or not a given environment variable should result in an additional build. Thus, environment variables can be defined as being "build" or "non_build". 
This decision has shaken up at least of core feature of ASV: how to identify an `Environment`, in that multiple `Environment`s can henceforth refer to the same build on the file system, allowing for quicker build time and less disk usage waste.
 
This PR is very close to completion (5 tests failing... I haven't been able to figure out why as of yet), but since I'm most likely not going to have time to finish it for a few weeks I figured I'd create it to kick-start the discussion that I'm sure will need to happen.

Closes: #165

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -20,6 +20,7 @@ import subprocess
 
 import six
 
+from asv.config import normalize_env_matrix
 from .console import log
 from . import util
 from . import build_cache
@@ -166,10 +167,11 @@ def match_rule(target, rule):
     return True
 
 
-def get_env_name(tool_name, python, requirements):
-    """
-    Get a name to uniquely identify an environment.
-    """
+def _get_env_name(tool_name,
+                  python,
+                  requirements,
+                  env_vars_combination,
+                  build):
     if tool_name:
         name = [tool_name]
     else:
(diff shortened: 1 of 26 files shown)


pv:
I'd perhaps prefer to make the split below Environment instead of above it: i.e. separate EnvironmentConfiguration (runtime configuration including specific environment variables) from Environment (the conda/virtualenv directory), instead of special-casing the handling of environment variables. Different EnvironmentConfiguration instances could use the same Environment.

This could also enable marking packages in the `matrix` as "non-build". Quite often with Python packages, the wheel can be built also without the required packages, and the build result is independent of them. If the build step is slow (e.g. c++ code), this can be useful.

pv:
Ok, I read it now more carefully, and the overall implementation approach seems fine. The association of Environment with an unique installation directory is less important than their association with specific runtime environment, and we can well relax the former. Since environment setup is idempotent, there's no real need to split the class.

One place that needs more careful handling is the parallel creation of environments. For simplicity, we could scrap the parallelism here, as I doubt it's very useful in practice since environments are long-lived.

Minor comments:
- `env_vars_combination` -> `env_vars` (the `requirements` option is not called `requirements_combination` either)
- `_test_` -> `_benchmark_` or `_bench_` (benchmarks are not tests)
- Please avoid whitespace changes in parts of code you are not touching.

Style comments:
- Linewidths up to 120 are OK.
- Splitting things such as `get_*_env_vars` to tiny functions makes it more difficult to read --- one function with kwargs seems preferable here.
- For tests with data sets, not using `pytest.mark.parametrize` is preferable --- for loop is better.

pv:
Rebase + minor refactoring + small bugfixes

Alphare:
Thanks for your detailed thoughts and your help.

Indeed, marking some dependencies as non-build could be very useful as well. Not sure if we want to introduce that in this PR or in a separate one, though.

A few comments:

- Why do you say that a for loop is preferable to `parametrize`? I only see a (subjective) syntactic readability issue with `parametrize`, otherwise I think it presents clear advantages, a few examples: separate nodes in `pytest`, test output is easier to read, tests don't stop at the first failure in the set.
- I think building environments in parallel is still very important. At Octobus, we lose a significant amount of time building environments, the faster the better.
- I'm not sure that the current file naming scheme is viable. In fact, this change makes it significantly worse by appending environment variables. Windows' 256 path length limitation will be hit very easily, and it kind of gives the impression of a half-baked solution for storing metadata.

If you think I should tackle a specific task let me know. Otherwise I don't want to start working on the same thing as you.

asv/environment.py
```diff
@@ -327,14 +327,11 @@ def get_environments(conf, env_specifiers, verbose=True):
...
 
-        env_matrix_combinations = iter_env_matrix_combinations(conf.env_matrix)
-
-        if not env_matrix_combinations:
-            env_matrix_combinations = [{}]
-
         for requirements in requirements_iter:
             python = requirements.pop('python')
 
+            env_matrix_combinations = iter_env_matrix_combinations(conf.env_matrix)
```
Alphare: Putting this iterable outside of the loop was deliberate since it is invariant. Is there something I'm missing?

pv:
Separate PR is probably better for the rest of the matrix.

I think the biggest issue at this point would be to actually test this in real use and see if problems appear.

For parametrize, readability counts: fitting the function in one screen vertically, and the dataset goes into wrong place with the decorator (above test function name). The test cases here also seemed too small for separate nodes to me, and stopping on first failure is probably not a problem here, as the code is deterministic. Matters of taste of course.

The parallelization was easy to fix, so that's now taken care of.

The file name switches to using the env hash if the name becomes too long.

pv:
It's a generator, so you don't get a new iterator for it each time the for loop starts, unlike for a list.

asv/environment.py
Alphare: Ha, of course "it's a generator".

pv:
I'm probably not going to touch this before the weekend, so if you want to make changes, please feel free.

PR #828, opened 2019-06-06
FIX do not try to test abstract Benchmark classes

### LINKED ISSUE #827: asv  tries to instanciate abstract benchmark classes
Title says it all (reproducer below)
```python
from abc import ABCMeta, abstractmethod


class AbstractBenchmarkClass(metaclass=ABCMeta):

    @property
    @abstractmethod
    def obj(self):
        raise NotImplementedError

    def time_do_stuff(self):
        return id(self.obj)


class ConcreteBenchmarkClass(AbstractBenchmarkClass):
    obj = 10
```

Using this as a benchmark file leads to 
```
   TypeError: Can't instantiate abstract class AbstractBenchmarkClass with abstract methods obj

```

pierreglaser:
Fixes #827

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -916,7 +916,8 @@ def disc_benchmarks(root, ignore_import_errors=False):
             (k, v) for k, v in module.__dict__.items()
             if not k.startswith('_')
         ):
-            if inspect.isclass(module_attr):
+            if (inspect.isclass(module_attr) and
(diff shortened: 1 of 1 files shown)


pierreglaser:
@pv how do you feel about this? Does it need a test?

pv:
This part of benchmark discovery probably needs other work (#610), possibly also limiting names per user configuration, but this looks ok to include.

PR #829, opened 2019-06-07
ENH: When encountering multiple benchmark matches in find/profile, use exact match if unambiguous

qwhelan:
When attempting to run `asv find` or `asv profile` on a suite with benchmarks that contains substrings of other benchmarks (eg, both `time_iter` and `time_iter_preexit`), it is not possible to run the substring benchmark directly due to the following exception:
```
 * 'time_iter' matches more than one benchmark
```
This PR adds support for this case by detecting that it is a full and unique match, issuing a warning, and continuing:
```
 * 'time_iter' matches more than one benchmark, using exact match
```

### DIFF
--- a/asv/commands/find.py
+++ b/asv/commands/find.py
@@ -100,8 +100,13 @@ class Find(Command):
             log.error("'{0}' benchmark not found".format(bench))
             return 1
         elif len(benchmarks) > 1:
-            log.error("'{0}' matches more than one benchmark".format(bench))
-            return 1
+            exact_matches = benchmarks.filter_out([x for x in benchmarks if x != bench])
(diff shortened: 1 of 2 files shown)


pv:
Does `^time_iter$` work? Here, fnmatch would be quite much more user-friendly...

qwhelan:
@pv On its own, no:
```
· '^timeseries.Iteration.time_iter$' benchmark not found
```
With some manipulation of the regex, this works:
```
 '^timeseries.Iteration.time_iter\(.*\)$'
```
Escaping is essential, otherwise the parens create a group and match on both benchmarks:
```
· '^timeseries.Iteration.time_iter(.*)$' matches more than one benchmark
```

Issue #609, opened 2018-01-28
Consider making JSON output more compact


astrofrog:
At the moment, the result JSON files include the names of each tests several times, e.g.

```json
    "benchmark_version": {
        "coordinates.FrameBenchmarks.time_init_array": "c674b97d5a26e5c61c240678ebdcc4a8fbb346db4aacedbb7868aefb3230f080",
        "coordinates.FrameBenchmarks.time_init_nodata": "c3d603a1d100c4d8f50750ddfac91c7db732186bbe5091b354ab1ee7bd4703d5",
        "coordinates.FrameBenchmarks.time_init_scalar": "defe667485d7109897f4a6a2770d80eb89ff9295c31bc941252bf4b8f6e15288",
...
    "ended_at": {
        "coordinates.FrameBenchmarks.time_init_array": 1517086912000,
        "coordinates.FrameBenchmarks.time_init_nodata": 1517086915000,
        "coordinates.FrameBenchmarks.time_init_scalar": 1517086918000,
...
    "started_at": {
        "coordinates.FrameBenchmarks.time_init_array": 1517086908000,
        "coordinates.FrameBenchmarks.time_init_nodata": 1517086912000,
        "coordinates.FrameBenchmarks.time_init_scalar": 1517086915000,
...
    "results": {
        "coordinates.FrameBenchmarks.time_init_array": {
```

For projects with many tests and commits, it would be more efficient to store this as a list of benchmarks with 'stared_at', 'results', and 'ended_at' keys for each one, to avoid repeating the long test names.

It would also be a bit more space efficient to store sections like:

```json
            "stats": [
                {
                    "ci_99": [
                        0.0023126445813953537,
                        0.002361580779069762
                    ],
                    "max": 0.0023916572325581296,
                    "mean": 0.0023257523279069763,
                    "min": 0.0023118635116279058,
                    "n": 10,
                    "q_25": 0.002314243011627908,
                    "q_75": 0.002323088261627908,
                    "std": 2.2691405732192584e-05,
                    "systematic": 0.0
                }
            ]
        },
```

without all the indentation as there is a lot of whitespace there. The precision of the values could probably also be reduced.

Finally, the hashes for the benchmark versions probably don't need to be as long as they are:

```json
        "coordinates.FrameBenchmarks.time_init_array": "c674b97d5a26e5c61c240678ebdcc4a8fbb346db4aacedbb7868aefb3230f080",
```

[...]

PR #847, opened 2019-07-15
More compact results json format

pv:
Avoid repeating key names, drop unused data, and store stats/duration
at reduced precision. Gives roughly 3x space saving. Probably no further
significant reductions can be made as long as using text format.

Closes: #609


pv:
Ok, I guess I'm happy with this and dogfooded sufficiently.

PR #858, opened 2019-07-23
Add `fix_setup` method to clean up if setup fails

Alphare:
If the `setup` method fails, the tests session waits for timeout (1800s per
benchmark!) provided a sub process had been created beforehand. This lead to
excruciatingly long test sessions, appearing fully stuck.


Alphare:
Maybe it would be a better idea to allow this method to return a new exception to be raised, because I'm not certain of the use case of silencing the exception... or maybe do both?

pv:
Why use `try: expect:` in your `setup` instead?

Alphare:
This is easier if I have multiple setup methods, or when using inheritence. I'd say it's arguably also uglier. My real sweet spot would be something akin to `pytest` fixtures, but that's for another day. :)

Alphare:
I've remove the "raise or not" mechanism, which fixes the issue of multiple `fix_setup` not being able to run if one raises.

pv:
I see. That the setup/teardown routines are not paired is a design problem.

If the fixture-type behavior is wanted, how about instead a `finalize_teardown` hook, which runs last also when setup/call/teardown failed? If you like, it could even receive an exc_info argument for any exceptions.

Alphare:
Sure, this was a bit of a placeholder name, I didn't want to waste time on that considering we were probably going to make changes in this PR.

This `finalize_teardown` method might be next best thing to real composable fixtures. I'll have a jab at it.

asv/benchmark.py
```diff
@@ -1151,17 +1157,21 @@ def main_run(args):
...
+        try:
+            if skip:
+                result = float('nan')
+            else:
+                result = benchmark.do_run()
+                if profile_path is not None:
+                    benchmark.do_profile(profile_path)
+        finally:
+            benchmark.do_teardown()
+    except BaseException as e:
```
pv: `finally:` I think

Alphare: Using `finally` would mean resorting to using `sys.exc_info()` to get the exception info and checking if it differs from `(None, None, None)`, which to me feels more convoluted than just using an `except` clause.

pv: Yes, but the point is that the finalize should run always, not only for exceptions.

Alphare:
Note: I am aware of the CI failures but need to figure #871 out before attempting to fix the tests since it would be very time consuming and spammy otherwise to just force-push every time.

HaoZeke:
@Alphare could you give maintainers access to modify the PR? I'd like to rebase and clean this up.

PR #861, opened 2019-08-05
Print if setup was skipped with the corresponding message if provided

Alphare:
I find it very useful to see if a benchmark was skipped at all during execution and to get the reason as they can be quite intricate sometimes.

I was not sure how to print the message, as this is a peculiar file. I chose `stderr` since it is unbuffered and feels semantically right, but I have no strong opinions on the matter.

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -548,8 +548,9 @@ class Benchmark(object):
         try:
             for setup in self._setups:
                 setup(*self._current_params)
-        except NotImplementedError:
+        except NotImplementedError as e:
             # allow skipping test
+            print("asv: skipped: {!r} ".format(e))
             return True
         return False
 


Alphare:
Sorry for the force-push, I had an issue with git, I'm too used to hg now, heh.

pv:
stdout/stderr does not matter, but use `print()` and format message as in here: https://github.com/airspeed-velocity/asv/blob/b64bce7f6f982680ca7ee7f1877fe1d787b04216/asv/benchmark.py#L1118-L1119

pv:
However, the benchmark status should say `n/a` when it's skipped. Is this additional message necessary then?

Alphare:
> However, the benchmark status should say `n/a` (not `failed`) when it's skipped. Is this additional message necessary then?

It does, but I'm always left wondering why it happens and had to debug a few times to get that information, which is important in our project.

pv:
ok, I guess a bit more noise with `-e`  won't hurt much.

asv/benchmark.py
```diff
@@ -548,8 +548,9 @@ def do_setup(self):
         try:
             for setup in self._setups:
                 setup(*self._current_params)
-        except NotImplementedError:
+        except NotImplementedError as e:
             # allow skipping test
+            print("Skipped: {!r} ".format(e.message if e.message else "no message provided."))
```
pv: `"asv: skipped: {!r}"` so it can be distinguished from messages printed by user code

Alphare:
FYI, I'm on vacation until the 20th. I should be able to update my PRs pretty soon after.

PR #863, opened 2019-08-07
Sort the environment variables when building the environment name

Alphare:
The environment variables were added after sorting the list. This makes tests unreliable.


pv:
CI fail. `sorted(six.iteritems(env_vars))` may be better so they're not mixed in with package names.

PR #864, opened 2019-08-10
MAINT: switch from deprecated lockfile to filelock

qwhelan:
The `lockfile` package is deprecated: https://pypi.org/project/lockfile/
```
Note: This package is deprecated
```
Additionally, the upstream repo has gone out of its way to hide the source code: https://opendev.org/openstack/pylockfile/

The `filelock` package implements an equivalent API, is actively maintained, and already included in the base dependencies by anaconda.


pv:
thanks

PR #865, opened 2019-08-15
Add --strict flag to exit non-zero on benchmark failures

### LINKED ISSUE #449: Return non-zero status code if there are any failures
For the [astropy-benchmarks](https://github.com/astropy/astropy-benchmarks), I would like the Travis status to be a failure if there are any failures during the benchmarks (currently there are three benchmark failures but Travis passes). Maybe a flag could be added to this effect?

mtreinish:
This commit adds a new flag, --strict, to the run (and by extension dev)
and continous commands. When set any benchmark failures encountered
will be treated as an error and the command will exit non-zero when the
run is finished. This is especially useful when running benchmarks in
a CI system or similar automated environment and wish to know if a
change was made to break a benchmark.

Fixes #449

### DIFF
--- a/asv/commands/run.py
+++ b/asv/commands/run.py
@@ -224,7 +224,7 @@ class Run(Command):
             repo.pull()
 
         # Track failures across the run command
-        failures = []
+        failures = False
 
         # Comparison period for date_period filtering
         old_commit_hashes = None
@@ -506,6 +506,7 @@ class Run(Command):
 
                         if not skip_save:
                             result.save(conf.results_dir)
+
                         if strict:
                             failures = failures or any(
                                 code != 0 for code in result.errcode.values())
--- a/asv/commands/run.py
+++ b/asv/commands/run.py
@@ -507,17 +507,15 @@ class Run(Command):
                         if not skip_save:
                             result.save(conf.results_dir)
                         if strict:
-                            failed = [
-                                failed_bench for failed_bench in
-                                result.errcode if
-                                result.errcode[failed_bench] != 0]
-                            failures += failed
+                            failures = failures or any(
+                                code != 0 for code in result.errcode.values())
 
                         if durations > 0:
                             duration_set = Show._get_durations([(machine, result)], benchmark_set)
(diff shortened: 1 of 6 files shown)


asv/commands/run.py
```diff
@@ -499,10 +506,18 @@ def iter_rounds_commits():
...
+                                failed_bench for failed_bench in
+                                result.errcode if
+                                result.errcode[failed_bench] != 0]
+                            failures += failed
 
                         if durations > 0:
                             duration_set = Show._get_durations([(machine, result)], benchmark_set)
                             log.info(cls.format_durations(duration_set[(machine, env.name)], durations))
+        if failures and strict:
+            return len(failures)
```
pv: That'll exit with success if you have 256*n failures (sys.exit takes lowest bits).

Probably should be: `2 if failures else 0` (maybe different error code than default is preferable).

asv/commands/run.py
```diff
@@ -499,10 +506,18 @@ def iter_rounds_commits():
 
                         if not skip_save:
                             result.save(conf.results_dir)
+                        if strict:
+                            failed = [
+                                failed_bench for failed_bench in
+                                result.errcode if
+                                result.errcode[failed_bench] != 0]
+                            failures += failed
```
pv: Maybe better to make `failures` to boolean and do `failures = failures or any(code != 0 for code in result.errcode.values())` -- storing the failed benchmarks in a list is not necessary, so best not do it.

asv/commands/run.py
mtreinish: Fixed in: https://github.com/airspeed-velocity/asv/pull/865/commits/324ba8016bd9a4369a40d35a2b19a703852c9ab2

asv/commands/run.py
mtreinish: Fixed in: https://github.com/airspeed-velocity/asv/pull/865/commits/324ba8016bd9a4369a40d35a2b19a703852c9ab2

pv:
thanks, merged

PR #873, opened 2019-09-05
ENH: step_detect: exclude size-1 intervals

pv:
Make detect_steps to disallow size-1 intervals. Although in principle
they might be significant, in practice it appears outliers of this type
appear quite often and should usually be ignored.

Adds back the manual fine-tuning removed in b43b88f4ee8a366dd6


pv:
Not the right thing, though, as large peaks just get a size-2 intervals with a second unrelated commit.

PR #888, opened 2019-10-22
Fix content overflow in graph tooltips

Alphare:
I'm not 100% sure if the `white-space` property was set to `pre` for a specific reason, but this fixes the very annoying content overflow that happens with a lot of/long variants.

PS: I'm finding time again for ASV, I should be able to get back to my pending PR.


pv:
Thanks, sorry for the delay.

PR #893, opened 2019-11-07
Make summary grid have multiple levels

Alphare:
The current summary grid can get pretty unwieldy when you have a lot of different benchmarks. This is a proposal for making it easier to read by nesting them in a tree-like fashion. 

This assumes that you have a coherent hierarchy of benchmarks: `a.b.c` would be shadowed by `a.b`.


Alphare:
I made it so the output was sorted alphabetically.

pv:
I'd prefer to split to subgroups only if there are more than 5 (or something) benchmarks. The number of benchmarks per class apparently varies between projects, and e.g. the scipy benchmark grid gets very long with this change.

PR #896, opened 2019-11-16
More reasonable data filling behavior in summary graphs

### LINKED ISSUE #894: Summary plot can be misleading
Summary plot may show that performance has degraded:

ex_summary_plot

But this is not the case when you click on the plot to see the actual benchmark data:

ex_expanded_plot

xref https://www.astropy.org/astropy-benchmarks/#modeling.compound.EvaluateCompoundModelWithUnits.time_medium

@astrofrog ~should know which~ said dev `asv` version was used.

pv:
When constructing the averaged data series for the summary graphs,
limit the missing data filling to a fraction of the number of available points,
and don't fill beyond the bounds of available points.

Closes #894

### DIFF
--- a/asv/graph.py
+++ b/asv/graph.py
@@ -6,6 +6,7 @@ from __future__ import (absolute_import, division, print_function,
 
 import os
 import traceback
+import math
 
 import six
 from six.moves import xrange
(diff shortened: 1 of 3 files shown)


pllim:
Looking forward to the improved summary plots. Thank you!

astrofrog:
Seems to do the trick - before:

Screenshot 2019-11-18 at 14 59 19

after:

Screenshot 2019-11-18 at 15 06 31

Thanks!

PR #902, opened 2020-02-01
Save intermediate results from 'asv find', with flag to disable

qwhelan:
Currently, the intermediate binary search steps of `asv find` are never saved, requiring a separate `asv run` to persist the appropriate commit in regression results. This greatly reduces the number of full `asv run`s while also increasing regression precision.

This PR saves intermediate results under the assumption they are comparable to `asv run` and provides a new flag `--skip-save` to opt-out of this behavior.

### DIFF
--- a/asv/commands/find.py
+++ b/asv/commands/find.py
@@ -147,6 +147,9 @@ class Find(Command):
                 env.env_vars
             )
 
+            if not skip_save:
+                result.load_data(conf.results_dir)
+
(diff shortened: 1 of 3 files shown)


mattip:
What do people think of this? Worthwhile? Reverse the default? Does it cause significant file bloat if many results are saved?

HaoZeke (approved):
I think this looks good, makes sense to save intermediates for `asv compare`.

mattip:
Thanks @qwhelan

PR #905, opened 2020-02-11
remove deprecated `--no-site-packages` flag in virtualenv

luizirber:
Hello,

`--no-site-packages` was deprecated some time ago, and removed in `20.x`. I noticed this because it [broke CI][0] for my project.

I can do a better job of checking `virtualenv` versions and keeping the flag for older versions, but just removing it should be fine for most `virtualenv` usages?

[0]: https://github.com/dib-lab/sourmash/runs/439242791#step:5:36


cottrell:
Is there anything blocking merging this?

pv:
Fixed in master now.

PR #917, opened 2020-05-11
Virtualenv no longer supports --no-site-packages; remove it.

DWesl:
I got crashes when trying to use asv with latest virtualenv.

I think `virtualenv >= 20.0.0b1` removed `--no-site-packages`, with no readily apparent replacement.
However, `virtualenv >= 1.7` apparently [has this as the default](https://virtualenv.pypa.io/en/legacy/changes.html#v1-7-2011-11-30), so that may be less needed now than it once was.

It may be worth adding a setuptools `virtualenv` extra to make sure the virtualenv version is recent enough.

Closes #906 
Alternate: #905

PR #919, opened 2020-05-13
virtualenv: remove --no-site-packages option

pared:
Seems like since virtualenv 20 there is no `--no-site-packages` option.


pared:
Just noted its already in PR:
https://github.com/airspeed-velocity/asv/pull/917
closing

Issue #938, opened 2020-07-03
regressions: publication date should be commit date


pared:
I think that `updated` field in RSS field should be tied to commit creation date.

Currently it is run date, so in situation, when for example:
- we change the benchmark runner (eg we have new machine)
- want to rerun all benchmarks
we end up getting updates on regressions from the past, that might have been already handled.

pv:
I guess for this case the relevant part is the content of the `id` field, not as much "updated"?
My intention was that the `id` only depended on the commit range, but that's not true and could be fixed.

PR #939, opened 2020-07-04
Take Atom feed id date from commit date instead of benchmark run

pv:
The Atom feed ids should be 1-to-1 with the commits where the
regressions occurred, so the commit date should be used instead of
benchmark run time in the tag: uris.

See gh-938

### DIFF
--- a/asv/plugins/regressions.py
+++ b/asv/plugins/regressions.py
@@ -207,8 +207,10 @@ class Regressions(OutputPublisher):
                 # as the same one, as long as the benchmark name and
                 # commits match.
                 id_context = [name, revision_to_hash.get(rev1, ""), revision_to_hash.get(rev2, "")]
+                id_rev = rev1 if rev1 is not None else rev2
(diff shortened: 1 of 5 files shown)


asv/plugins/regressions.py
```diff
@@ -207,8 +207,10 @@ def _save_feed(cls, conf, benchmarks, data, revisions, revision_to_hash):
                 # as the same one, as long as the benchmark name and
                 # commits match.
                 id_context = [name, revision_to_hash.get(rev1, ""), revision_to_hash.get(rev2, "")]
+                id_rev = rev1 if rev1 is not None else rev2
```
pared, line 210: Shouldn't id_rev be set, by default to `rev2`?
`rev2` seems to be the latest change, and in case of regression, it is the one "introducing" it.

pv:
Yes, should be `rev2`.

But I'm not sure if this PR solves your problem, since the github widget you used does not seem to keep track of entry ids (like RSS readers should) but instead only looks at the entry update date.

PR #950, opened 2020-11-19
Add id attribute to benchmark group containers to make them linkable.

cmarmo:
This pull request add an id attribute to benchmark group containers.
This make the benchmark group sections linkable with anchors `#<group>`.
I personally find that useful when link to specific sections are needed from a documentation.

Thanks for considering it.

### DIFF
--- a/asv/www/summarygrid.js
+++ b/asv/www/summarygrid.js
@@ -111,7 +111,7 @@ $(document).ready(function() {
         }
 
         $.each(get_benchmarks_by_groups(), function(group, benchmarks) {
-            var group_container = $('<div class="benchmark-group"/>')
+            var group_container = $('<div id="' + group + '" class="benchmark-group"/>')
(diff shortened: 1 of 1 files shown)


asv/www/summarygrid.js
```diff
@@ -111,7 +111,7 @@ $(document).ready(function() {
         }
 
         $.each(get_benchmarks_by_groups(), function(group, benchmarks) {
-            var group_container = $('<div class="benchmark-group"/>')
+            var group_container = $('<div id="' + group + '" class="benchmark-group"/>')
```
pv, line 114:
```suggestion
            var group_container = $('<div class="benchmark-group"/>')
            group_container.attr('id', group)
```

pv:
Merged, sorry for the delay. Note however that they're probably still not directly linkable, because the hash string is used for navigation in the JS application.

PR #953, opened 2020-12-21
make benchmark template work with Python 3

saraedum:
afaik, xrange and iterkeys do not exist anymore in Python 3.


pv:
Thanks, merged.

PR #954, opened 2020-12-21
Fix JavaScript navigation for method names with escaped entities

saraedum:
In Python 3, a method such as `def time_β(): pass` is valid. However,
asv's HTML output struggles with characters in benchmark names that get
URL encoded. The following shows up in the browser console for such
a method:

```
Uncaught TypeError: index.benchmarks[current_benchmark] is undefined
```

The problem here seems to be that `current_benchmark` is not correctly
decoded and contains the URL encoding of `β`.


pv:
Thanks, merged.

Issue #544, opened 2017-09-19
Run asv for single commit


astrofrog:
I am trying to run asv for a single commit but it seems to pick up all commits since the start of the project:

```
$ asv run --bench CsvFloat c2ef85974c13cbba5dc72df72607ff284cffc6d5
· Creating environments
· Discovering benchmarks
·· Uninstalling from conda-py2.7-Cython-jinja2-nomkl-numpy1.12-scipy
·· Building for conda-py2.7-Cython-jinja2-nomkl-numpy1.12-scipy..............
·· Installing into conda-py2.7-Cython-jinja2-nomkl-numpy1.12-scipy..
· Running 7996 total benchmarks (3998 commits * 1 environments * 2 benchmarks)
[  0.00%] · For astropy commit hash c2ef8597:
```

Is there a way to run single commits?

pv:
Same as for "git log":
c2ef85974c13cbba5dc72df72607ff284cffc6d5^!

astrofrog:
I wonder if we could add that example to the ``asv run`` help? This is some git-fu that not everyone will necessarily know.

pv:
Yes, adding example somewhere, documentation, and maybe add in `asv run --help` something like following:

```
Run a benchmark suite.

examples:
  asv run master^!      run benchmarks for one commit (in git repository)
```

PR #968, opened 2021-06-16
example to run single commit

erwanp:
I couldn't find how to benchmark a specific tag or commit. 

This Issue gave the answer, https://github.com/airspeed-velocity/asv/issues/544#issuecomment-330463218, but the information is only shown in the `git run` command-line help and not on the website. 

I added one example in the "Using" section.


quasiben:
Looking at the docs today I don't see this line from this commit:
https://asv.readthedocs.io/en/stable/using.html

Does RTD need to be refreshed ?

PR #981, opened 2021-12-11
Remove unused sys.exit

datapythonista:
This `sys.exit` call is never executed, and adds no value, since `main` already calls `sys.exit` and won't return.


mattip:
seems like a reasonable cleanup. Any other thoughts here?

alimcmaster1 (approved):
Agree - main always calls sys.exit

mattip:
Thanks @datapythonista

PR #986, opened 2022-01-24
Modify files to follow PEP8/Flake8 guildlines

dorothykiz1:
Currently, most asv files are not following Flake8 guidelines this PR  
fixes Maximum line length to 79 characters 
and also removes the redundant variables and imports


asv/benchmarks.py
```diff
@@ -182,12 +184,11 @@ def iter_unique(iter):
         log.info("Discovering benchmarks")
         with log.indent():
             last_err = None
-            for env, commit_hash in itertools.product(environments, try_hashes):
+            tools = itertools.product(environments, try_hashes)
+            for env, commit_hash in tools:
```
datapythonista, line 188: Probably easier, and to me at lease, more readable, to use this:

```suggestion
            for env, commit_hash in itertools.product(environments,
                                                      try_hashes):
```

asv/benchmarks.py
```diff
@@ -67,10 +65,13 @@ def __init__(self, conf, benchmarks, regex=None):
                         ', '.join(param_set))
                     if not regex or any(re.search(reg, name) for reg in regex):
                         self[benchmark['name']] = benchmark
-                        self._benchmark_selection[benchmark['name']].append(idx)
+                        bms = self._benchmark_selection[benchmark['name']]
+                        bms.append(idx)
```
datapythonista, line 69: I'd do. Particularly in this case, you're creating an alias of this list, which makes the code significantly harder to read IMHO.

```suggestion
                        (self._benchmark_selection[benchmark['name']]
                             .append(idx))
```

asv/feed.py
```diff
@@ -43,7 +44,9 @@ class FeedEntry(object):
         Default: same as *updated*
 
     """
-    def __init__(self, title, updated, link=None, content=None, id_context=None, id_date=None):
+    def __init__(
+                    self, title, updated, link=None, content=None,
+                    id_context=None, id_date=None):
```
datapythonista, line 49: I don't think this is valid PEP8 code, and even if it is, this indentation seems quite arbitrary. Just split when the line is too long.

datapythonista (changes requested):
Thanks @dorothykiz1, nice job. Besides the inline comments, there are couple of things we should fo before merging here:
- We should discuss what's the maximum line length we want to allow in this repo. I'd say 99 is better than the 79 you're enforcing here. Some of the changes here to satisfy the 79 columns constraint seem to make the code less clear. Can you open an issue to discuss please?
- You should add validation for flake8 for the files you're fixing in the CI here. So we can see this is working as expected, and also we make sure no more PEP8 violations are introduced here after your changes.
- We should consider whether we want to drop Python 2 support before these changes. You've got couple of places in this PR where we could use f-strings to make the code cleaner. We should first make sure there are no objections to it (there is an issue about dropping py2 support, you can ask there). Then, we should remove any py2 build from the CI. And after that we can start writing py3 only code.

asv/benchmarks.py
dorothykiz1: Totally agree from your point of view

datapythonista: This is pending. Note that you can change this manually in your local copy and commit. But you can also change it from the GitHub UI, in the commit dropdown to "Commit suggestion".

asv/.github/workflows/flake8.yml
```diff
@@ -0,0 +1,24 @@
...
+name: flake8
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10"]
```
datapythonista, line 11: This looks good, but for linting probably linting with the latest python version is enough. When linting the asv code is not executed, so we're just saying here to run flake8 with those different versions of python, which doesn't add a lot of value. For testing this would make total sense.

datapythonista (commented):
Looks good. There is just one thing. The CI you're implementing here won't run until this PR is merged. What would be better is to add the github CI file in a separate PR, without running flake8 (since it'll fail), and then run flake8 in the CI here, so we can see if the CI is green with your changes here (I think it won't yet, but better to discuss when we can see the errors in the CI).

asv/.github/workflows/flake8.yml
dorothykiz1: Great analogy, in case of feature new versions, don't have to come back to edit it.

dorothykiz1:
To be revised first

PR #989, opened 2022-01-26
Add flake8 CI configuration

dorothykiz1:
To add validation for flake8 in the CI for the files on push or pull requests on this repository 
following  this guide https://github.com/marketplace/actions/python-flake8-lint


datapythonista (changes requested):
Thanks for working on this @dorothykiz1.

Looks great, there is just one detail. I don't think our repo is passing flake8 at the moment, so this will make the CI for master and all PRs red. We should add this, but not failing the build. I guess instead of the action you will have to run flake8 manually, and use something like `flake8 . | true`. Once CI is passing this should be fine as is.

dorothykiz1:
Thank You @datapythonista  for your review, do you reckon I remove "on" the action.

datapythonista (changes requested):
Thank you @dorothykiz1, did you check if this works?

If you didn't, let's start with what we first discussed, and add a simple hello world CI. The thing is that since the GitHub Actions workflow file is not yet committed, this CI is not running, and we can't know if this will start failing.

For the flake8 action, there are few things that don't make so much sense. Like using `uses` and `run` together. Or what is the path `src` referring to. Also, having the configuration in this yaml file for what to ignore, and the max line length instead of in the `setup.cfg` we discussed, means that when running flake8 locally, the default settings will be used, and the errors reported will differ in local environments and in the CI.

So, unless I'm missing something, let's start with a CI that simply prints a message in the logs, let's merge that quickly, and then add flake8 making sure things work as expected.

LucyJimenez:
@dorothykiz1 & @datapythonista 

Since we already have a first version of the CI running, do you think we still need this PR?

datapythonista:
This has already been implemented in other PRs, closing.

PR #988, opened 2022-01-25
Fix PEP-8 violations reported by flake8 on asv folder

LucyJimenez:
Check flake8 and fix issues pep8 related on asv/asv folder for:

1. statistics.py
2. runner.py
3. results.py
4. repo.py
5. main.py

PR #995, opened 2022-01-27
Fix Flake8 violations

dorothykiz1:
To fix Flake8 violations for ;
1. asv/__init__.py
2. asv/config.py


asv/__init__.py
```diff
@@ -4,6 +4,7 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
+
```
datapythonista, line 7: I don't think this is required by flake8, is it?

dorothykiz1: No, it's not, this has been fixed. Thank You, Marc

asv/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
-
+
```
datapythonista, line 6: I think flake8 should be complaining about this. Are you running flake8 at all to check your changes?

datapythonista:
CI is red. You also need to merge master into your branch, and run flake8 in the CI. You can use #988 as an example.

Same for the rest of PRs.

dorothykiz1:
To be revised first

PR #996, opened 2022-01-27
Fix Flake8 violations

dorothykiz1:
To fix Flake8 violations for ;
1. asv/console.py
2. asv/environment.py
3. asv/graph.py


dorothykiz1:
To be revised

dorothykiz1:
To be revised first

PR #1002, opened 2022-01-30
Fix pypy version environment test on latest pypy

QuLogic:
It appears to print out as a named tuple


datapythonista:
Thanks for the contribution @QuLogic. I'm happy to merge this, but I fail to understand where this is coming from. Is the affected test failing if we run tests with pypy? We're in the process of moving the tests to GitHub Actions, do you think it'd make sense to run them on pypy there?

QuLogic:
This test is about pypy itself.

datapythonista:
Superseded by #1111

PR #1004, opened 2022-01-31
Fix flake8 violations

dorothykiz1:
resolve flake 8 violations for;

- test/test_repo_template/asv_test_repo/__init__.py
- test/conftest.py
- test/test_check.py
- test/test_benchmarks.py


dorothykiz1:
To be revised first

PR #1007, opened 2022-02-02
Fix PEP-8 violations reported by flake8 on asv/util.py

LucyJimenez:
closes https://github.com/pandas-benchmarks-grant/board-issues/issues/6
Fix pep8 issues on util.py


asv/util.py
```diff
@@ -531,7 +531,7 @@ def get_content(header=None):
     if env and WIN and sys.version_info < (3,):
         # Environment keys and values cannot be unicode
         def _fix_env(s):
-            return s.encode('mbcs') if isinstance(s, unicode) else s
+            return s.encode('mbcs') if isinstance(s, unicode) else s  # noqa F821 undefined name 'unicode'
```
datapythonista, line 534:
```suggestion
            return s.encode('mbcs') if isinstance(s, unicode) else s  # noqa F821 'unicode' does not exist in py2
```

asv/util.py
```diff
@@ -566,7 +566,7 @@ def _fix_env(s):
         debug_log = DebugLogBuffer(log)
         dots = False
     else:
-        debug_log = lambda c: None
+        debug_log = lambda c: None # noqa
```
datapythonista, line 569: We shouldn't have bare `noqa`, we should always specify the error we want to ignore, and why.

In this case, why we don't fix the error properly?

asv/util.py
```diff
@@ -730,7 +729,7 @@ def sig_forward(signum, frame):
 
     # Flush and disconnect debug log, if any
     debug_log(None)
-    debug_log = lambda c: None
+    debug_log = lambda c: None  # noqa
```
datapythonista, line 732: same

asv/util.py
```diff
@@ -81,28 +81,28 @@ def reraise(self):
             raise self
 
 
-def human_list(l):
+def human_list(non_human_list):
```
datapythonista, line 84: This is a normal list that wants to be printed in a human readable way. `non_human_list` doesn't seem to be a very clear name of what it is, maybe better just `input` or `input_list` or even `list_to_format`.

datapythonista (commented):
Looks good, thanks for the fixes. Added couple of comments, but nice contribution.

asv/util.py
LucyJimenez: You are right. I was checking the other changes first, and I'm going to fix them.

datapythonista (approved):
Thanks @LucyJimenez, looks great.

datapythonista:
Can you rebase please?

datapythonista:
Can you fix the conflicts please?

datapythonista:
Thanks @LucyJimenez, really nice!

PR #1019, opened 2022-02-07
Fix all flake8 violations for asv/benchmark.py

dorothykiz1:


asv/benchmark.py
```diff
@@ -41,7 +41,7 @@
 import copy
 try:
     import cProfile as profile
-except:
+except Exception:
```
datapythonista, line 44: We shouldn't capture all exceptions here, just with `ImportError` is enough.

asv/benchmark.py
```diff
@@ -574,7 +568,7 @@ def do_run(self):
 
     def do_profile(self, filename=None):
         def method_caller():
-            run(*params)
+            run(*params) # noqa see #1020
```
datapythonista, line 571: When using `noqa` we need to specify the exact error code we want to ignore. Also, if besides pointing to the issue you can write a very short sentence on what's the problem  and why this should be skipped, that would be useful

asv/benchmark.py
dorothykiz1: let me add this too

dorothykiz1:

datapythonista (commented):
Lools good, thanks for the work on this @dorothykiz1. Are all the problems in this file fixed? Can we validate it in the CI, so we know the changes here are correct, and future changes to this file don't introduce more problems please?

dorothykiz1:
@datapythonista thanks for your review let me  merge this with the other PR and add it the CI

Issue #1022, opened 2022-02-07
Changes required to set minimum Python version to 3.7


datapythonista:
xref #909

We are removing  support for old versions of Python, and cleaning the code. We'll be moving the CI from Appveyor to GitHub actions (see #985), and running the tests with the minimum and maximum supported Python version. Besides that, we should be updating the required Python version anywhere needed. This includes:
- `setup.py`
- The documentation. If we don't mention the minimum Python version already, we should add it, at least in the installation page.
- Requirements files if a Python version is specified (like `requirements-dev.txt`, `environment.yml`)
- Anywhere else relevant

If there are no objections, the minimum Python version supported by asv will be Python 3.7, the oldest officially supported.

@LucyJimenez do you mind taking care of this please?

kanderso-nrel:
Hello, asv user here!  I'm happy to see the recent maintenance activity for this package and excited about the plans for future development.

> If there are no objections, the minimum Python version supported by asv will be Python 3.8, the oldest officially supported. If that's a problem for anyone, I think 3.8 or 3.7 could also be fine.

Not necessarily a problem, but I'm curious about the motivation to drop support for 3.7, which as I understand still has over a year until EOL: https://devguide.python.org/#status-of-python-branches

LucyJimenez:
Yes, I'm going to start working on this.

LucyJimenez:
@datapythonista  This task was done, was closed with https://github.com/airspeed-velocity/asv/pull/1025
We can close this issue.

PR #1025, opened 2022-02-09
Changes required to set minimum Python version to 3.7

LucyJimenez:
xref #1022 

Update the required Python version on `setup.py` and some parts of the documentation.

@datapythonista do you think we need to update the examples on the exclude and include part? https://asv.readthedocs.io/en/stable/asv.conf.json.html

### DIFF
--- a/docs/source/installing.rst
+++ b/docs/source/installing.rst
@@ -2,7 +2,7 @@ Installing airspeed velocity
 ============================
 
 **airspeed velocity** is known to work on Linux, Mac OS-X, and Windows.
-It is known to work with Python 3.7.
+It is known to work with Python 3.7 and higher.
 It works also with PyPy.
 
 **airspeed velocity** is a standard Python package, and the latest
--- a/docs/source/installing.rst
+++ b/docs/source/installing.rst
@@ -2,7 +2,7 @@ Installing airspeed velocity
 ============================
 
 **airspeed velocity** is known to work on Linux, Mac OS-X, and Windows.
-It is known to work with Python 2.7, 3.4, 3.5, and 3.6.
+It is known to work with Python 3.7.
 It works also with PyPy.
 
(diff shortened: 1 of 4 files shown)


docs/source/installing.rst
```diff
@@ -2,7 +2,7 @@ Installing airspeed velocity
 ============================
 
 **airspeed velocity** is known to work on Linux, Mac OS-X, and Windows.
-It is known to work with Python 2.7, 3.4, 3.5, and 3.6.
+It is known to work with Python 3.7.
```
datapythonista, line 5: do you mind saying `with Python 3.7 and higher.` please?

docs/source/using.rst
```diff
@@ -214,11 +214,11 @@ installed to support other environment tools.  The
 tool used to create environments.
 
 When using ``virtualenv``, ``asv`` does not build Python interpreters
-for you, but it expects to find each of the Python versions specified
+for you, but it expects to find the Python versions specified
 in the ``asv.conf.json`` file available on the ``PATH``.  For example,
 if the ``asv.conf.json`` file has::
 
-  "pythons": ["2.7", "3.6"]
+  "pythons": ["3.7"]
```
datapythonista, line 221: I think the example is clearer if you use 2 versions, maybe 3.7 and 3.10.

datapythonista (commented):
Thanks for the changes @LucyJimenez 

I think updating the versions in the asv.conf.json page makes sense. IMO it looks better not using versions so much in the past, and maybe a bit misleading to show unsupported versions. But up to you.

LucyJimenez:
> I think updating the versions in the asv.conf.json page makes sense. IMO it looks better not using versions so much in the past, and maybe a bit misleading to show unsupported versions. But up to you.

Perfect, I'll do it!

datapythonista (approved):
Thanks for taking care of this @LucyJimenez, looks great!

PR #1029, opened 2022-02-14
Fix flake8 violations

dorothykiz1:
Closes [15](https://github.com/pandas-benchmarks-grant/board-issues/issues/15)

Fix flake8 violations in files below;

- test/test_statistics.py
- test/test_quickstart.py
- test/tools.py


test/tools.py
```diff
@@ -80,12 +82,11 @@ def _check_conda():
     # Conda can install required Python versions on demand
     _check_conda()
     HAS_CONDA = True
-except (RuntimeError, IOError) as exc:
+except (RuntimeError, IOError):
     HAS_CONDA = False
 
 
 try:
-    import virtualenv
```
datapythonista, line 88: This will change the behavior of the program, we need to revert.

test/test_results.py
```diff
@@ -14,8 +14,6 @@
 from asv import results, runner, util
 import pytest
 
-from .tools import example_results
```
datapythonista, line 17: Looks like this is needes to load the fixture, that's one of the failures in the CI.

test/tools.py
dorothykiz1: This was failing with a flake8 error imported but not used

dorothykiz1:

test/test_results.py
```diff
@@ -14,7 +14,7 @@
...
+from .tools import example_results # noqa :F401 imported but unused
```
datapythonista, line 17: For the comment, we don't want the flake8 error, which doesn't add much value. What we want is to explain why we are ignoring the flake8 error. Reading the comment, the obvious thing to think is that this import should be deleted, since it's imported but unused. A helpful comment would be to say that this is required even if unused, so the example_results fixture is loaded.

For this particular case, what it may make more sense is to move the fixtures from `tools.py` to the standard `conftest.py`. This way this import won't be needed, as pytest will take care of everything. We don't want to have such a change in a PR with minor style changes. So, one option would be to:
- Change this line to something like `from . import tools  # noqa F401 needed to load fixtures (see #1234)`
- Remove the other `noqa` because of redefinition of `example_results`, which shouldn't be needed anymore
- Open an issue to explain what's going on here, to move the fixtures to the right location

test/tools.py
```diff
@@ -80,12 +82,12 @@ def _check_conda():
...
+    import virtualenv # noqa :F401 imported but unused
```
datapythonista, line 90: Same, this comment doesn't add any value, what we want is to explain why we keep this import even if it's unused, not just the flake8 error message, which doesn't provide any value to the readers of this

test/test_results.py
dorothykiz1: Thank you @datapythonista
for your review let me incorporate these changes

test/test_results.py
```diff
@@ -14,7 +14,7 @@
 from asv import results, runner, util
 import pytest
 
-from .tools import example_results
+from .tools import example_results # noqa F401 needed to load fixtures (see #1030)
```
datapythonista, line 17: `from . import tools` should fix the CI

test/test_results.py
dorothykiz1: Hello @datapythonista this still happens when i remove the # noqa

```
test/test_results.py:165:31: F811 redefinition of unused 'example_results' from line 17
test/test_results.py:208:39: F811 redefinition of unused 'example_results' from line 17
```

datapythonista: I left a new review explaining how I would fix this. You need the noqa here with the proposed import, but not the errors you have in your comment I think.

test/test_results.py
```diff
@@ -162,7 +161,7 @@ def test_get_result_hash_from_prefix(tmpdir):
     assert 'one of multiple commits' in str(excinfo.value)
 
 
-def test_backward_compat_load(example_results):
+def test_backward_compat_load(example_results): # noqa uses fixture
```
datapythonista, line 164: Can you ignore only the error code we have now.

test/test_results.py
```diff
@@ -205,7 +204,7 @@ def test_json_timestamp(tmpdir):
     assert values['duration'] == duration
 
 
-def test_iter_results(capsys, tmpdir, example_results):
+def test_iter_results(capsys, tmpdir, example_results): # noqa uses fixture
```
datapythonista, line 207: same

test/test_results.py
```diff
@@ -161,7 +161,7 @@ def test_get_result_hash_from_prefix(tmpdir):
     assert 'one of multiple commits' in str(excinfo.value)
 
 
-def test_backward_compat_load(example_results): # noqa uses fixture
+def test_backward_compat_load(example_results): # noqa F811
```
datapythonista, line 164: Can you add the comment on why we are ignoring this error please

test/test_results.py
```diff
@@ -204,7 +204,7 @@ def test_json_timestamp(tmpdir):
     assert values['duration'] == duration
 
 
-def test_iter_results(capsys, tmpdir, example_results): # noqa uses fixture
+def test_iter_results(capsys, tmpdir, example_results): # noqa F811
```
datapythonista, line 207: same

datapythonista:
Thanks @dorothykiz1, very nice

PR #1032, opened 2022-02-16
Add pytest to CI

LucyJimenez:
Add pytest to `ci.yml`.


LucyJimenez:
WIP!

LucyJimenez:
@datapythonista Could you please take a look at this?

datapythonista (approved):
Thanks @LucyJimenez, looks good. There are few improvements that can be made, like adding better logging to pytest, run in a matrix of the min and latest python versions, run in pypy, add flake8 to the dependencies, add a job to build the docs... But we can do those in follow up PRs, and keep things simple at every step.

PR #1033, opened 2022-02-16
Run pytest with different Python versions in GitHub actions

LucyJimenez:
xref: https://github.com/airspeed-velocity/asv/issues/1036
On this PR:
- Add matrix python versions
- Separate test and flake8 on different builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,3 +1,5 @@
+
+   
 name: CI
 
 on: [push, pull_request]
@@ -6,11 +8,21 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
+
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout source
+        uses: actions/checkout@v2
 
       - name: Install dependencies
-        run: pip install flake8
+        run: |
+          pip install -r requirements-dev.txt
+          pip install flake8
       
       - name: flake8
-        run: flake8 . 
+        run: flake8 .
+      
+      - name: Run tests
+        run: pytest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,3 @@
-
-   
 name: CI
 
 on: [push, pull_request]
@@ -10,19 +8,22 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.10"]
 
     steps:
       - name: Checkout source
         uses: actions/checkout@v2
 
       - name: Install dependencies
-        run: |
-          pip install -r requirements-dev.txt
-          pip install flake8
-      
-      - name: flake8
-        run: flake8 .
+        run: pip install -r requirements-dev.txt
       
       - name: Run tests
         run: pytest
+
+  flake8:
+    steps:
+      - name: Install dependencies
+        run: pip install flake8
+      
+      - name: flake8
+        run: flake8 .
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,3 +1,5 @@
+
+
```
datapythonista, line 2: Better to remove those.

.github/workflows/ci.yml
```diff
@@ -6,6 +8,10 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
```
datapythonista, line 13: Just the first and the last versions should be enough. Too many builds add noise and consume resources. Or we may want to add something else to the matrix (like CPython/PyPy), and then the matrix will explode (like 8 builds just for that change.

.github/workflows/ci.yml
```diff
@@ -19,4 +25,4 @@ jobs:
         run: flake8 .
```
datapythonista, line 25: I think it's probably better to have flake8 and pytest in separate builds. Doesn't make so much sense to test flake8 in different Python versions.

datapythonista (commented):
Thanks for working on this. Looks good, added some minor things.

.github/workflows/ci.yml
LucyJimenez: Thank you for the recommendations, I’m going to implement it.

.github/workflows/ci.yml
```diff
@@ -3,20 +3,25 @@ name: CI
 on: [push, pull_request]
 
 jobs:
-  build:
+  test:
+    name: test w/ Python
```
datapythonista, line 7: Just tests should be enough. To show that it's not pypy we should say CPython, but pypy is likely to be in the matrix, so better to just name this tests

.github/workflows/ci.yml
```diff
@@ -3,20 +3,25 @@ name: CI
...
     runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.10"]
```
datapythonista, line 11: Did you check that tests are being run with these Python versions?

.github/workflows/ci.yml
```diff
@@ -3,20 +3,25 @@ name: CI
...
-      
-      - name: flake8
-        run: flake8 .
-      
+        run: pip install -r requirements-dev.txt
       - name: Run tests
         run: pytest
+  
+  flake8-formatting:
+    name: code autoformatting with flake8
```
datapythonista, line 20: This is misleading. We are not formatting or autoformatting anything. Linting would be the generic name for checking code styles, flake8 is the tool fot checking them.

datapythonista (commented):
Thanks @LucyJimenez, added some comments.

.github/workflows/ci.yml
LucyJimenez: Yes, there is just one warning:
image

image

.github/workflows/ci.yml
LucyJimenez: Great, thanks for the explanation :)

.github/workflows/ci.yml
LucyJimenez: Thank you for pointing out that problem. Now the CI is working with Python 3.7 and 3.10. There is no warning reported on Python 3.7, and [two warnings](https://github.com/airspeed-velocity/asv/runs/5259920032?check_suite_focus=true) on Python 3.10. I'm going to check it in another PR.

image

image

LucyJimenez:
Additional data that could be interesting about test:

- `CI Py2.7` 201 passed, 15 skipped in 466.88s (0:07:46)
- `CI Py3.10` 201 passed, 15 skipped, 2 warnings in 551.07s (0:09:11)
- `Appveyor` 168 passed, 48 skipped in 1259.22s (0:20:59)

datapythonista:
Thanks @LucyJimenez, really nice

PR #1049, opened 2022-02-23
CLN: use shutil terminal size

fangchenli:


datapythonista (approved):
Thanks @fangchenli, looks great. Do you thinl this is worth a release note? I guess there is no user impact.

fangchenli:
> Do you thinl this is worth a release note? I guess there is no user impact.

It's probably not worth it.

datapythonista:
Thanks @fangchenli, nice clean up.

PR #1054, opened 2022-02-26
Remove imports from __future__ module4

mariamadronah:


datapythonista:
Thanks for the contribution @mariamadronah. CI is red. The problem is that when you remove the import, you should just remove the line, and not add a blank line in its place. If you check the diff in GitHub, should be quite easy to see what I mean (there shouldn't be empty green lines in the diff next to the red of the removals. Do you mind having a look and fixing those, so we can merge please? Feel free to ping me when CI is green, or if you have any problem.

datapythonista:
Btw, since you're editing those files to remove the `__futures__`, maybe you also want to remove the `# -*- coding: utf-8 -*-` which also can be removed now that the code is Python 3 only. I guess it should be almost the same effort to remove one or both.

datapythonista:
You've got almost all there fixed now, but still two more failures. Let me know if you need help finding what are the errors in the CI. Also, no big difference, but even if flake8 is happy with 2 blank lines, probably having just one is less noisy. No need to change the files you already finish, just letting you know in case you fix new ones.

mariamadronah:
have worked on the  errors in the CI , let me get this # -*- coding: utf-8 -*- removed too. thanks for the help rendered

datapythonista:
> yes please help me find the errors in C1 because i guess am not seeing that

In this `Conversation` tab, at the end you've got the results of the checks. That's what we usually name CI, and it can include things like checking the styles of the code, running unit tests, checking that the documentation builds without errors and similar things.

Now you can see it says `All checks have passed`, and shows in green. When something is not right, based on the checks implemented for the project, it'd show in red. There is a list of the checks, at least one will be in red. If you click on the details of the red job (or one of them if more than one), you'll be taken to another page named `Actions`. There you'll see the job that failed, with all its steps, showing one in red. If you expand that step, you should have the command that was being used for the test (in this case was `flake8`) and the log, which should include what went wrong. If you get anything else in red, you can have a look, and let me know if anything is still unclear.

asv/plugins/snakeviz.py
```diff
@@ -1,10 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 # -*- coding: utf-8 -*-
 
-from __future__ import absolute_import, division, unicode_literals, print_function
 
-from __future__ import (absolute_import, division, print_function,
-                        unicode_literals)
```
datapythonista, line 6: Here we have 4 blank lines in a row after the change. The CI is not failing, because at the moment we're skipping some files like this one from checking the styles. That's alright for this PR, I'm going to merge it, but if you work on new ones, we could remove the line with `# -*- coding: utf-8 -*-` and all the blank lines. So, the beginning of the file would be:
```python
from .. import profiling
...
```

That would be the best in my opinion, as developers using this file would avoid the extra noise of the blank lines. When there are imports they useful to group the imports in blocks, but here they won't bring much value.

datapythonista (approved):
Thanks @mariamadronah, really nice, keep them coming

Issue #1037, opened 2022-02-19
Remove PY2 encoding declaration


dorothykiz1:
We are trying to drop any python 2 related syntaxes, comments, or code and replace this with python 3 equivalent code.

Most files in the asv directory have the first-line comment on top of the file `# -*- coding: utf-8 -*-`

The idea is to remove this line from all files as this encoding declaration is related to python 2

LucyJimenez:
As example you can see the PR: https://github.com/airspeed-velocity/asv/pull/1048/files

Tushcathy:
take

datapythonista:
Same here, looks like all them are now gone. Thanks a lot @Tushcathy and @mariamadronah

PR #1055, opened 2022-02-26
Remove py2 encoding declaration

Tushcathy:
xref #1037


datapythonista:
Thanks @Tushcathy, very nice

PR #1058, opened 2022-02-27
Remove six library

Leenkiz:


datapythonista:
Thanks for the contribution @Leenkiz. Looks like you've got conflicts, can you please merge master into your branch, resolve the conflicts, and push..  again. We need this to be able to run the CI checks and merge. Thanks!

dorothykiz1:
Thanks for this @Leenkiz the conflicts mean that the upstream has new changes that your local copy doesn't have so you need to run **git fetch upstream** and 
**git merge upstream/master** then work on your changes and push.

dorothykiz1:
Thanks @Leenkiz ,one more change,

import socketserver will solve your issue as socketserver is used but not defined

datapythonista:
@dorothykiz1 @LucyJimenez, looks like there are test failures in appveyor that are not showing in GitHub actions. We should have a look and see why is this happening. Do you mind creating an issue, linking to both the appveyor and GitHub actions builds in this PR (not the PR itself, since it'll get lost when the test is fixed), and then have a look and find what's the problem.

In the CI logs you should find the commands that have been executed in both systems. You should be able to implement the bug causing the test failure locally, run both commands, and see one failing and the other passing.

dorothykiz1:
> @dorothykiz1 @LucyJimenez, looks like there are test failures in appveyor that are not showing in GitHub actions. We should have a look and see why is this happening. Do you mind creating an issue, linking to both the appveyor and GitHub actions builds in this PR (not the PR itself, since it'll get lost when the test is fixed), and then have a look and find what's the problem.
> 
> In the CI logs you should find the commands that have been executed in both systems. You should be able to implement the bug causing the test failure locally, run both commands, and see one failing and the other passing.

I have created issue #1060 to reference this

asv/commands/preview.py
```diff
@@ -84,9 +83,9 @@ def run_from_conf_args(cls, conf, args):
     def run(cls, conf, port=0, browser=False):
         os.chdir(conf.html_dir)
 
-        class Handler(SimpleHTTPServer.SimpleHTTPRequestHandler):
+        class Handler(http.server.SimpleHTTPRequestHandler):
```
datapythonista, line 86: Looks like `http.server` is not defined. We should import it.

@dorothykiz1 looks like we don't have tests for preview. It'd be good to create an issue for it. And also, if you want to have a look at other commands that are not tested, that would be very helpful. Not having any test for a command means that like here, even a bug like this one can be added unnoticed (this would fail as soon as the command is executed, for all cases).

dorothykiz1: Thanks @datapythonista i will take a look at this

datapythonista:
Thanks @Leenkiz, nice clean up. I reverted the part that was tricky, so we could merge the other changes. But feel free to continue in a new PR if you want.

PR #1059, opened 2022-02-28
Contribution7

mariamadronah:


asv/repo.py
```diff
@@ -1,12 +1,9 @@
-# -*- coding: utf-8 -*-
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-
 import datetime
 
 from . import util
 
-
```
datapythonista, line 9: This one is needed. In short, the standard is that classes and functions should have two blank lines to separate them, the rest should be separated maximum by one. This is the full standard: https://www.python.org/dev/peps/pep-0008/

test/conftest.py
```diff
@@ -1,7 +1,6 @@
 import os
 import contextlib
 
-
```
datapythonista, line 4: This is the other one also needed.

datapythonista (commented):
Thanks @mariamadronah, really nice clean up. There are just couple of blank lines that need to be restored, other than that this looks perfect.

mariamadronah:
thanks a lot @datapythonista for the help through out me solving this issue its been great learning

PR #1067, opened 2022-03-01
Fix flake8 errors in some files in asv/ directory

dorothykiz1:
This PR fixes all flake8 errors in ;

- asv/template/benchmarks/benchmarks.py 
- asv/template/benchmarks/__init__.py 
- asv/runner.py 
- asv/results.py 
- asv/machine.py 
- asv/graph.py 
- asv/environment.py 
- asv/benchmarks.py


datapythonista (approved):
Thanks @dorothykiz1, really nice. Will merge on green.

datapythonista:
Thanks @dorothykiz1, something that we could do now that flake8 errors are in 4 main directories, is to update `setup.cfg` to exclude those. Not sure about test, how much it's being validated there, but for the rest it can make sense. So, when an issue is closed, is quick to just delete one line from the excludes.

PR #1068, opened 2022-03-01
Remove six

Leenkiz:


asv/commands/quickstart.py
```diff
@@ -3,7 +3,7 @@
 import os
 import shutil
 
-from six.moves import input as raw_input
+import input as input
```
datapythonista, line 6: Thanks  @Leenkiz. `import` is what in Python is named a built-in, a function that is always available (no need to import it). So, you can just remove the import. Also, `import input as input` is redundant, `import input` will already import it as `input`.

datapythonista:
Also, in general what you would do, is every time you want to work on something new, you `git checkout master`, you update it (`git fetch upstream && git merge upstream/master` if you have things properly set up), and then you create a new branch.

This way, even before this is merged, you can do it, and start working in something else. Otherwise, you can just work in one thing at a time, which is not great, while waiting for a PR to be merged.

asv/commands/quickstart.py
Leenkiz: > Also, in general what you would do, is every time you want to work on something new, you `git checkout master`, you update it (`git fetch upstream && git merge upstream/master` if you have things properly set up), and then you create a new branch.
> 
> This way, even before this is merged, you can do it, and start working in something else. Otherwise, you can just work in one thing at a time, which is not great, while waiting for a PR to be merged.

Noted!! Many thanks

datapythonista:
Closing in favor of #1083

Issue #1040, opened 2022-02-19
Remove all six functionality


dorothykiz1:
[`six`](https://six.readthedocs.io/) provides simple utilities for wrapping over differences between Python 2 and Python 3.

Follow this guide  `https://six.readthedocs.io/`to find python 3 equivalent for python 2 six functionality.

Remove all six imports and functionality and replace them with python 3 equivalent.
ie. from `six.xrange()` to `range()`

Leenkiz:
#take

datapythonista:
Those are all the uses of `six` that we still got pending:
```
asv/step_detect.py:import six
asv/graph.py:from six.moves import xrange
asv/util.py:import six
asv/util.py:from six.moves import xrange
asv/plugins/regressions.py:from six.moves.urllib.parse import urlencode
asv/commands/preview.py:from six.moves import SimpleHTTPServer, socketserver
asv/commands/quickstart.py:from six.moves import input as raw_input
asv/results.py:from six.moves import zip as izip
test/test_workflow.py:import six
test/test_console.py:import six
test/tools.py:import six
test/tools.py:from six.moves import SimpleHTTPServer
test/test_web.py:from six.moves.urllib.parse import parse_qs, splitquery, splittag
test/test_results.py:import six
```

PR #1079, opened 2022-03-02
RemovePY2 six from asv/commands/quickstart.py and test/test_console.py files

dorothykiz1:
xref #1040

This will remove six functionality in the following files;

-     asv/commands/quickstart.py
-     test/test_console.py

### DIFF
--- a/test/test_console.py
+++ b/test/test_console.py
@@ -1,6 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-import six
 import io
 import os
 import sys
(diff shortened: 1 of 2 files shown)


test/test_console.py
```diff
@@ -21,53 +20,40 @@ def check_write(value, expected, stream_encoding, preferred_encoding):
...
-                    got = stream.read()
-                    assert got == expected
+            with io.open(fn, 'w', encoding=stream_encoding) as stream:
+                _write_with_fallback(value, stream)
+            with open(fn, 'rb') as stream:
+                got = stream.read()
+                assert got == expected
 
             # Check writing to Py2 files
-            if not six.PY3:
```
datapythonista, line 62: This is an `if not`. We're always `PY3`, so this should just be removed.

datapythonista:
Thanks @dorothykiz1

Issue #1065, opened 2022-03-01
Fix flake8 violations for all files in the asv/commands/ directory


dorothykiz1:
We are trying to clean out all the files to follow Flake8 standards

Use this guide below to install flake8 and how to use it.
Flake8 guide

 Here are a few steps to follow:

-     You will be required to remove the files you would like to work on from the setup. cfg file under the extend attribute .
-     Create a virtual environment virtualenv venv venv will be the name for the environment and activate it by running source venv/bin/activate these commands are dependent on the operating system. (These are Linux commands, you can find for windows and macOS online)
-     Run pip install flake8 to install flake8
-     Run flake8 --exclude="" asv/commands/ .
-     Work on all the highlighted errors

Expected errors to be fixed:
```
asv/commands/common_args.py:156:82: E226 missing whitespace around arithmetic operator
asv/commands/machine.py:3:1: F401 'six' imported but unused
asv/commands/rm.py:3:1: F401 'six' imported but unused
asv/commands/run.py:13:1: F401 'six' imported but unused
asv/commands/run.py:220:100: E501 line too long (100 > 99 characters)
asv/commands/run.py:244:100: E501 line too long (104 > 99 characters)
asv/commands/run.py:472:49: W503 line break before binary operator
asv/commands/run.py:526:100: E501 line too long (104 > 99 characters)
asv/commands/show.py:3:1: F401 'os' imported but unused
asv/commands/show.py:57:26: W503 line break before binary operator
asv/commands/show.py:258:100: E501 line too long (105 > 99 characters)
asv/commands/find.py:103:100: E501 line too long (101 > 99 characters)
asv/commands/find.py:176:30: E225 missing whitespace around operator
asv/commands/find.py:223:13: E303 too many blank lines (2)
asv/commands/__init__.py:6:1: F401 'six' imported but unused
asv/commands/compare.py:142:26: W503 line break before binary operator
asv/commands/compare.py:161:14: E713 test for membership should be 'not in'
asv/commands/profiling.py:183:100: E501 line too long (109 > 99 characters)
asv/commands/continuous.py:4:1: F401 'six' imported but unused
asv/commands/continuous.py:88:100: E501 line too long (100 > 99 characters)
asv/commands/preview.py:26:25: E226 missing whitespace around arithmetic operator
asv/commands/preview.py:27:50: E226 missing whitespace around arithmetic operator
asv/commands/preview.py:27:55: E226 missing whitespace around arithmetic operator
[...]

PR #1082, opened 2022-03-02
Fix flake8 issues on asv/commands/run.py

LucyJimenez:
xref https://github.com/airspeed-velocity/asv/issues/1065

Clean out the `run.py` files to follow Flake8 standards on the asv/commands folder.

### DIFF
--- a/asv/commands/run.py
+++ b/asv/commands/run.py
@@ -214,8 +214,8 @@ class Run(Command):
             if dry_run:
                 raise util.UserError("--interleave-rounds and --dry-run cannot be used together")
             if has_existing_env:
-                raise util.UserError("--interleave-rounds cannot be used with existing environment "
-                                     "(or python=same)")
+                raise util.UserError("--interleave-rounds cannot be used with existing "
+                                     "environment (or python=same)")
         elif interleave_rounds is None:
             # Enable if possible
             interleave_rounds = not (dry_run or has_existing_env)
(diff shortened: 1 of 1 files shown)


asv/commands/run.py
```diff
@@ -238,7 +238,8 @@ def run(cls, conf, range_spec=None, steps=None, date_period=None,
 
         if range_spec is None:
             try:
-                commit_hashes = list(set([repo.get_hash_from_name(branch) for branch in conf.branches]))
+                commit_hashes = list(set([(repo.get_hash_from_name(branch) for
```
datapythonista, line 241: You are adding a parenthesis after the squared bracket. That is changing the code, instead of the comprehension returning a list of the iterated items, this is returning a list with a single value which is a generator. This is why tests are failing.

LucyJimenez: Thank you for the explanation. I'm going to fix it!

datapythonista:
The CI problem is unrelated to your changes. We should create an issue, and mention that this happens sometimes (seems like the conda servers are not very reliable, something that it's probably going to fix itself, but better to have an issue, in case something can be done in our end).

Then, we've got these files excluded from flake8 validation in the CI. We should stop excluding them if you are fixing their errors, so we see this is working as expected, and no more style errors can be introduced to those files in future PRs.

LucyJimenez:
I will review the last file in the folder `asv/commands` and then excluded it from flake8 validation in the CI.

datapythonista:
Thanks @LucyJimenez, very nice

PR #1083, opened 2022-03-03
Remove PY2 six from asv/commands/quickstart.py

### LINKED ISSUE #1040: Remove all six functionality

dorothykiz1:
closes #1040

This will remove six functionality in the following files;

- asv/commands/quickstart.py


dorothykiz1:
For this PR, I get this error 
`FAILED test/test_quickstart.py::test_quickstart - OSError: pytest: reading from stdin while output is captured.`
i see suggestions of [pytest monkeypatch](https://stackoverflow.com/questions/59986625/how-to-simulate-two-consecutive-console-inputs-with-pytest-monkeypatch) or [mocking](https://stackoverflow.com/questions/35851323/how-to-test-a-function-with-input-call). what are your thoughs on this?

datapythonista:
The problem here is that `pytest` captures the output of the program, and `input` requires the user to provide something (which shouldn't happen in tests, since there is no user).

I think the best for now is:
- Call `pytest` with the `-s` parameter in the CI (which is not great, as the output is quite valuable), but should fix this for now
- Create an issue to allow calling this function from the tests without having to get input from the user, and to remove the `-s` from the pytest calls

What do you think?

datapythonista:
@dorothykiz1 do you want to finish this one? I see that this is a duplicate of #1068, you may want to coordinate with that one, and get one of them working, so it can be merged.

dorothykiz1:
This suggestion pytest -s  test/test_quickstart.py didn't work so we need to go for a solution of not calling user input in tests.

Issue #1077, opened 2022-03-02
Use f-strings where possible


datapythonista:
xref #909 

Since we support Python > 3.7 now, we can make use of f-strings, to make code more readable.

As an example, the next `asv` code:

```python
machine_env_name = "{}/{}".format(machine, env_name)
```

or the next, which is equivalent:

```python
machine_env_name = "{0}/{1}".format(machine, env_name)
```

Could be written as:

```python
machine_env_name = f"{machine}/{env_name}"
```
(note the `f` before the string quotes)

There are lots of those in the codebase. This would be good as something to work on during sprints, but we should create smaller issues, for specific files for those.

The next command returns the list of strings to replace: `grep -R --exclude-dir="build" --exclude="*.css" ".format(" *`

dorothykiz1:
Refer to this example PR [#1087](https://github.com/airspeed-velocity/asv/pull/1087)

LucyJimenez:
@datapythonista and @dorothykiz1 I'm going to work on this to create a couple of examples more for the sprint.

HaoZeke:
Closed by #1258.

PR #1087, opened 2022-03-03
Replace format() function with PY3 fstrings

dorothykiz1:
xref #1077

### DIFF
--- a/asv/plugins/summarylist.py
+++ b/asv/plugins/summarylist.py
@@ -52,7 +52,8 @@ class SummaryList(OutputPublisher):
                     pretty_name = benchmark['pretty_name']
 
                 if idx is not None:
-                    pretty_name = f'{pretty_name}({", ".join(benchmark_param)})'
+                    bench_param = ", ".join(benchmark_param)
+                    pretty_name = f'{pretty_name}({bench_param})'
 
                 # Each environment parameter combination is reported
                 # separately on the summarylist page
--- a/asv/plugins/summarylist.py
+++ b/asv/plugins/summarylist.py
@@ -52,8 +52,7 @@ class SummaryList(OutputPublisher):
                     pretty_name = benchmark['pretty_name']
 
                 if idx is not None:
-                    pretty_name = '{0}({1})'.format(pretty_name,
-                                                    ", ".join(benchmark_param))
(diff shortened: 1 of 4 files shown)


asv/plugins/summarylist.py
```diff
@@ -52,8 +52,7 @@ def publish(cls, conf, repo, benchmarks, graphs, revisions):
                     pretty_name = benchmark['pretty_name']
 
                 if idx is not None:
-                    pretty_name = '{0}({1})'.format(pretty_name,
-                                                    ", ".join(benchmark_param))
+                    pretty_name = f'{pretty_name}({", ".join(benchmark_param)})'
```
datapythonista, line 55: This is fine, but I don't think we're improving readability that much. I forgot to mention in the issue I think, but for non-trivial code, would be better to first assign to a variable, and then have the f-string, but with the variable name, not with the whole expression. In this case it's not that complex a simple join, but having quotes and commas on the expression, makes the f-string quite unreadable in my opinion.

asv/plugins/virtualenv.py
```diff
@@ -130,15 +130,15 @@ def _setup(self):
...
+        log.info(f"Creating virtualenv for {self.name}")
         util.check_call([
             sys.executable,
             "-mvirtualenv",
             "-p",
             self._executable,
             self._path], env=env)
 
-        log.info("Installing requirements for {0}".format(self.name))
+        log.info("Installing requirements for {self.name}")
```
datapythonista, line 141: Not an f-string. We need to be careful with that, since it's very likely that not flake8 and not the tests will warn us when we miss the `f` to make an f-string, and we'll lose the value of the variables in the string.

datapythonista (commented):
Looks good, thanks for working on this. Just added couple of things that need to be fixed.

asv/plugins/summarylist.py
dorothykiz1: Thanks i had thought about this,makes total sense

asv/plugins/virtualenv.py
dorothykiz1: noted

datapythonista:
Thanks @dorothykiz1, really nice

Issue #1044, opened 2022-02-21
Fix flake8 violations for all files in the asv/plugins/ directory


dorothykiz1:
We are trying to clean out all the files to follow Flake8 standards

Use this guide below to install flake8 and how to use it.
[ Flake8 guide](https://siderlabs.com/blog/automatically-review-code-for-python-projects-using-flake8-6fcb056a001a/)

Here are a few steps to follow:
- You will be required to remove the files you would like to work on from the `setup. cfg` file under the `extend` attribute .
- Create a virtual environment `virtualenv venv` venv will be the name for the environment and activate it by running `source venv/bin/activate` these commands are dependent on the operating system. (These are Linux commands, you can find for windows and macOS online)
- Run `pip install flake8` to install flake8
- Run `flake8 --exclude=""  asv/plugins/` . 
- Work on all the highlighted errors

Expected errors to be fixed:
```
asv/plugins/summarylist.py:5:1: F401 'six' imported but unused
asv/plugins/summarylist.py:15:1: E303 too many blank lines (3)
asv/plugins/summarylist.py:94:60: E226 missing whitespace around arithmetic operator
asv/plugins/virtualenv.py:8:1: F401 'six' imported but unused
asv/plugins/virtualenv.py:54:13: F401 'virtualenv' imported but unused
asv/plugins/git.py:36:28: E127 continuation line over-indented for visual indent
asv/plugins/git.py:69:17: W503 line break before binary operator
asv/plugins/git.py:152:9: F841 local variable 'err' is assigned to but never used
asv/plugins/__init__.py:2:1: W391 blank line at end of file
asv/plugins/conda.py:7:1: F401 'multiprocessing' imported but unused
asv/plugins/conda.py:9:1: F401 'six' imported but unused
asv/plugins/conda.py:25:1: E302 expected 2 blank lines, found 1
asv/plugins/conda.py:183:100: E501 line too long (116 > 99 characters)
asv/plugins/conda.py:187:100: E501 line too long (104 > 99 characters)
asv/plugins/mercurial.py:12:1: F841 local variable 'exc' is assigned to but never used
asv/plugins/mercurial.py:94:16: F523 '...'.format(...) has unused arguments at position(s): 2
asv/plugins/regressions.py:7:1: F401 'six' imported but unused
asv/plugins/regressions.py:58:28: E128 continuation line under-indented for visual indent
asv/plugins/regressions.py:145:100: E501 line too long (103 > 99 characters)
[...]

PR #1088, opened 2022-03-03
Fix flake8 violations for all files in the asv/plugins/ directory

LucyJimenez:
xref https://github.com/airspeed-velocity/asv/issues/1044

Clean out some of the files to follow Flake8 standards on the asv/plugins folder.

### DIFF
--- a/asv/plugins/__init__.py
+++ b/asv/plugins/__init__.py
@@ -1,2 +1,1 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-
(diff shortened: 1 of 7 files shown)


datapythonista:
Thanks @LucyJimenez. The `noqa` message is a bit confusing, not sure which previous step you are referring to. The import is needed even if not used, since the goal is to check if the library exists.

LucyJimenez:
> Thanks @LucyJimenez. The `noqa` message is a bit confusing, not sure which previous step you are referring to. The import is needed even if not used, since the goal is to check if the library exists.

Thanks @datapythonista. It is a nice recommendation to take in mind for the next messages :)

PR #1089, opened 2022-03-03
Minor changes for #1065 and #1044

LucyJimenez:
Closes https://github.com/airspeed-velocity/asv/issues/1065 https://github.com/airspeed-velocity/asv/issues/1044
Minor changes:
- improve noqa message on `asv/plugins/virtualenv.py`
- remove folder excluded for flake8:  `asv/plugins`, `asv/commands`


asv/plugins/virtualenv.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, conf, python, requirements, tagged_env_vars):
                                          tagged_env_vars)
 
         try:
-            import virtualenv # noqa F401 unused, but required in case the previous step does not work
+            import virtualenv # noqa F401 unused, but is needed even if not used
```
datapythonista, line 51: I think what you're saying here doesn't add much value. It's clearly needed if unsused, that's why we have the noqa. Would be more useful to explain why is needed.

datapythonista: This line has an import, which is unused. We don't want to have unused imports in general, there is no point. In this case we're going to leave it, and we're telling flake8 this is fine, we want to leave an unused import. That's fine, but let readers of this code know why we're leaving this import. And saying "because it's needed" it's like not saying anyting. Please add here a comment on why this import is needed. In this case it's quite obvious, and could not even be needed, but a short sentence to help readers would still be nice.

setup.cfg
```diff
@@ -5,12 +5,10 @@ ignore =
     E741  # E741: Do not use variables named 'I', 'O', or 'l'
 exclude = setup.py,
         asv/benchmarks,
-        asv/commands,
         asv/environment.py,
         asv/machine.py,
         asv/graph.py,
         asv/extern,
-        asv/plugins,
         asv/template/benchmarks,
```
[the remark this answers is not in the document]
LucyJimenez: I created new issues
- https://github.com/airspeed-velocity/asv/issues/1097
- https://github.com/airspeed-velocity/asv/issues/1099

setup.cfg
```diff
@@ -4,12 +4,11 @@ ignore =
     W504,  # W504: Line break occurred after a binary operator
     E741  # E741: Do not use variables named 'I', 'O', or 'l'
 exclude = setup.py,
-        asv/commands,
+        asv/benchmarks,
```
datapythonista, line 7: Why are we reintroducing this to the excludes?

asv/plugins/virtualenv.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, conf, python, requirements, tagged_env_vars):
                                          tagged_env_vars)
 
         try:
-            import virtualenv # noqa F401 unused, but required in case the previous step does not work
+            import virtualenv # noqa F401 unused, but required to manage an environment
```
datapythonista, line 51: This is not required to manage any environment. This import is just used to test whether virtualenv is installed or not, and its goal is to fail here with a useful error message (instead of elsewhere I guess) if the package is not installed.

setup.cfg
LucyJimenez: Sorry, it was a mistake.

asv/plugins/virtualenv.py
LucyJimenez: Great, thank you so much for the explanation!

datapythonista:
Thanks @LucyJimenez

Issue #1030, opened 2022-02-15
Move pytest fixtures to conftest.py


dorothykiz1:
When fixtures are not defined in the standard `conftest.py`  and are defined elsewhere in other files, 
once imported for reference they are not recognized.

Flake8 example error  : ` F401 '.tools.example_results' imported but unused`

What is expected is to move all fixtures to the `conftest.py` file, this will not require the unnecessary imports for fixtures hence no errors of not found fixtures in the CI as shown below.

ie.` fixture 'example_results' not found`

dorothykiz1:
Most of the fixtures have been moved to the conftest.py file and 3 files with  4 fixtures are pending,i have opened PRs for them i will be editing this comment once merged

```
(venv) dorothy@dorothy-Latitude-5480:~/Desktop/sprint/asv$ grep -R --exclude-dir="build" --exclude="*.css" "@pytest.fixture" *
test/test_dev.py:@pytest.fixture
test/test_dev.py:@pytest.fixture
test/test_step_detect.py:@pytest.fixture(params=[
test/test_publish.py:@pytest.fixture(params=[
test/conftest.py:@pytest.fixture(params=[
test/conftest.py:@pytest.fixture
test/conftest.py:@pytest.fixture
test/conftest.py:@pytest.fixture
test/conftest.py:@pytest.fixture(scope="session")
test/conftest.py:@pytest.fixture(scope="session")
test/conftest.py:@pytest.fixture
test/conftest.py:@pytest.fixture(scope="session")
test/conftest.py:@pytest.fixture
test/conftest.py:@pytest.fixture
```

PR #1100, opened 2022-03-04
Move example-results fixture to conftest.py file 

dorothykiz1:
xref #1030

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,5 +1,11 @@
 import os
 import contextlib
+from os.path import abspath, join, dirname
+from .tools import run_asv_with_conf
+import pytest
+from asv import config
+import shutil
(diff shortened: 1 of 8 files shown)


dorothykiz1:

datapythonista:
Thanks @dorothykiz1. It'd be better to not mix things, like moving the fixture and fixing flake8 issues, in the same PR. Better to address them separately.

dorothykiz1:
Thanks @datapythonista totally agree,however i have noticed if i make changes the CI flags If the files are not Flake 8 compliant

datapythonista:
I see. It would still be better to open a PR first for flake8 or any file you need, and then a separate PR for the move.

dorothykiz1:
Thanks for your guidance, i will follow this.

Issue #1063, opened 2022-03-01
Add a job to build the docs on CI


LucyJimenez:
This issue is part of the improvements to the CI because we want to move from Appveyor to GitHub actions. One of the improvements includes adding a job to build the docs.

- Move docs dependencies to `requirements-dev.txt`, and delete the docs requirements (too much complexity having two files, for little gain)
- Create a new build in the CI similar to the tests one (install deps, build extension), and run `sphinx-build -W docs/source html` instead of pytest
- Fix any problem in the docs that happens when the above is executed

datapythonista:
All this needs is:

- Move docs dependencies to `requirements-dev.txt`, and delete the docs requirements (too much complexity having two files, for little gain)
- Create a new build in the CI similar to the tests one (install deps, build extension), and run `sphinx-build -W docs/source html` instead of pytest
- Fix any problem in the docs that happens when the above is executed

LucyJimenez:
Thank you, Marc. I'll include it in the description.

LucyJimenez:
I'm going to work on this issue.

datapythonista:
Fixed in #1107

PR #1107, opened 2022-03-07
Add a job to build the docs on CI

LucyJimenez:
xref https://github.com/airspeed-velocity/asv/issues/1063
On this PR create a job for docs


datapythonista:
Did you check if the docs are being built correctly?

LucyJimenez:
> Did you check if the docs are being built correctly?

Yes, now it’s possible to create the HTML. I’m working on fixing the problems, and I’ll be listing the issues here.

- `asv/docs/source/manindex.rst:document isn't included in any toctree`
- `asv/docs/source/benchmarks.rst:148:py:obj reference target not found: time.clock`

LucyJimenez:
@datapythonista The PR is ready for your review; thanks!

.github/workflows/ci.yml
```diff
@@ -47,3 +47,18 @@ jobs:
 
       - name: Check code with flake8
         run: flake8 .
+
+  docs:
+    name: builds the documentation using Sphinx
```
datapythonista, line 52: I think you can delete this row. The default `docs` is easier to read in my opinion.

docs/source/reference.rst
```diff
@@ -3,6 +3,7 @@ Reference
 
 .. toctree::
 
+   manindex.rst
```
datapythonista, line 6: This seems to be a duplicate of reference, just remove the page instead of including it here.

LucyJimenez:
@datapythonista, I included the last comments. The PR is ready for review. Thanks a lot!

.github/workflows/ci.yml
```diff
@@ -46,6 +46,20 @@ jobs:
 
       - name: flake8
         run: flake8 .
-
+
```
datapythonista, line 49: You should review the diffs when you're done, and things like this shouldn't be introduced by mistake.

datapythonista (commented):
If having an `xref` in the description correct? Or should we have a close and let this automatically close the issue?

datapythonista:
Thanks @LucyJimenez

PR #1113, opened 2022-03-08
CLN: clean setup.py, remove py.test command

fangchenli:


setup.py
```diff
@@ -1,49 +1,11 @@
-#!/usr/bin/env python
```
datapythonista, line 1: I'd leave this.

setup.py
```diff
@@ -1,49 +1,11 @@
...
-            test_args += self.pytest_args.split()
-        if self.coverage:
-            test_args += ['--cov', os.path.abspath('asv')]
-        errno = pytest.main(test_args)
-        if errno != 0:
-            raise DistutilsError("Tests failed")
+from setuptools import Extension, setup
+from setuptools.command.build_ext import build_ext
+from setuptools.command.sdist import sdist
+from setuptools.errors import CCompilerError, ExecError, PlatformError
```
datapythonista, line 8: Looks like `CCompilerError` can't be imported, not sure what's the new exception.

datapythonista (commented):
Nice clean up, just couple of small things.

setup.py
fangchenli: setuptools vendored those exceptions from distutils. And it's only available on very recent releases. I'll revert this.

datapythonista:
Thanks for the clean up @fangchenli

PR #1118, opened 2022-03-09
BLD: move metadata to setup.cfg, use PEP 517, install asv in CI

fangchenli:


datapythonista:
Thanks @fangchenli, very nice

PR #1122, opened 2022-03-09
Move test/test_show.py  fixture to conftest.py

dorothykiz1:
xref #1030


dorothykiz1:
@datapythonista this PR is ready for your review, I also noticed first, the PyPy 3.8 CI had failed as the other , PyPy 3.7 CI was loading, after it all finally passed, just wondering if it is something to worry about,

datapythonista:
Thanks @dorothykiz1. Would be nice to have some context on how many fixtures we're moving. Can you add the list of the ones that still need to be moved to the issue?

dorothykiz1:
Thanks @datapythonista ,I will do this

PR #1125, opened 2022-03-10
CLN: Remove py2 compat in extension

fangchenli:


asv/_rangemedian.cpp
```diff
@@ -408,10 +399,7 @@ static PyTypeObject RangeMedianType = {
 
 static PyTypeObject *RangeMedian_init_type(PyObject *m)
 {
-#if PY_MAJOR_VERSION < 3
     RangeMedianType.ob_type = &PyType_Type;
```
datapythonista, line 402: Shouldn't this be removed?

datapythonista:
Thanks @fangchenli, nice clean up. I'm not sure if this extension is actually being used in the released version. There is a fallback for it when it's not available, and we've been releasing as `noarch`, so I guess it may not be getting compiled, and ignored when installing asv with `pip install asv`. Not sure how useful it is compared to the fallback, maybe we should keep things simple and get rid of the extension.

PR #1126, opened 2022-03-10
CLN: clean stats functions

fangchenli:


test/test_statistics.py
```diff
@@ -1,14 +1,13 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-import math
 import random
 import warnings
+from math import factorial, inf
```
datapythonista, line 5: Why do you prefer this way? It doesn't feel like using `math.inf` in the code is too verbose. And I think it's clearer, since reading the code, without checking the imports you know whether it's `math.inf`, `numpy.inf`, the previously used here `inf = float('inf')` or something else.

datapythonista:
Thanks @fangchenli, very nice clean up

PR #1127, opened 2022-03-10
Replace format() function with f-strings

LucyJimenez:
Change f-strings instead `.format()` on `asv/console.py`.


LucyJimenez:
@datapythonista, this PR is ready for your review. Thanks a lot!

asv/console.py
```diff
@@ -218,8 +218,7 @@ def _stream_formatter(self, record):
         continued = getattr(record, 'continued', False)
 
         if self._total:
-            progress_msg = '[{0:6.02f}%] '.format(
-                (float(self._count) / self._total) * 100.0)
+            progress_msg = f'[{(float(self._count) / self._total) * 100.0:6.02f}%] '
```
datapythonista, line 221: This is very difficult to read, you can define a variable instead.

asv/console.py
```diff
@@ -218,8 +218,8 @@ def _stream_formatter(self, record):
         continued = getattr(record, 'continued', False)
 
         if self._total:
-            progress_msg = '[{0:6.02f}%] '.format(
-                (float(self._count) / self._total) * 100.0)
+            percentage = (float(self._count) / self._total) * 100.0
+            progress_msg = f'[{percentage:6.02f}%] '
```
datapythonista, line 222: The original code wasn't great. In Python you can actually do this:
```python
>>> value = .2
>>> f'{value:.2%}'
'20.00%'
```

Also, converting `self._count` to `float` only made sense in Python 2, in Python 3 it's not needed. So, this could be something like (make sure we're not changing the result):
```python
progress_msg = f'[{self._count / self._total:6.02%}] '
```

LucyJimenez: Done! thanks for the suggestion :)

PR #1129, opened 2022-03-10
Replace format() function with f-strings on asv/benchmark.py

LucyJimenez:
Change f-strings instead `.format()` on `asv/benchmark.py`


LucyJimenez:
@datapythonista Do you know the best way to replace the `.format()` function in those cases? Could you please help me?

image

image

dorothykiz1:
@LucyJimenez, I think you can declare with a variable and call that variable in the `f string `to make it more readable

datapythonista:
> @datapythonista Do you know the best way to replace the `.format()` function in those cases? Could you please help me?
> 
> image
> 
> image

Can you add this as a comment in the relevant code, instead of having an image, so I can comment over the code, or copy it, please?

I'd probably use `'    ' + src.replace(...)`

asv/benchmark.py
```diff
@@ -344,7 +342,8 @@ def get_source_code(items):
             src = "class {0}:\n    {1}".format(
                 class_name, src.replace("\n", "\n    "))
         elif class_name:
-            src = "    {0}".format(src.replace("\n", "\n    "))
+            nl = '\n'
+            f"    {src.replace(nl, nl+'    ')}"
```
datapythonista, line 346: This doesn't make sense.

datapythonista, line 346:
```suggestion
            src = "    " + src.replace("\n", "\n    ")
```

datapythonista:
Thanks @LucyJimenez

PR #1131, opened 2022-03-10
Move fixture from test/test_step_detect.py file

dorothykiz1:
xref #1030

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -7,13 +7,15 @@ from os.path import abspath, dirname, join
 import pytest
 import selenium
 
-from asv import config, environment, repo
+from asv import config, environment, repo, step_detect
 from asv.repo import get_repo
+from asv.step_detect import L1Dist, _rangemedian
 
 from . import tools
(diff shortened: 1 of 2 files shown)


dorothykiz1:
I ran `isort` locally on this particular file seems fine but complains here , could it be skipping the imports with inline comments?

datapythonista:
> @datapythonista I ran `isort` locally on this particular file seems fine but complains here , could it be skipping the imports with inline comments? Is there a way to not skip imports with comments? Below is the error returned;
> 
> ```
> ERROR: /home/runner/work/asv/asv/test/test_step_detect.py Imports are incorrectly sorted and/or formatted.
> Skipped 1 files
> Error: Process completed with exit code 1.
> ```

Can you better comment in the relevant part of the code? It's making my life unnecessarily harder to have to read your comment in one tab, and switch to another and find the exact part you're taling about, to see what code you're referring to. Thanks!

dorothykiz1:
Let me do this

datapythonista:
Thanks @dorothykiz1

PR #1133, opened 2022-03-10
Move fixture from test/test_show.py file 

dorothykiz1:
xref #1030


dorothykiz1:
This was a duplicate PR with #1122  that is already merged

PR #1134, opened 2022-03-11
BLD: use setuptools-scm for versioning

fangchenli:


datapythonista (commented):
Very nice clean up. Did you check if there is a test for this? Would be nice to know with a CI test if this is working as expected.

fangchenli:
> Very nice clean up. Did you check if there is a test for this? Would be nice to know with a CI test if this is working as expected.

There is no explicit test just for this. But there are tests using `asv.__version__` which is written by `setuptools-scm`. Also, it doesn't work, we couldn't even import `asv`.

PR #1135, opened 2022-03-11
BLD: simplify build_ext, remove doc build command

fangchenli:


setup.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 import os
 import subprocess
-from distutils.errors import CCompilerError, DistutilsExecError, DistutilsPlatformError
+from distutils.errors import CompileError
```
fangchenli, line 4: removing this requires either a very new setuptools version or using the wildcard `BaseException`.

PR #1148, opened 2022-03-14
CLN:Clean up setup.cfg file and flake8 errors on test/test_repo_template/setup.py file

dorothykiz1:


test/test_repo_template/asv_test_repo/__init__.py
```diff
@@ -1,3 +1,3 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-dummy_value = {dummy_value}
+dummy_value = {dummy_value}  # noqa see #1147
```
datapythonista, line 3:
```suggestion
dummy_value = {dummy_value}  # noqa <FLAKE8_ERROR_CODE> This is a template that will be rendered to valid Python
```

test/test_repo_template/setup.py
```diff
@@ -3,7 +3,7 @@
 from setuptools import setup
 
 with open('asv_test_repo/build_time_env.py', 'w') as f:
-    f.write("env = {{}}\n".format(repr(dict(os.environ))))
+    f.write("env = {{}}\n".format(repr(dict(os.environ))))  # noqa see #1144
```
datapythonista, line 6: Can you simply change the string for `f'env = {os.environ}'`. Or if it doesn't work just `env = {}`. This code doesn't make any sense, it simply writes `env = {}`, but I guess the intention was to have the envionment, so let's try first if that works.

dorothykiz1: Thank you,Yes,ran it locally it just returns `env = {}`

datapythonista: I see, can you try with `f'env = {os.environ}'` which makes more sense.

test/test_repo_template/setup.py
```diff
@@ -3,7 +3,7 @@
 from setuptools import setup
 
 with open('asv_test_repo/build_time_env.py', 'w') as f:
-    f.write("env = {{}}\n".format(repr(dict(os.environ))))
+    f.write(f'env = {os.environ}')
```
datapythonista, line 6:
```suggestion
    f.write('env = {}')
```

datapythonista:
Can you split the changes here in two separate PRs please? Can't understand what's the problem, and having unrelated changes mixed doesn't help.

test/test_repo_template/setup.py
dorothykiz1, line 6: looks like this we need to be specific on the key we want to return from this particular output that has environment varibles, hence `E                   KeyError: 'os' `

datapythonista, line 6:
```suggestion
    f.write('env = {}')
```

test/test_repo_template/setup.py
```diff
@@ -3,7 +3,7 @@
 from setuptools import setup
 
 with open('asv_test_repo/build_time_env.py', 'w') as f:
-    f.write("env = {{}}\n".format(repr(dict(os.environ))))
+    f.write('env = {}')
```
datapythonista, line 6:
```suggestion
    f.write("env = {{}}\n".format(repr(dict(os.environ))))
```

datapythonista:
Not sure what's going on with this. You can leave the file in the exclude list for now, or add a `noqa` with the error to ignore, and a link to an issue explaining the problem.

datapythonista:
The file is already excluded, this doesn't make much sense if you don't remove `setup.py`, which is more generic and is also excluding other files with the same name.

And as said, if we're to exclude this file, add the issue with the problem to the exclude list.

dorothykiz1:
Thanks i will do it

datapythonista:
Superseded by #1160

PR #1149, opened 2022-03-14
CLN: remove submodule, fetch js file from cdn

fangchenli:
the vendored js files is from https://github.com/xuanluo/flot-axislabels


datapythonista:
Thanks @fangchenli, really nice

PR #1188, opened 2022-03-20
CLN: clean pytest config

fangchenli:
Those markers come with their plugins.


datapythonista:
Cool, nice catch, thanks for the clean up

PR #1194, opened 2022-03-25
Replace format() with fstrings in asv/statistics.py and asv/config.py

### LINKED ISSUE #1169: [Sprint] Use f-strings instead .format() function on asv/config.py
xref [#909](https://github.com/airspeed-velocity/asv/issues/909)

Since we support Python > 3.7 now, we can make use of f-strings, to make code more readable.

As an example, the next `asv` code:
`machine_env_name = "{}/{}".format(machine, env_name)`
or the next, which is equivalent:
`machine_env_name = "{0}/{1}".format(machine, env_name)`
Could be written as:
`machine_env_name = f"{machine}/{env_name}"`
(note the `f` before the string quotes)

The next command returns the list of strings to replace: 
 `grep -R --exclude-dir="build" --exclude="*.css" ".format(" asv/config.py`

For this file, you will find the following issues to solve:
```
asv/config.py:        self.pythons = ["{0[0]}.{0[1]}".format(sys.version_info)]
asv/config.py:            raise util.UserError("Config file {0} not found.".format(path))
asv/config.py:                "No repo specified in {0} config file.".format(path))
```

dorothykiz1:
close #1169
xref #1077

### DIFF
--- a/asv/config.py
+++ b/asv/config.py
@@ -21,7 +21,7 @@ class Config:
         self.repo = None
         self.repo_subdir = ""
         self.branches = [None]
-        self.pythons = ["{0[0]}.{0[1]}".format(sys.version_info)]
+        self.pythons = [f"{sys.version_info[0]}.{sys.version_info[1]}"]
         self.matrix = {}
         self.exclude = []
         self.include = []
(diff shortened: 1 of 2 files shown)


datapythonista:
Thanks @dorothykiz1

Issue #1173, opened 2022-03-18
[Sprint] Use f-strings instead .format() function on asv/plugins/mercurial.py


LucyJimenez:
xref [#909](https://github.com/airspeed-velocity/asv/issues/909)

Since we support Python > 3.7 now, we can make use of f-strings, to make code more readable.

As an example, the next asv code:
`machine_env_name = "{}/{}".format(machine, env_name)`
or the next, which is equivalent:
`machine_env_name = "{0}/{1}".format(machine, env_name)`
Could be written as:
`machine_env_name = f"{machine}/{env_name}"`

(note the `f` before the string quotes)

---

The next command returns the list of strings to replace: 

 `grep -R --exclude-dir="build" --exclude="*.css" ".format(" asv/plugins/mercurial.py`

For this file, you will find the following issues to solve:
```
asv/plugins/mercurial.py:        return '{0}::{1} and not {0}'.format(commit_a, commit_b)
asv/plugins/mercurial.py:        return '{0}::{1}'.format(latest_result, self.get_branch_name(branch))
asv/plugins/mercurial.py:        range_spec = self._encode("sort({0}, -rev)".format(range_spec))
asv/plugins/mercurial.py:        return self.get_hash_from_name('p1({0})'.format(name))
asv/plugins/mercurial.py:        return self.get_hashes_from_range(query.format(self.get_branch_name(branch)),
```

LucyJimenez:
I'm going to work on thin one.

datapythonista:
Not useful anymore, and add noise to the issues list.

PR #1198, opened 2022-03-31
Replace format() function with f-strings on asv/plugins/mercurial.py

LucyJimenez:
Closes https://github.com/airspeed-velocity/asv/issues/1173

Change f-strings instead `.format()` on `asv/plugins/mercurial.py'

### DIFF
--- a/asv/plugins/mercurial.py
+++ b/asv/plugins/mercurial.py
@@ -92,10 +92,10 @@ class Hg(Repo):
         return False
 
     def get_range_spec(self, commit_a, commit_b):
-        return '{0}::{1} and not {0}'.format(commit_a, commit_b)
+        return f'{commit_a}::{commit_b} and not {commit_a}'
 
     def get_new_range_spec(self, latest_result, branch=None):
-        return '{0}::{1}'.format(latest_result, self.get_branch_name(branch))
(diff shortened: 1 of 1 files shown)


asv/plugins/mercurial.py
```diff
@@ -169,11 +169,10 @@ def get_date_from_name(self, name):
 
     def get_branch_commits(self, branch):
         if self._repo.version >= (4, 5):
-            query = "branch({0})"
+            query = f"branch({self.get_branch_name(branch)})"
```
datapythonista, line 172: The original is ok, you can remove the `0` and just leave `{}`, but don't convert to a f-string

asv/plugins/mercurial.py
```diff
@@ -169,11 +169,10 @@ def get_date_from_name(self, name):
...
         else:
-            query = "ancestors({0})"
-        return self.get_hashes_from_range(query.format(self.get_branch_name(branch)),
-                                          followfirst=True)
+            query = f"ancestors({self.get_branch_name(branch)})"
+        return self.get_hashes_from_range(query, followfirst=True)
```
datapythonista, line 175: Same, this was ok, you can just remove the `0`

mattip:
Superseded by #1258

PR #1202, opened 2022-04-05
Remove asv dev command

### LINKED ISSUE #1200: Remove asv dev command
The asv dev command is confusing, and made few projects pin to asv 0.4 after 0.5 changed it's behavior.

I'm going to remove it. In asv 0.6, I'll keep the command, but making it fail with a descriptive error message that asv run should be used instead, and then it can be removed in 0.7

datapythonista:
Closes #1200 

`asv dev` is literally a shortcut for `asv run --python=same`, so it's not that useful to have. And it caused confusion, as users expected it to also use `--quick`, which was true before asv 0.5. For now `asv dev` will continue to exist, but will return an exception informing the user on what to do.

`test_dev.py` contained some tests unrelated to `asv dev`. Those are moved to the right location here.


HaoZeke (approved):
LGTM, thanks!

PR #1204, opened 2022-04-05
Remove --strict option, and fail by default

### LINKED ISSUE #1199: Make --strict default and remove parameter
I think asv interface can be improved by making `--strict` parameter of the run command to be the default. It makes the command return an exit code != 0 if things break, which I think it's the behavior that user would expect.

I'll also remove the parameter, since ignoring errors can be simply done as `asv | true`.

datapythonista:
Closes #1199

`--strict` will now be the default, and the option will be removed. To ignore the exist status, for example in CI systems, `asv run || true` or similar can be used.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ New Features
 
 API Changes
 ^^^^^^^^^^^
+- ``asv run`` and ``asv continuous`` don't implement the ``--strict`` option anymore,
+  and they will always return a non-zero (i.e. ``2``) exit status if any benchmark fail.
 
 Bug Fixes
 ^^^^^^^^^
--- a/asv/commands/continuous.py
+++ b/asv/commands/continuous.py
@@ -47,10 +47,6 @@ class Continuous(Command):
         parser.add_argument(
             "--no-interleave-processes", action="store_false", dest="interleave_rounds",
             help=argparse.SUPPRESS)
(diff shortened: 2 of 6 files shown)


alimcmaster1 (approved):
Think this is sensible for consistency with other tooling - will this breaking API change warrant a new minor e.g 0.6 opposed to 0.5.2? And how come exit code 2 opposed to 1?

datapythonista:
We dropped py2 and will probably start raising few deprecation warnings, so planning 0.6.

Exiting with 2 was already implemented. I was wondering if a regular Python exception would raise 1, and 2 would mean benchmarks broke, and 1 would mean asv broke. But didn't check, so far it'll be consistent with that was implemented, we can double check what's the exit with a regular python traceback, and changes the codes if it makes sense.

yarikoptic:
overall: I welcome the changes like this!  And I do acknowledge that asv is in eternal `0.x` (like many of my projects too) so "no promises".  

But I think it is better to break API a little more gently, e.g. with a deprecation cycle (issue DeprecationWarning) so if downstream tools decide to be proactive and react proactively to become broken API (e.g. via `-WError` raising warnings to errors), they can, but also mortals which just use `asv` can keep using it for longer time without having their CIs etc get unnecessarily broken with a fresh 0.minor release of ASV.

PR #1206, opened 2022-04-05
Making the base Command class a base class, and documenting methods

datapythonista:
Just a small clean up to make command classes fail earlier if they miss a required method, and adding documentation to them.


alimcmaster1:
Looks great!

datapythonista:
Thanks for the review @alimcmaster1

PR #1207, opened 2022-04-07
Sync sphinx docs benchmark examples with quickstart template file

kanderso-nrel:
#953 updated the quickstart template benchmarks to work with python 3.  However, the same 2.x examples were copy/pasted into the sphinx docs and are still there today: https://asv.readthedocs.io/en/stable/using.html

This PR uses RST `.. include::` to read in the actual template file and include it as formatted python code during the docs build, removing the need to manually keep the docs in sync with the actual template.

### DIFF
--- a/docs/source/using.rst
+++ b/docs/source/using.rst
@@ -94,40 +94,10 @@ Once you've set up the project's configuration, you'll need to write
 some benchmarks.  The benchmarks live in Python files in the
 ``benchmarks`` directory.  The ``quickstart`` command has created a
 single example benchmark file already in
-``benchmarks/benchmarks.py``::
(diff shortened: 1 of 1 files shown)


datapythonista (approved):
Thanks @kanderso-nrel, looks great

kanderso-nrel:
FYI in case you're not aware, ReadTheDocs has a very handy "PR Builds" feature: https://docs.readthedocs.io/en/stable/pull-requests.html

The main advantage over the current `docs` GH Action would be that you could view the built docs at a special url.  Or maybe there's a way to do that with the current CI job?  Anyway just pointing it out :)

PR #1209, opened 2022-05-02
BUG: Allow branches that conflict with file names

NickCrews:
Before, if I had a file named "benchmarks" and a branch specified in asv.conf.json also named "benchmarks", then when I ran `asv publish` I would get

```
$ asv publish
[ 11.11%] · Loading machine info
[ 22.22%] · Getting params, commits, tags and branches
[ 22.22%] ·· Error running /usr/local/bin/git rev-list --first-parent benchmarks (exit status 128)
             STDOUT -------->
             
             STDERR -------->
             fatal: ambiguous argument 'benchmarks': both revision and filename
             Use '--' to separate paths from revisions, like this:
             'git <command> [<revision>...] -- [<file>...]'

  File "/Users/nickcrews/Documents/projects/dedupe/.venv/bin/asv", line 8, in <module>
    sys.exit(main())
....
```

Tested locally, it works fine now.

### DIFF
--- a/test/test_publish.py
+++ b/test/test_publish.py
@@ -435,7 +435,8 @@ def test_regression_atom_feed_update(dvcs_type, tmpdir):
         b_content = b.find('{http://www.w3.org/2005/Atom}content')
         assert a_content.text != b_content.text
 
-def test_gh1209(tmpdir):
+def test_branch_name_is_also_filename(tmpdir):
+    # gh-1209
     tmpdir = str(tmpdir)
     dvcs = tools.generate_test_repo(tmpdir, list(range(10)),
                                     dvcs_type="git",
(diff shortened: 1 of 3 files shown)


NickCrews:
If it's not too hard, and if you point me to where you want it, I can add a test for this regression. Thanks!

mattip:
Still missing a test ...

NickCrews:
If you're familiar with it, could you point me to where exactly you want me to add the test? I can dig around but I'm not at all familiar with this codebase.

HaoZeke:
Should be good to go now @mattip.

test/test_publish.py
```diff
@@ -432,3 +434,27 @@ def test_regression_atom_feed_update(dvcs_type, tmpdir):
         a_content = a.find('{http://www.w3.org/2005/Atom}content')
         b_content = b.find('{http://www.w3.org/2005/Atom}content')
         assert a_content.text != b_content.text
+
+def test_gh1209(tmpdir):
```
mattip, line 438: A docstring or changing the test name to something indicative of what is being tested will prevent confusion in the future. Maybe `test_branch_name_is_also_filename` ?

NickCrews:
Thank you for finishing this up, I really appreciate it!

PR #1211, opened 2022-05-14
DEPS: update asizeof, add license

### LINKED ISSUE #1139: DEPS: dependency on pympler
asv uses vendored `asizeof` function from `pympler` to compute the memory size of python objects. We need to update the vendored code and add its license or officially adopt it as a runtime/optional dependency.

fangchenli:
closes #1139


mattip:
Done differently in #1234

PR #1216, opened 2022-08-30
Give GraalPy some time to warm up by default

timfel:
[GraalPy](https://github.com/oracle/graalpython/) has two main modes, execution as native binary or hosted on the JVM. Both need warmup, but the latter needs more and since determining which is which is a bit specific, I figured it shouldn't hurt to use the conservative warmup time for both.


mattip:
Is that a time in seconds? It seems like 5.0 is a lot for something like the NumPy suite which has hundreds of small benchmarks.

PR #1220, opened 2022-10-12
Add machine comparison list for published results

qunaibit:
ASV tracks performance regression and improvements per machine and python versions. However, tracking performance on multiple configurations/optimizations options for python builds can only be done as separate machines configurations. Though, comparing machines against each other is only visible per benchmark view.

This PR will provide a comparison list for machines against each other. It can, also, exports the comparison list to `markdown` format ([example](https://github.com/hpyproject/numpy-hpy/wiki/Numpy-Benchmarks-(Sept-25,-2022))).

### DIFF
--- a/LICENSE.rst
+++ b/LICENSE.rst
@@ -1,4 +1,5 @@
 Copyright (c) 2011-2018, Michael Droettboom, Space Telescope Science Institute, Pauli Virtanen
+Copyright (c) 2022, Oracle and/or its affiliates.
 
 All rights reserved.
 


LICENSE.rst
```diff
@@ -1,4 +1,5 @@
 Copyright (c) 2011-2018, Michael Droettboom, Space Telescope Science Institute, Pauli Virtanen
+Copyright (c) 2022, Oracle and/or its affiliates.
```
datapythonista, line 2: Surely not merging with this.

datapythonista (changes requested):
Thanks for the contribution. Probably better if you implement this as a third-party package. Not only because the copyright, also because asv is mostly unmaintained, and the last thing we want is even more code to maintain in it.

mattip:
@qunaibit would you like to continue with this?

PR #1225, opened 2022-12-08
MAINT: Update bundled asizeof module

HaoZeke:
Personally I'd like to just depend on `pympler` instead of bundling...


mattip:
Done differently in #1234

PR #1221, opened 2022-11-20
Update copyright to 2022

hmaarrfk:
will likely get out of date too soon.

But it helps to signal "this project is alive and well" !!!

PR #1229, opened 2022-12-30
TST,BUG: Fix error with submodules

HaoZeke:
`git` had some security fixes due to CVE-2022-39253, which breaks the existing implementation for pulling submodules. [This post](https://vielmetti.typepad.com/logbook/2022/10/git-security-fixes-lead-to-fatal-transport-file-not-allowed-error-in-ci-systems-cve-2022-39253.html) has a good explanation.

### DIFF
--- a/asv/plugins/git.py
+++ b/asv/plugins/git.py
@@ -87,13 +87,13 @@ class Git(Repo):
     def checkout(self, path, commit_hash):
         def checkout_existing(display_error):
             # Deinit fails if no submodules, so ignore its failure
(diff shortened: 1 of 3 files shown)


HaoZeke:
@mattip this should fix the test errors blocking #1221.

mattip:
Still failing ...

HaoZeke:
Sorry, updated again to use the correct config.

mattip:
There is a linting problem. I will merge this, could you add a linting fix to another PR?

PR #1232, opened 2023-01-02
ENH: Handle default branch

### LINKED ISSUE #1231: BUG: `git` testsuite fails with `init.defaultBranch`
A lot of tests assume that when `git` repositories are created, the initial branch is `master`. This is still the default (hence no CI failures) but many users will have it set to something else. Anything else causes the test-suite to fail (all the shell commands error out when pointing to the wrong initial branch).

Reproducer:

```bash
git config --global init.defaultBranch main
pytest tests
```

HaoZeke:
Closes #1231. The general implementation is essentially [EAFP](https://docs.python.org/3/glossary.html#term-EAFP) and is a little clunky, but hopefully at some point the entire `git.py` plugin will go away in favor of say, [pygit2](https://www.pygit2.org).

Some implementation details are highlighted in the comments.

### DIFF
--- a/asv/plugins/git.py
+++ b/asv/plugins/git.py
@@ -15,12 +15,15 @@ from .. import util
 
 class Git(Repo):
     dvcs = "git"
-    _default_branch = "master"
 
     def __init__(self, url, mirror_path):
         self._git = util.which("git")
         self._path = os.path.abspath(mirror_path)
         self._pulled = False
+        # default branch
+        self._default_branch = self._run_git(['config',
+                                              'init.defaultBranch'],
+                                             cwd=None)
 
         if self.is_local_repo(url):
             # Local repository, no need for mirror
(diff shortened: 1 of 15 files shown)


asv/plugins/git.py
```diff
@@ -15,12 +15,19 @@
...
 
     def __init__(self, url, mirror_path):
         self._git = util.which("git")
         self._path = os.path.abspath(mirror_path)
         self._pulled = False
+        # default branch
+        try:
+            self._default_branch = self._run_git(['config',
+                                                  'init.defaultBranch',
+                                                  ], display_error=False,
```
HaoZeke, line 27: Without `display_error=False`, tests which capture `stderr` and `stdout` to test against reference strings fails.

rgommers:
A late comment now that I'm seeing this: conceptually this isn't quite right. `git config` reads first repo-local and then global config. Repo-local config is usually empty, so it falls back to global. I just tried in a repo which still uses `master`, and am (as expected) getting `main` as the answer from `git config init.defaultBranch`. To fix it, `.git/config` needs to have:
```
[init]
	defaultBranch = main
```
I'd think it would be a lot easier to just add the config value to `asv.conf.json`.

PR #970, opened 2021-06-23
add support for mamba

jaimergp:
Trying to add support for `mamba` envs. Closes #962.

Should `mamba` be part of the tool autodiscovery mechanism specified in [`pythons`](https://asv.readthedocs.io/en/stable/asv.conf.json.html#pythons)? It's not clear to me how one can prioritize the env type [here](https://github.com/airspeed-velocity/asv/blob/master/asv/environment.py#L404).

Issue #908, opened 2020-02-24
REF: install asv in the profiled environment


jbrockmendel:
Having asv install itself in the profiled environment would open up a lot of options.  In particular, I think it would be necessary for something like #482 to be implemented.

I'd like to poke at this and see what it makes possible, but my first couple attempts at injecting asv into the environment failed.  @pv is there an easy way to do this?

HaoZeke:
Not required anymore. `asv_runner` is installed per environment, and includes decorators.

PR #1234, opened 2023-01-03
MAINT: Fixup and remove bundled dependencies

HaoZeke:
This supersedes https://github.com/airspeed-velocity/asv/pull/1225 and https://github.com/airspeed-velocity/asv/pull/1211.

Also closes https://github.com/airspeed-velocity/asv/issues/1139.

cc @mattip


HaoZeke:
These pass locally, will debug soon.

mattip:
This could use a refresh

asv/benchmark.py
```diff
@@ -704,16 +705,22 @@ def __init__(self, name, func, attr_sources):
...
-        # the asv package in the benchmarking process.
-        path = os.path.join(
-            os.path.dirname(__file__), 'extern', 'asizeof.py')
-        asizeof = importlib.machinery.SourceFileLoader('asizeof', path).load_module()
+        if ON_PYPY:
+            raise NotImplementedError("asizeof doesn't work on pypy")
+            return
+
+        try:
+            subprocess.check_output(['python', "-mpip", "install", 'pympler'])
```
HaoZeke, line 713: This works with all three environment backends, `conda`, `mamba` (as of #970) and `virtualenv` but is syntactically less than ideal. A better approach would be to modify the constraints of the `environment` class which is the base of all the backends.. Though YAGNI probably.

mattip:
After the move from setup.cfg to pyproject.toml this now has conflicts.

HaoZeke:
This seems to run locally on machines I test, will look into the build failure in more detail, it seems like `pip` cannot install anything on the CI...

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         run: pip install .
 
       - name: Run tests
-        run: python -m pytest -v --timeout=1800 --durations=20 test
+        run: python -m pytest -v --timeout=1800 --durations=20 test -k test_runner
```
HaoZeke, line 40: This is a temporary change to fail fast when the `mem` benchmark doesn't work.

asv/benchmark.py
```diff
@@ -704,16 +706,35 @@ def __init__(self, name, func, attr_sources):
...
+                if os.path.isfile(python_path) and os.access(python_path, os.X_OK):
+                    cand_path = Path(python_path).parent.parent / "lib"
+                    asizeof_paths = [x for x in cand_path.rglob("asizeof.py")]
+                    if len(asizeof_paths) > 0:
+                        # FIXME: This will return the first path found
+                        asizeof = importlib.machinery.SourceFileLoader(
+                            "asizeof", str(asizeof_paths[0])
+                        ).load_module()
+                        return asizeof
+            return NotImplementedError("asizeof not found anywhere")
```
HaoZeke, line 726: This is a horrible hack, mostly because of persistent CI failures. A much nicer approach is to ensure either that the `virtualenv`, `mamba`, and `conda` environments always adds `pympler`, or even cleaner and less obtrusively, the approach in https://github.com/airspeed-velocity/asv/pull/1234/commits/c1245ccc34e3bec18c2cbb041402e6de414ce96e

Basically:
```python
        try:
            from pympler.asizeof import asizeof
        except ImportError:
            try:
                subprocess.run([sys.executable, "-m", "pip", "install", "pympler"], check=True)
                from pympler.asizeof import asizeof
            except (subprocess.CalledProcessError, ImportError) as e:
                raise NotImplementedError(f"Failed to install pympler: {e}")
```

Which works locally but fails on the CI persistently with `versions none` from `pip install`.

HaoZeke:
Not sure about the AppVeyor build, the environment needs to have both `json5` and `pympler` installed now.

asv/benchmark.py
```diff
@@ -704,16 +707,64 @@ def __init__(self, name, func, attr_sources):
...
-        # the asv package in the benchmarking process.
-        path = os.path.join(
-            os.path.dirname(__file__), 'extern', 'asizeof.py')
-        asizeof = importlib.machinery.SourceFileLoader('asizeof', path).load_module()
+        if ON_PYPY:
+            raise NotImplementedError("asizeof doesn't work on pypy")
+            return
+
+        def import_asizeof():
+            """Import asizeof, searching system Pythons in PATH."""
```
mattip, line 715: Why do we need all this? If the import fails, can't we just fail the run?

HaoZeke: The problem here is that the import fails where it shouldn't, e.g. on the CI. This seems to be because of the process isolation (each ASV run is in a different subprocess). Perhaps, now that I'm thinking about it, it would be cleaner to instead import `asv` itself, which would automatically bring in `pympler` and is discussed in https://github.com/airspeed-velocity/asv/issues/908

HaoZeke: If I went with that approach then the Import Mechanism would instead load ASV rather than `asizeof`, but I'm not sure how to simplify this further without breaking the subprocess approach.

mattip:
I am fine to merge this as-is and then to iterate on solving the import problem, what do you think?

HaoZeke:
> I am fine to merge this as-is and then to iterate on solving the import problem, what do you think?

Sure that sounds great.

PR #1237, opened 2023-02-11
MAINT: Cleanup

HaoZeke:
This ensures that `pytest.mark.flaky` works, and includes the refactor to cleanup the `git` changes a bit, addressing https://github.com/airspeed-velocity/asv/pull/1227#discussion_r1060189402

PR #1238, opened 2023-02-12
ENH: Add a `mamba` plugin

### LINKED ISSUE #962: Ability to use mamba instead of conda for environment creation
For more complex environments (eg in the pandas asv suite), the step of creating the environment can take quite a long time with conda. Using mamba could speed up this step.

HaoZeke:
This is much faster than the `conda` solver and is more robust with a better Pythonic API. Closes #962.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.7", "3.10", "=3.8[build=*_pypy]", "=3.9[build=*_pypy]"]
+        python-version: ["3.7", "3.10", "pypy-3.7", "pypy-3.8"]
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v3
@@ -24,27 +24,21 @@ jobs:
         with:
           python-version: 3.7
 
-      - name: Setup Conda with ${{ matrix.python-version }}
-        uses: conda-incubator/setup-miniconda@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-          channels: conda-forge
+          cache: pip
+          cache-dependency-path: pyproject.toml
 
       - name: Install dependencies
-        shell: bash -el {0}
-        run: pip install -r requirements-dev.txt
-
-      - name: Get mamba
-        shell: bash -el {0}
-        run: conda install mamba -c conda-forge
+        run: python -m pip install -r requirements-dev.txt
 
       - name: Install asv
-        shell: bash -el {0}
         run: pip install .
 
       - name: Run tests
-        shell: bash -el {0}
-        run: python -m pytest -v --timeout=300 --durations=20 test
+        run: python -m pytest -v --timeout=300 --durations=100 test
 
   docs:
     runs-on: ubuntu-latest
@@ -57,23 +51,11 @@ jobs:
           cache: pip
           cache-dependency-path: pyproject.toml
 
-      - name: Setup Conda
-        uses: conda-incubator/setup-miniconda@v2
-        with:
-          channels: conda-forge
-
-      - name: Get mamba
-        shell: bash -el {0}
-        run: conda install mamba -c conda-forge
-
       - name: Install dependencies
-        shell: bash -el {0}
         run: python -m pip install -r requirements-dev.txt
 
       - name: Install asv
-        shell: bash -el {0}
         run: pip install ".[doc]"
 
       - name: Build docs
-        shell: bash -el {0}
         run: sphinx-build -W docs/source html
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.7", "3.10", "=3.7[build=*_pypy]", "=3.8[build=*_pypy]"]
+        python-version: ["3.7", "3.10", "=3.8[build=*_pypy]", "=3.9[build=*_pypy]"]
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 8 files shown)


HaoZeke:
Expects #1237

zacharyburnett:
@HaoZeke nice PR, this should speed up the benchmarking process! You can also try using the https://github.com/marketplace/actions/provision-with-micromamba action in the CI to make it even faster

HaoZeke:
> @HaoZeke nice PR, this should speed up the benchmarking process! You can also try using the https://github.com/marketplace/actions/provision-with-micromamba action to make the CI even faster

Yup, however, I think I'd need to add another plugin for `micromamba` since it isn't covered by the `python` API used here.

HaoZeke:
Installing dependencies on `pypy-3.7` leads to an obscure error:

```
Installing collected packages: sphinx_bootstrap_theme, sortedcontainers, snowballstemmer, sgmllib3k, pytz, python-hglib, mccabe, distlib, charset-normalizer, zipp, urllib3, typing-extensions, tomli, sphinxcontrib-serializinghtml, sphinxcontrib-qthelp, sphinxcontrib-jsmath, sphinxcontrib-htmlhelp, sphinxcontrib-devhelp, sphinxcontrib-applehelp, sniffio, pyyaml, PySocks, Pygments, pyflakes, pycodestyle, packaging, numpy, MarkupSafe, isort, iniconfig, imagesize, idna, filelock, feedparser, execnet, exceptiongroup, docutils, certifi, babel, attrs, async-generator, alabaster, requests, platformdirs, outcome, Jinja2, importlib-metadata, h11, wsproto, virtualenv, trio, sphinx, pluggy, flake8, trio-websocket, pytest, selenium, pytest-xdist, pytest-timeout, pytest-rerunfailures
ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/usr/share/miniconda/envs/test/site-packages/../bin/rst2html.py'

Error: Process completed with exit code 1.
```

Dropping support for `3.7` in general might be a good idea? Especially since newer plugins (e.g. `mamba`, `memray`) require newer versions as well.

asv/environment.py
```diff
@@ -428,18 +438,24 @@ def get_environment_class(conf, python):
         if cls.matches_python_fallback and cls.matches(python):
             return cls
     raise EnvironmentUnavailable(
-        "No way to create environment for python='{0}'".format(python))
+        "No way to create environment for python='{0}'".format(python)
+    )
```
mattip, line 442: f-strings might be nicer

asv/environment.py
```diff
@@ -428,18 +438,24 @@ def get_environment_class(conf, python):
...
 
 def get_environment_class_by_name(environment_type):
     """
     Find the environment class with the given name.
     """
     for cls in util.iter_subclasses(Environment):
+        if "mamba" in environment_type and (
+            sys.version_info.major >= 3 and sys.version_info.minor < 8
+        ):
+            environment_type = "conda"
```
mattip, line 453: why the upper limit? Also: what happens if "conda" is used instead of "mamba"?

mattip (commented):
It is much easier to review if the formatting changes are done in a separate PR, and not mixed in with code changes.

Rather than hard-code `pip` to install the project, shouldn't there be a way for the project to say what backend to use?

.github/workflows/ci.yml
```diff
@@ -23,20 +23,26 @@ jobs:
         with:
           python-version: 3.7
 
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+      - name: Setup Conda with ${{ matrix.python-version }}
+        uses: conda-incubator/setup-miniconda@v2
```
fangchenli, line 27: Setting up conda and memba in the CI takes longer than using `setup-python`. For a library with very few dependencies, it's not worth it. Also, the CI changes are unrelated to the topic of this PR and should be addressed separately.

fangchenli (changes requested):

.github/workflows/ci.yml
HaoZeke: Makes sense, originally I thought of testing the new environment handler in CI as well, which is why I wanted to add it here, but I guess that could be elsewhere.

HaoZeke:
Some benchmarks for cold starts `time asv run` with `.asv` removed:

```bash
# Conda
asv run  63.61s user 4.06s system 101% cpu 1:06.50 total
# Mamba
asv run  13.42s user 2.43s system 120% cpu 13.152 total
# Virtualenv
asv run  8.35s user 1.83s system 116% cpu 8.751 total
```

For this [toy project](https://github.com:HaoZeke/asv_experiments)

HaoZeke:
> It is much easier to review if the formatting changes are done in a separate PR, and not mixed in with code changes.
Sorry, reverted those.

> Rather than hard-code `pip` to install the project, shouldn't there be a way for the project to say what backend to use?

The project is installed with the `install_command` configuration option, the plugin only uses `pip` for the dependencies..

asv/environment.py
HaoZeke: If `conda` is used it continues to use the old solver. The idea here was that it should silently fall back to the `conda` mechanism before `libmambapy` is present. The upper limit shouldn't be there, sorry.

HaoZeke:
I have cleaned this up and simplified it. There are no tests, other than manual ones. Tests would require changes to the CI which can be done elsewhere.

asv/environment.py
```diff
@@ -428,7 +428,7 @@ def get_environment_class(conf, python):
         if cls.matches_python_fallback and cls.matches(python):
             return cls
     raise EnvironmentUnavailable(
-        "No way to create environment for python='{0}'".format(python))
+        "No way to create environment for python='{python}'")
```
zacharyburnett, line 431: does this need an `f`? You could use `flynt` or something to do these automatically (maybe out of scope for this PR)

HaoZeke: Yup, good eye. Also I'll open another PR for `flynt` but I think @jarrodmillman has converted a lot of it already.

mattip:
Thanks @HaoZeke

PR #1244, opened 2023-02-20
Remove Travis CI conf

jarrodmillman:
This seems to be cruft.


mattip:
Indeed. There was a webhook for travis.org, which migrated to travis.com two years ago and was not updated. Thanks.

PR #1246, opened 2023-02-26
ENH: Add a memray based peak memory tracker

HaoZeke:
Note that for a simple function like this:

```python
class TimeSuite:
    def ray_list(self):
        return [0] * 256

    def peakmem_list(self):
        return [0] * 256

    def mem_list(self):
        return [0] * 256
```

The times are a bit, odd:

```bash
[58.33%] ··· benchmarks.TimeSuite.ray_list                                                          2.05k
[66.67%] ··· benchmarks.TimeSuite.mem_list                                                          2.1k
[75.00%] ··· benchmarks.TimeSuite.peakmem_list                                                     15.6M
```

Though other metrics can also be extracted / returned from `memray`, that would require some more work in terms of the frontend which needs to track them sa well. E.g. the number of allocations, and top 5 functions (in terms of memory)

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -64,6 +64,7 @@ from hashlib import sha256
 from importlib import import_module
 from collections import Counter
 from time import process_time
+from pathlib import Path
 
 wall_timer = timeit.default_timer
 
@@ -771,6 +772,40 @@ class MemBenchmark(Benchmark):
         return sizeofcopy - sizeof2
 
 
+class RayMemBenchmark(Benchmark):
+    """
+    A single benchmark for tracking the memory of an object with memray
+    """
+
+    name_regex = re.compile("^(Ray[A-Z_].+)|(ray_.+)$")
+
+    def __init__(self, name, func, attr_sources):
+        Benchmark.__init__(self, name, func, attr_sources)
+        self.type = "memory"
+        self.unit = "bytes"
+        pass
+
+    def run(self, *param):
+        try:
+            import memray
+            from memray import FileReader
+        except ImportError:
+            subprocess.check_output([sys.executable, "-mpip", "install", "memray"])
+            import memray
+            from memray import FileReader
+        import uuid
+
+        u_id = uuid.uuid4()
+        temp_dir = tempfile.gettempdir()
+        tfile_loc = Path(f"{temp_dir}/{u_id}.bin")
+        with memray.Tracker(
+            destination=memray.FileDestination(tfile_loc, overwrite=True)
+        ):
+            self.func(*param)
+        freader = FileReader(str(tfile_loc))
+        return freader.metadata.peak_memory
+
+
 class PeakMemBenchmark(Benchmark):
     """
     Represents a single benchmark for tracking the peak memory consumption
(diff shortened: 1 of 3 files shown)


asv/benchmark.py
```diff
@@ -718,6 +719,40 @@ def run(self, *param):
...
+        pass
+
+    def run(self, *param):
+        try:
+            import memray
+            from memray import FileReader
+        except ImportError:
+            subprocess.check_output([sys.executable, "-mpip", "install", "memray"])
+            import memray
+            from memray import FileReader
```
fangchenli, line 742: We need to add `memray` as a runtime or optional dependency. Running `pip install` during runtime creates an undesired side effect.

fangchenli (changes requested):

asv/benchmark.py
HaoZeke: This was implemented in this manner so `memray` is part of each environment, since `asv` doesn't allow importing modules (including `asv` itself) within this `benchmark.py` file.

fangchenli: > This was implemented in this manner so `memray` is part of each environment, since `asv` doesn't allow importing modules (including `asv` itself) within this `benchmark.py` file.

I guess this was the reason `asv` initial vendored the `sizeof` module.

HaoZeke: Yup, but it wasn't a very flexible design. The `pytest` approach to this (collection / isolation) is probably something worth looking into more.

asv/plugins/benchmarks/asvmemray.py
```diff
@@ -0,0 +1,299 @@
...
+            if hasattr(method_caller, 'func_code'):
+                code = method_caller.func_code
+            else:
+                code = method_caller.__code__
+
+            self.redo_setup()
+
+            profile.runctx(
+                code, {'run': self.func, 'params': self._current_params},
+                {}, filename)
```
HaoZeke, line 271: This is the minimal amount of code needed from `benchmark.py` in order to write "external" plugins. @mattip perhaps this could be spun out into a separate pure-python package? If it has no dependencies other than simple `python` like `benchmark.py` it shouldn't cause any additional overhead in installations.

mattip: > this could be spun out into a separate pure-python package?

Do you mean create a `asv-memray` package and put it on PyPI, so it can be pip-installed? Sounds good to me, but maybe there are additional opinions: @pv, @datapythonista ?

HaoZeke: Yup I could do that, but also I was thinking of a bit of a broader refactor:
- Spin out the pure-python part (highlighted) into something like `asv-core` or `asv-bench`
- Then use `asv-core` in both `asv` and `asv-memray`

The goal is to separate out the "pure python" part of the benchmark into its own package which can be installed without causing dependency issues with other packages, while still letting plugins (and `asv`) have other dependencies.

HaoZeke:
This is now deprecated by https://github.com/airspeed-velocity/asv_runner/pull/36. Probably a minor documentation update will supercede this.

mattip:
closing, please reopen if it is still needed

PR #1248, opened 2023-02-26
MAINT: Switch to using the `virtualenv` API

HaoZeke:
What it says on the tin. Might be a reasonable cleanup.

### DIFF
--- a/asv/plugins/virtualenv.py
+++ b/asv/plugins/virtualenv.py
@@ -126,16 +126,19 @@ class Virtualenv(environment.Environment):
         Then, all of the requirements are installed into
         it using `pip install`.
         """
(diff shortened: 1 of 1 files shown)


asv/plugins/virtualenv.py
```diff
@@ -126,17 +127,21 @@ def _setup(self):
         Then, all of the requirements are installed into
         it using `pip install`.
         """
+        import virtualenv
```
HaoZeke, line 130: This plugin actually has a try: except: block which would fail without `virtualenv` being present, so it makes sense to use it.

HaoZeke:
I don't actually see any use for this. The API itself seems a bit of a WIP too.

PR #1251, opened 2023-03-05
Lint all things

HaoZeke:
Addresses https://github.com/airspeed-velocity/asv/pull/1238#pullrequestreview-1314914872 in an opinionated way. The settings are chosen to prevent conflicts between the tools, and the line length was also as before.

`black` conflicts far too often with `isort`, so `yapf` is more reasonable (configurable).


HaoZeke:
@mattip perhaps the one-time code churn would be OK? Or at some point might be a good plan.

PR #1252, opened 2023-03-05
BUG: Cleanup and use git lookup order

### LINKED ISSUE #1250: Buggy git-config handling
A late comment now that I'm seeing this: conceptually this isn't quite right. `git config` reads first repo-local and then global config. Repo-local config is usually empty, so it falls back to global. I just tried in a repo which still uses `master`, and am (as expected) getting `main` as the answer from `git config init.defaultBranch`. To fix it, `.git/config` needs to have:
```
[init]
	defaultBranch = main
```
I'd think it would be a lot easier to just add the config value to `asv.conf.json`.

_Originally posted by @rgommers in https://github.com/airspeed-velocity/asv/issues/1232#issuecomment-1453531290_

HaoZeke:
Closes #1250.

### DIFF
--- a/asv/util.py
+++ b/asv/util.py
@@ -1383,9 +1383,23 @@ except ImportError:
 
 def git_default_branch():
     try:
-        default_branch = check_output([which('git'),
-                                       'config', 'init.defaultBranch'],
-                                      display_error=False).strip()
+        # Local name gets precedence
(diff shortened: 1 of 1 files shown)


mattip:
Thanks @HaoZeke

Issue #1219, opened 2022-10-06
Improving Python benchmarking tooling


asmeurer:
At the NumFOCUS summit a few weeks ago, several people had a conversation about the limitations of asv and the other Python benchmarking tooling. I'd like to use this issue to kick off the discussion about this so that we can gather a wishlist and a list of problems with the existing tooling, and figure out a plan on how to improve things. It's not clear yet whether that will mean improving asv, some other tool, or creating new tools.

Some people who I want to make sure are involved in this discussion (please feel free to CC others):

@jreback
@oscarbenjamin
@wkerzendorf
@jarrodmillman

asmeurer:
Here's my own wishlist/list of concerns with asv:

My biggest issue with asv is that it is very tied to one specific use case, namely, running a suite of benchmarks across a history of commits for a project, and analysing the history of runtimes. When you stay within asv's designed use-case, it works very well. However, it is very difficult to do anything that strays even a little bit from that use case. Examples of things that are difficult (or even impossible) to do currently with asv:

- Run just a single benchmark against `master`. This requires some confusing command line flags.
- Run benchmarks against a development piece of code. i.e., you change some code and want to see how it affects things. Typically if I want to do this I end up just copying the benchmark code and pasting it into an IPython `%timeit`. 
  
  This is hard because asv is designed to download the code from a remote repository (which is hard coded in a config file), and install everything into an isolated virtual environment. It doesn't have a concept of "running a benchmark locally". At best, you can manually edit the config file (which usually is a checked-in file), but it's very annoying. 

  In general, asv has a specific workflow on how it runs the benchmarks (download the code, create a virtualenv, etc.), and there's virtually no way to run benchmarks that doesn't fit into this paradigm.
- It's not possible to run the `setup()` function per benchmark (https://github.com/airspeed-velocity/asv/issues/966). More generally, there's limited ways that you can configure how the workflow of `setup()`, `run_benchmark()`, `teardown()` happen. 
- The virtual environment installation is limited in its configurability. In another project I work on at Quansight, I have a [very hacky](https://github.com/deshaw/versioned-hdf5/blob/master/benchmarks_install.py) script that overwrites how the virtualenv installation happens, because I have a dependency that must have a specific version installed depending on what commit is used. 
[...]

Issue #808, opened 2019-04-07
Better documentation how to run "asv profile" with parameterized method


mitar:
In fact, something like that does not work:

```
asv profile --config asv.conf.json 'metadata.Dataset.time_dataset_load(50000)' HEAD --gui snakeviz
```

But this does:

```
asv profile --config asv.conf.json 'metadata.Dataset.time_dataset_load\(50000\)' HEAD --gui snakeviz
```

Because method name is passes as a regex, so you have to escape parenthesis. Documentation (command line help hint) does not tell that and even simply uses unescaped parenthesis in the example.

Issue #801, opened 2019-04-02
running profile without commit results in obscure  AttributeError: 'Results' object has no attribute 'env'


yarikoptic:
```shell
$> asv profile  api.testds.time_create_test_dataset2x2      
Traceback (most recent call last):
  File "/home/yoh/proj/datalad/datalad/venvs/dev/bin/asv", line 11, in <module>
    load_entry_point('asv', 'console_scripts', 'asv')()
  File "/home/yoh/proj/misc/asv/asv/main.py", line 38, in main
    result = args.func(args)
  File "/home/yoh/proj/misc/asv/asv/commands/__init__.py", line 49, in run_from_args
    return cls.run_from_conf_args(conf, args)
  File "/home/yoh/proj/misc/asv/asv/commands/profiling.py", line 94, in run_from_conf_args
    **kwargs)
  File "/home/yoh/proj/misc/asv/asv/commands/profiling.py", line 148, in run
    for env in environments)
  File "/home/yoh/proj/misc/asv/asv/commands/profiling.py", line 148, in <genexpr>
    for env in environments)
AttributeError: 'Results' object has no attribute 'env'
```
and runs ok if I give HEAD
```shell
$> asv profile api.testds.time_create_test_dataset2x2 HEAD
· Discovering benchmarks
·· Uninstalling from virtualenv-py2.7.
·· Building ed8136f8 <bm-monkey-patch-is-interactive> for virtualenv-py2.7.
·· Installing ed8136f8 <bm-monkey-patch-is-interactive> into virtualenv-py2.7.
· Profile data does not already exist. Running profiler now.
·· Benchmarking virtualenv-py2.7
··· Running (api.testds.time_create_test_dataset2x2--).
··· api.testds.time_create_test_dataset2x2         
...
```

```shell
$> asv --version
asv 0.4.dev1608+7afea8e5
```

jbrockmendel:
I'm getting the same error, but specifying HEAD doesn't help

HaoZeke:
Can't reproduce this on `master`.

brandonwillard:
I'm seeing this issue with `asv` at 0.6.3 and `asv_runner` 0.2.1.  

Looking at the `asv.results.Results` class and I don't see where `Results` gets a `.env` field, nor do I see it referenced or used anywhere in `asv` aside from the `asv.profiling.Profile.run` (i.e. where the error occurs).

Changing `result.env.name` to `result.env_name` works until `result.env.checkout_project` is needed.

Issue #71, opened 2014-01-27
Create a way to download profiles


mdboom:
It's non-trivial, since profile files are not portable -- they contain paths to source files specific to a particular machine.

PR #1253, opened 2023-03-19
ENH: Return pstats.Stats for --profile

### LINKED ISSUE #971: Question: How to use/extract profiling data after running 'asv run -p'
First of all thanks for the awesome tool ❤️ 

I'm wondering how to access the profiling data recorded when using `asv run` with the `-p` flag.
I can't find profiling anywhere in the generated HTML page (`asv publish`), nor can I find a config option or flag to add it to the HTML page (maybe I'm just blind 😅 ). 

So I'm wondering what to do after recording the profiling data.

I saw that the result json files have a key `profiles` containing benchmark names as keys and a massive text blob as value (`base64` of the `pstat` binary blob?).
So is the intended use to parse the result json file, decode the blob and then process it with a tool like `gprof2dot`?

Thanks in advance for clarifying

HaoZeke:
This, along with documentation changes chronicles how to use the results of `asv run --profile`. Closes #971, #808, #801, #71.

Addresses (7) of https://github.com/airspeed-velocity/asv/issues/1219#issuecomment-1270968475.

### DIFF
--- a/asv/commands/profiling.py
+++ b/asv/commands/profiling.py
@@ -46,7 +46,7 @@ class Profile(Command):
             help="""The benchmark to profile.  Must be a
             fully-specified benchmark name. For parameterized benchmark, it
             must include the parameter combination to use, e.g.:
-            benchmark_name\(param0, param1, ...\)""")
+            benchmark_name\\(param0, param1, ...\\)""")
         parser.add_argument(
             'revision', nargs='?',
             help="""The revision of the project to profile.  May be a
--- a/asv/commands/profiling.py
+++ b/asv/commands/profiling.py
@@ -214,5 +214,6 @@ class Profile(Command):
             color_print('')
             with temp_profile(profile_data) as profile_path:
                 stats = pstats.Stats(profile_path)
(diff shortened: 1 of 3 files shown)


HaoZeke:
@mattip I think this should be good to go as well.

mattip:
Thanks @HaoZeke

mattip:
Any thoughts about why this might have [timed out on PyPy3.7](https://github.com/airspeed-velocity/asv/actions/runs/4775650402/jobs/8490113288#step:7:496)? I think we should drop that version from CI since it is no longer supported. We could add PyPy3.9 instead...

HaoZeke:
> Any thoughts about why this might have [timed out on PyPy3.7](https://github.com/airspeed-velocity/asv/actions/runs/4775650402/jobs/8490113288#step:7:496)? I think we should drop that version from CI since it is no longer supported. We could add PyPy3.9 instead...

I'm not sure, the PyPy tests are pretty flakey though, especially 3.7, I've opened https://github.com/airspeed-velocity/asv/pull/1276 to switch to 3.9.

PR #1245, opened 2023-02-20
Use fstrings

jarrodmillman:

PR #1197, opened 2022-03-31
Replace format() function with f-strings on asv/plugins/conda.py

LucyJimenez:
xref https://github.com/airspeed-velocity/asv/issues/1077
Closes https://github.com/airspeed-velocity/asv/issues/1172

Change f-strings instead `.format()` on `asv/plugins/conda.py`

PR #1196, opened 2022-03-30
Replace format() function with f-strings on asv/environment.py

LucyJimenez:
xref https://github.com/airspeed-velocity/asv/issues/1077
Closes https://github.com/airspeed-velocity/asv/issues/1178

Change f-strings instead `.format()` on `asv/environment.py`

PR #1258, opened 2023-03-25
MAINT: Convert almost everything to f-strings

HaoZeke:
With `flynt`. Superceedes https://github.com/airspeed-velocity/asv/pull/1198, https://github.com/airspeed-velocity/asv/pull/1197, https://github.com/airspeed-velocity/asv/pull/1196, https://github.com/airspeed-velocity/asv/pull/1245.

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -364,12 +364,12 @@ class Benchmark:
         try:
             self.param_names = [str(x) for x in list(self.param_names)]
         except ValueError:
-            raise ValueError("%s.param_names is not a list of strings" % (name,))
+            raise ValueError(f"{name}.param_names is not a list of strings")
 
         try:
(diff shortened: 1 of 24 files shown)


mattip:
This has a merge conflict

mattip:
Does this take into account the comments in #1245?

HaoZeke:
> Does this take into account the comments in #1245?

Yup, most of them are not relevant since `extern` has been removed. However, for this:

> Nit: it is a bit weird to be using these old versions of python (line 39?) and numpy (dunno where that comes from). If not too much trouble could you update them? If not no big deal.

I'm not sure they should be updated here, since those are part of the `asv` files generated for the tests, might be good to have an issue.

mattip:
Thanks @HaoZeke

PR #1259, opened 2023-03-25
use `tox` to build test environments and `OpenAstronomy` reusable workflow

zacharyburnett:


HaoZeke:
Is there an issue for this? I'm not very familiar with `tox` and it feels like quite a bit of complication. I was thinking at some point of `nox` for the environment management in #1293 but I'm not sure the CI needs to be reworked with this? @mattip any suggestions?

mattip:
Personally, I find tox confusing and the additional layer over testing problematic. Closing. Please reopen with some further discussion if you wish to pursue this.

PR #1260, opened 2023-03-25
CI: use `ruff` for linting instead of `flake8`

zacharyburnett:
replace `flake8` with `ruff`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,7 +97,6 @@ extend-ignore = [
     "E741", # Do not use variables named 'I', 'O', or 'l'
 ]
 extend-exclude = [
-    "asv/extern",
     "test/example_results/cheetah",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ doc = [
     "sphinx_bootstrap_theme",
 ]
 dev = [
-    "flake8",
+    "ruff",
     "isort >= 5.11.5",
 ]
 hg = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -85,6 +85,16 @@ namespaces = false
 [tool.pytest.ini_options]
 addopts = "-p no:logging"
 
+[tool.ruff]
+line-length = 99
+extend-ignore = [
+    "E741", # Do not use variables named 'I', 'O', or 'l'
+]
+extend-exclude = [
+    "asv/extern",
```
HaoZeke, line 94: This shouldn't be needed, `external` isn't present anymore.

HaoZeke (commented):
LGTM, minor nit aside. Does this conflict with `isort` and the other linters? Have you checked that `ruff` can used on all existing files without conflicting with `isort` or the other tools?

zacharyburnett:
> LGTM, minor nit aside. Does this conflict with `isort` and the other linters? Have you checked that `ruff` can used on all existing files without conflicting with `isort` or the other tools?

`ruff` actually has `isort` built into it, though it is disabled by default unless selected. `ruff` doesn't currently implement the `only_sections` parameter however, so kept the `isort` check and it seems to pass the `pre-commit` with no changes.

HaoZeke (approved):
LGTM, the appveyor seems to be unrelated, could you close/open this so CI runs again?

pyproject.toml
zacharyburnett, line 94:
```suggestion
```

zacharyburnett:
You could also keep `.flake8` to ensure backwards compatibility with editors that look for it; however, it would require that any changes to the `ruff` config be manually mirrored to `.flake8`

mattip:
I think we should choose one tool as "the way" to format code. The project does not have enough contributors to make installing ruff vs. installing flake8 a big issue. I might have preferred not to touch this at all, but since @zacharyburnett made the effort, let's give it a try.

mattip:
Thanks @zacharyburnett and @HaoZeke

PR #1263, opened 2023-04-02
ENH: Add a markdown renderer for asv compare

### LINKED ISSUE #1210: FEAT: add `--format=markdown` option to `asv compare`
In the [dedupe library](https://github.com/dedupeio/dedupe), we made [a bot](https://github.com/dedupeio/dedupe/blob/main/.github/workflows/benchmark-bot.yml) that, when you comment `@benchmark` on a PR, will run ASV and add a comment with the results, so you can see how your PR affects performance. See [this PR](https://github.com/dedupeio/dedupe/pull/1017) for an example of how that looks.

Currently, we post the raw output from `asv compare`, but as you can see from the example above, it doesn't look that great. It would be awesome if `asv compare` supported a `--format=markdown` option to use the markdown table format. I imagine (no evidence) that asv is used a lot in CI, and therefore its output posted to markdown-supporting UIs quite frequently.

Other options/thoughts:
1. Other option is to just make markdown format the default, don't add a flag. The current formatting isn't machine-parseable at all, so maybe just replace it. Since its' not machine readable currently, I don't think this would break anyone. This might end up being uglier for basic CLI users though.
2. This API would allow for other formats in the future like RST (`--format=rst`)
3. Possibly take [git's formatting API](https://devhints.io/git-log-format) into consideration, to make it even more tuneable. Probably overkill though.
4. Are there other asv commands besides `compare` that we need to consider/support? I'm just getting into asv so not that familiar yet.

Thanks!

HaoZeke:
Closes #1210. I think this would be a reasonable change throughout, for all the outputs. None of the current outputs are machine readable.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -25,7 +25,7 @@ install:
   - conda update conda -y
   - conda install -q --yes python=%PYTHON_VERSION% conda pip pytest pytest-xdist pytest-timeout filelock selenium conda-build bzip2 pympler
   - python -m pip install -U pip
-  - python -m pip install pytest-rerunfailures json5 pympler
+  - python -m pip install pytest-rerunfailures json5 pympler tabulate
 
   # Check that we have the expected version of Python
   - python --version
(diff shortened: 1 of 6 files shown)


HaoZeke:
PyPy timeout is unrelated (randomly shows up, #1264). AppVeyor error needs a closer look.

HaoZeke:
This is ready, @mattip, and it would be nice to have more formatted tables elsewhere too perhaps :)

mattip:
Thanks @HaoZeke

PR #1265, opened 2023-04-03
ENH: Import asv into the benchmarking process

HaoZeke:
This needs some work still, but eventually opens up to more decisive changes, including:
- User defined benchmarks
- Newer parameterization approaches


HaoZeke:
Unnecessary now that we spun out `asv_runner`.

PR #1267, opened 2023-04-08
ENH: Add cleaner matches in conda/mamba

HaoZeke:
Testing by creating an environment is much slower than simply searching. This PR updates the `match` mechanism and the utility function might be useful elsewhere as well.


HaoZeke:
@mattip this would be nice to have (faster)

mattip:
Thanks @HaoZeke

PR #1272, opened 2023-04-23
BUG: Use `colorama` on windows

### LINKED ISSUE #546: setting machine info doesn't work with Git Bash
With asv 0.2, for all of ``asv`` ``dev/run/machine`` the output when run from Git Bash on Windows 10 is:

    $ asv dev
    ▒ No information stored about machine 'C25302'. I know about nothing.

    ▒ Run asv at the console the first time to generate one.

It does work in PowerShell, and on Git Bash it can be worked around with:

    $ winpty asv dev

The [if not sys.stdout.isatty():](https://github.com/spacetelescope/asv/blob/011815d287456efee0a24a9efe5a2c0c49b720f2/asv/machine.py#L151) check seems overly strict.

HaoZeke:
The documentation mentions using `colorama` on Windows. This closes #546 , and essentially tracks what `ipython` does in https://github.com/ipython/ipython/commit/0a16e9d0e624a99578587eaa288ab1b6a3836a98. It seems reasonable to put it in the `Log` class constructor, since that's a singleton in the code.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,10 @@ dynamic = [
     "version",
 ]
 dependencies = [
-  "colorama; os_name == 'nt'",
+    "json5",
+    "tabulate",
+    "colorama; os_name == 'nt'",
+    "pympler; platform_python_implementation != \"PyPy\"",
 ]
 
 [project.urls]
@@ -49,8 +52,6 @@ test = [
     "selenium",
     "pytest-rerunfailures",
     "python-hglib",
-    "json5",
-    "pympler; platform_python_implementation != \"PyPy\"",
 ]
 doc = [
     "sphinx",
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -24,6 +24,12 @@ classifiers = [
 dynamic = [
     "version",
 ]
+dependencies = [
+    "json5",
+    "tabulate",
+    "colorama; os_name == 'nt'",
```
mattip, line 30: Interesting. I see  `os_name` and `platform_python_implementation` are part of the [grammer](https://peps.python.org/pep-0508/#grammar), but I am used to seeing them spelled as `sys_platform` (which would be `win32`) and `implementation_name` (which would be `!= "pypy").

mattip:
Thanks @HaoZeke

pyproject.toml
HaoZeke: I'm most used to the `sys_platform` one, but in the `asv` codebase there's a variable `WIN = os.name` so I thought I'd go with this here 😅

PR #1274, opened 2023-04-23
ENH: Add a gitignore template

### LINKED ISSUE #63: Add standard .gitignore to quickstart template

HaoZeke:
Closes #63. Maybe might be worthwhile adding `windows` and `macos` specific ignores to this?

### DIFF
--- a/asv/template/.gitignore
+++ b/asv/template/.gitignore
@@ -1,0 +1,162 @@
+#### python ####
+# From https://github.com/github/gitignore/blob/main/Python.gitignore
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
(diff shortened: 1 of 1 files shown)


mattip:
I guess we could do this. I somewhat prefer to encourage people to use global `.gitignore` files for stuff on their systems: ide projects, editor backups and the such, and only have project-specific ignore files. But I can see why we might want to add something minimal here.

HaoZeke:
> I guess we could do this. I somewhat prefer to encourage people to use global `.gitignore` files for stuff on their systems: ide projects, editor backups and the such, and only have project-specific ignore files. But I can see why we might want to add something minimal here.

The Jetbrains stuff seems weird, but I thought it'd be easiest to take the Github defaults for `python` projects, rather than hand-rolling something for `asv`. I can see that OS specific stuff shouldn't be here though.

PR #1275, opened 2023-04-23
ENH: Optionally add some kind of pickle replacement

HaoZeke:
Closes https://github.com/airspeed-velocity/asv/issues/347. I noticed this came up once before in a PR, #351, and the motivation wasn't discussed (its in the issue, essentially functions can be stored with `dill` over `pickle`).

In terms of `cloudpickle` (used by `dask`, discussed on the older PR), it seems like it has a slightly different use-case and this [StackOverflow post by the author](https://stackoverflow.com/a/32925601/1895378) seems to suggest the differences are minor if any, and `cloudpickle` [strongly discourages](https://pypi.org/project/cloudpickle/) usage for long term storage.

`dill` [doesn't support](https://dill.readthedocs.io/en/latest/index.html) the types `frame`, `generator`, `traceback`.

This is actually a problem, since we pickle `benchmark` objects which might as well have generators. Will switch to `cloudpickle`.

### DIFF
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -65,8 +65,7 @@ from collections import Counter
 from time import process_time
 
 try:
-    import dill as pickle
-    pickle.settings['recurse'] = True
+    import cloudpickle as pickle
 except ImportError:
     import pickle
 
--- a/asv/benchmark.py
+++ b/asv/benchmark.py
@@ -47,7 +47,6 @@ import inspect
 import itertools
 import json
 import os
-import pickle
 import re
 import subprocess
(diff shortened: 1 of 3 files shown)


mattip:
I am confused. Doesn't ASV expect long-term storage to work?

HaoZeke:
> I am confused. Doesn't ASV expect long-term storage to work?

Yeah, I guess this and the associated issues should be closed. For what `asv` needs `pickle` is exactly the right level of abstraction. We haven't seen any complaints against it either.

PR #1282, opened 2023-04-28
BUG: Fix poetry / pdm usage

### LINKED ISSUE #1281: BUG: Poetry doesn't have `--output`
I am getting an error when running with poetry and python 3.10, `The option "--output" does not exist`

_Originally posted by @DavidRConnell in https://github.com/airspeed-velocity/asv/pull/1261#discussion_r1178389218_

HaoZeke:
Closes #1281. Also implements handing `poetry` and `pdm` with a much cleaner approach.

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -504,7 +504,7 @@ class Environment:
         self._build_root = os.path.abspath(os.path.join(self._path, 'project'))
 
         if (Path.cwd() / "poetry.lock").exists():
-            self._requirements["poetry"] = ""
+            self._requirements["poetry-core"] = ""
 
         if (Path.cwd() / "pdm.lock").exists():
             self._requirements["pdm"] = ""
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -503,6 +503,7 @@ class Environment:
         self._cache = build_cache.BuildCache(conf, self._path)
         self._build_root = os.path.abspath(os.path.join(self._path, 'project'))
 
+        self._requirements = requirements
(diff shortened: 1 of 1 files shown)


DavidRConnell:
This is working for my poetry project. Thank you. Having pip handle the different backends like this is a nicer solution. Surprised pip doesn't automatically install the backend.

Should the requirement for poetry packages be `poetry-core` instead of `poetry` (line 507)? installing `poetry-core` is working for me as well and the [poetry-core README](https://github.com/python-poetry/poetry-core) indicates it is meant to replace the full poetry package for this purpose.

HaoZeke:
> This is working for my poetry project. Thank you. Having pip handle the different backends like this is a nicer solution. Surprised pip doesn't automatically install the backend.
 

Excellent, thanks for confirming and yeah it would make much more sense for `pip` to respect `build-backend`...


> Should the requirement for poetry packages be `poetry-core` instead of `poetry` (line 507)? installing `poetry-core` is working for me as well and the [poetry-core README](https://github.com/python-poetry/poetry-core) indicates it is meant to replace the full poetry package for this purpose.

It wasn't immediately clear to me that `poetry.core` is enough to handle all the additional fields `poetry` uses, e.g. I thought it would be ok for a `pyproject.toml` which just builds with `poetry` but not if there was heavy use of the keys under `[tool.poetry]`... I think it should be ok though, am updating, thanks for the heads up!

mattip:
Thanks @HaoZeke

Issue #1286, opened 2023-05-06
RFC: Factoring out `asv.benchmark`


HaoZeke:
`asv` in its current design has an extension problem. `benchmark.py` contains all the code for benchmarks, and is the only part of `asv` imported into the benchmarking process, but this means that it is painful to derive new `benchmark` instances, especially when they rely on external packages (see for example, the convoluted logic for `asizeof` https://github.com/airspeed-velocity/asv/blob/15075b76557e01758e4a561a39e3fdfc78a4e1d3/asv/benchmark.py#L716 or the `memray` plugin https://github.com/airspeed-velocity/asv/pull/1246). Some of this used to be vendored under `extern` but this is not scalable.

There are two possible solutions:
- Import `asv` into the benchmarking process #1265 
- **Split out** `asv.benchmark` into its own module

As @mattip noted, it is a good thing that `asv` in the benchmarking phase introduces no dependencies other than enforcing the lowest supported `python`. `asv` already handles rather complicated environments, and adding itself into the resolution mechanism is not desirable.

Additionally, as can be seen in the `memray` plugin above, with the pure `python` module factored out, additional benchmark types can be written externally and distributed with `pip`. Users will then be able to make conscious choices about what they add to their dependency matrix.

I believe this is the best way forward. From the design of the code, `benchmark.py` anyway makes a strong effort to remove itself from the rest of `asv` with some nifty `sys.path` hacking, none of which would be required if it is was a pure Python module distributed separately.

## Action Items
To make this discussion more concrete I intend to:
- [ ] Create an `asvcore` package on PyPI with the contents of `benchmark.py`, appropriately structured to allow extensions
- [ ] Demonstrate in a PR how this cleanup would look for ASV (this repo)
- [ ] (Hopefully) if acceptable, write the `asv-memray` external plugin for `pip` distribution

I would consider this to be a precursor to #1283, since it is not currently possible to define external benchmark types without monkey-patching `asv.benchmark.py`

## More design
I envision each `Benchmark` type, e.g. `TimeBenchmark` to be in its own module, and each module will have an `export_as_benchmark` attribute which contains the class the user will use. This means that `asv` (this repo) will populate `benchmark_types` (currently hard-coded in `benchmark.py`) on the fly, taking into account any external benchmark classes.

HaoZeke:
Please feel free to ping anyone who might have an opinion.

PR #1119, opened 2022-03-09
CI: Add Windows builds

datapythonista:
Closes #1064

PR #1287, opened 2023-05-07
ENH: Cleanup with `asvcore`

HaoZeke:
Proof of concept implementation of #1286. What's left is:
- [ ] Add a configuration option to select `mem` and conditionally add `pympler`


HaoZeke:
At the moment, if `pympler` is present; e.g. via:
```json
    "matrix": {
        "req": {
            "pympler": []
        },
    },
```
then `Mem` benchmarks will be run, otherwise they're skipped silently.

HaoZeke:
> At the moment, if `pympler` is present; e.g. via:
> 
> ```json
>     "matrix": {
>         "req": {
>             "pympler": []
>         },
>     },
> ```
> 
> then `Mem` benchmarks will be run, otherwise they're skipped silently.

This is a backwards incompatible change. Since `asv` hardcodes the `Mem` benchmarks, downstream projects expect it to continue working, including the names of the benchmark results files and the requirements matrix. I guess the most compatible way is to ensure `pympler` is part of `_base_requirements` whenever possible (i.e., not on `pypy`).

HaoZeke:
@mattip this is feature complete, and ready for review, I'm rather baffled by the Appveyor failure, hoping I can deprecate appveyor altogether with #1119.

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dynamic = [
     "version",
 ]
 dependencies = [
+    "asvcore",
```
mattip, line 28: How is the asvcore package on PyPI created? Is it from this repo?

mattip:
Where is the pyproject.toml for asvcore? It seems the [PyPI entry](https://pypi.org/project/asvcore/) is missing metadata like the repo, where to file issues, ...

pyproject.toml
HaoZeke: Currently it's over here:

https://github.com/HaoZeke/asvcore

If the PR and implementation looks good the next step would be to migrate it to the airspeed-velocity organization.

mattip: Is there a design document somewhere with how these pieces fit together: what belongs in asvcore and what remains in asv? I am not sure I understand the division.

HaoZeke:
> Where is the pyproject.toml for asvcore? It seems the [PyPI entry](https://pypi.org/project/asvcore/) is missing metadata like the repo, where to file issues, ...

Yup I thought it would be better to fill those out once it's migrated to the airspeed-velocity organization instead of where it is now (a repo under my github handle).

test/test_benchmarks.py
```diff
@@ -43,7 +40,10 @@ def test_discover_benchmarks(benchmarks_fixture):
     b = benchmarks.Benchmarks.discover(conf, repo, envs, [commit_hash],
                                        regex='example')
     conf.branches = old_branches
-    assert len(b) == 36
+    if util.ON_PYPY:
+        assert len(b) == 34
+    else:
+        assert len(b) == 36
```
HaoZeke, line 46: @mattip and I discussed that perhaps later it might make sense to have a mechanism for discovering all benchmarks, and then skipping or mark as fail or something where it isn't feasible to run.

Currently this design is because the regex is part of the classes and in the new approach the benchmark types are dynamically populated.

HaoZeke:
The `appveyor` failure might stem from an incorrect (hardcoded) file path separator as suggested by @mattip

HaoZeke:
Incredibly the `appveyor` failure seems to stem from [windows blacklisting](https://www.howtogeek.com/fyi/windows-10-still-wont-let-you-use-these-file-names-reserved-in-1974/) `aux.*` as a filename.. will update `asv_runner`.

HaoZeke:
@mattip can we go ahead and merge this since the test failures have been fixed? The rest of the issues (documentation, repo location) can be worked on after that.

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dynamic = [
     "version",
 ]
 dependencies = [
+    "asv-runner>=v0.0.4",
```
mattip, line 28: I thought we were going to change the name to `asv_runner` so that the pip install name is the same as the import.

HaoZeke: The name of the package is `asv_runner`, but `pip` (or PyPI) converts `_` to `-`. As discussed here:

https://github.com/ros/rosdistro/issues/18116#issuecomment-394897086

Also here: https://stackoverflow.com/a/52828384/1895378

mattip: I know that this Just Works(tm). But now you have to do `pip install avs-runner` and `import avs_runner`. I thnk the UX would be better if they were unified, no?

HaoZeke: > I know that this Just Works(tm). But now you have to do `pip install avs-runner` and `import avs_runner`. I thnk the UX would be better if they were unified, no?

Yup, but I can't think of a good naming paradigm to do so, since `asv-runner` is not PEP-8 compliant and with the underscores we get to this situation.

mattip: I suggest you change the name on PyPI to be PEP-8 compliant: `asv_runner` or `asvrunner` or `uls` or something else, but not `asv-runner`

HaoZeke: > I suggest you change the name on PyPI to be PEP-8 compliant: `asv_runner` or `asvrunner` or `uls` or something else, but not `asv-runner`

The package name (as per the `pyproject.toml` is `asv_runner`). PyPI swaps `_` for `-`. I was under the impression that most (all?) the plugin based projects used a similar naming convention (`_` but resolves to `-` on PyPI). I could switch to `asvrunner` but it might be best to stick to (and document) one convention, e.g. "use underscores" and let PyPI / `pip` do the renaming they do.

mattip: Ahh, got it thanks.

HaoZeke:
@mattip I've added some documentation, and also started [working on API documentation](https://haozeke.github.io/asv_runner/) for `asv_runner`. 

Let me know if it looks alright / what specific questions are not addressed :)

PR #1300, opened 2023-07-06
Bugfix: `mamba` used with an environment file never starts installation due to deadlock

akihironitta:
## Description of the Bug
When using mamba with an environment file, creating an environment never starts.

Here's a brief explanation of what's happening in `mamba.py`:
```python
import asv.util as util
util.new_multiprocessing_lock("mamba_lock")
def _mamba_lock():
    return util.get_multiprocessing_lock("mamba_lock")
def run_mamba():
    with _mamba_lock():
        print("Hello, world!")
with _mamba_lock():
    run_mamba()
```

## Description of the PR
This PR removes one of the locks around `_run_mamba` to avoid grabbing the lock inside the lock.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,7 @@ API Changes
 
 Bug Fixes
 ^^^^^^^^^
+- Fixed the deadlock when mamba is used with an environment file. (#1300)
 
 Other Changes and Additions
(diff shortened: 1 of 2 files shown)


HaoZeke (approved):
LGTM, thanks @akihironitta.

PR #1302, opened 2023-07-07
DOC: Update readthedocs and changelog

HaoZeke:
The changelog needs more work. RTD should be back online with this.

### DIFF
--- a/test/test_gh_pages.py
+++ b/test/test_gh_pages.py
@@ -8,6 +8,7 @@ from asv import util
 from . import tools
 
 
+@pytest.mark.skipif(tools.HAS_PYPY, reason="Flaky on pypy")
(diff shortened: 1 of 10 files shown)


mattip:
Let's put this in to get RTD back online. We can iterate on the finer points in subsequent PRs.

test/test_gh_pages.py
```diff
@@ -8,6 +8,7 @@
 from . import tools
 
 
+@pytest.mark.skipif(tools.HAS_PYPY, reason="Flaky on pypy")
```
QuLogic, line 11: Am I missing something about this test, or should this not be checking whether you are actually _testing on pypy_ rather than whether it just happens to be installed?

PR #1303, opened 2023-07-08
BUG,ENH: Update `mamba` solver, handle default `environment.yml`

### LINKED ISSUE #1301: `environment_type: mamba` doesn't pick up channels in `environment.yml`
## Description of the Bug
When I use `mamba` with `environment.yml`, asv doesn't read channels specified in the environment file.

For example, if I have this environment file, it fails to install `pytorch-cuda` (of `nvidia` channel) and other things as asv cannot find them in the channels that asv recognises:
```
name: benchmark

channels:
  - pytorch
  - nvidia
  - conda-forge
  - defaults

dependencies:
  - python=3.10
  - pip
  - pytorch
  - pytorch-cuda=11.8
```

---

I'm using 5ab2d29.

---

## Alternatives
A workaround is to specify channels via `asv.conf.json`:

```json
    "conda_channels": ["pytorch", "nvidia", "conda-forge", "defaults"],
```

HaoZeke:
Closes #1301. Also uses the `mamba.api` more and simplifies the plugin. Named environment handling has also been refactored (and fixed up).

### DIFF
--- a/asv/plugins/conda.py
+++ b/asv/plugins/conda.py
@@ -3,6 +3,7 @@ import re
 import os
 import tempfile
 import contextlib
+from pathlib import Path
 
 from .. import environment, util
 from ..console import log
@@ -92,8 +93,9 @@ class Conda(environment.Environment):
             log.debug("Skipping environment file due to conda_environment_file set to IGNORE")
             self._conda_environment_file = None
         elif not conf.conda_environment_file:
-            log.debug("Trying to use default environment.yml")
-            self._conda_environment_file = "environment.yml"
+            if Path("environment.yml").exists():
+                log.debug("Using environment.yml")
+                self._mamba_environment_file = "environment.yml"
 
         super(Conda, self).__init__(conf,
                                     python,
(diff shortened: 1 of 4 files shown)


akihironitta (approved):
Thank you! :)

asv/plugins/conda.py
```diff
@@ -87,6 +88,15 @@ def __init__(self, conf, python, requirements, tagged_env_vars):
...
         self._conda_channels = conf.conda_channels
         self._conda_environment_file = conf.conda_environment_file
+
+        if conf.conda_environment_file == "IGNORE":
+            log.debug("Skipping environment file due to conda_environment_file set to IGNORE")
+            self._conda_environment_file = None
+        elif not conf.conda_environment_file:
+            if Path("environment.yml").exists():
+                log.debug("Using environment.yml")
+                self._mamba_environment_file = "environment.yml"
```
h-vetinari, line 98: Should this be `self._conda_environment_file`? It's the only use of `_mamba` in this module, so it looks like a C&P mistake.

HaoZeke: Yup, good catch.

Issue #13, opened 2013-11-08
Add timeout


mdboom:
If a benchmark enters an infinite loop, things will hang.  We should have a time limit (project customizable) after which we forcibly terminate the subprocess and move along.

PR #1308, opened 2023-07-24
ENH: Add a configurable default benchmark timeout

### LINKED ISSUE #973: How to change the default benchmark timeout?
In some situations, the value of 60 seconds is not enough and I have to manually set `timeout` variable as a class attribute for each benchmark.
Is there a way to change this value for all benchmarks at once?

It would be nice to be able to customize this value through the config file or ASV CLI.

HaoZeke:
Closes #973, and also is closer to the original design discussed in #13. A minor change to `asv-runner` is needed for this as well.

Usage:
```bash
cat asv.conf.json
...
"default_benchmark_timeout": 0.02354213,
...
# Or, override via the commandline
asv run -a timeout=0.62
```

The implementation by design does not affect benchmarks which have the `timeout` attribute set.


anmyachev:
Thanks @HaoZeke!

Issue #1323, opened 2023-08-21
asv 0.6.0 errors on environments with missing `packaging` module


moorepants:
SymPy's CI that runs asv is breaking (see https://github.com/sympy/sympy/issues/25561) with this error:

```
Run asv run --config asv.conf.actions.json
Couldn't load asv.plugins.mamba because
No module named 'mamba'
Couldn't load asv.plugins.virtualenv because
No module named 'packaging'
· Unknown environment type 'virtualenv'
· No environments selected
Error: Process completed with exit code 1.
```

I've looked through the change log and the diff of 0.6.0 to the last release and see that there is a new mamba plugin, but I haven't been able to see why these failures are occurring or if we are supposed to update something on our end to work with the new release. Can you provide a suggestion?

Xemorr:
I am also experiencing this error

HaoZeke:
> I am also experiencing this error

Could you post the full error?

HaoZeke:
> SymPy's CI that runs asv is breaking (see [sympy/sympy#25561](https://github.com/sympy/sympy/issues/25561)) with this error:
> 
> ```
> Run asv run --config asv.conf.actions.json
> Couldn't load asv.plugins.mamba because
> No module named 'mamba'
> Couldn't load asv.plugins.virtualenv because
> No module named 'packaging'
> · Unknown environment type 'virtualenv'
> · No environments selected
> Error: Process completed with exit code 1.
> ```
> 
> I've looked through the change log and the diff of 0.6.0 to the last release and see that there is a new mamba plugin, but I haven't been able to see why these failures are occurring or if we are supposed to update something on our end to work with the new release. Can you provide a suggestion?

One of the possible environment management systems must be available. Although `mamba` may not be a worthwhile dependency in some situations, the `virtualenv` issue is that `packaging` is not present, and so it should be fixed by [adding it](https://pypi.org/project/packaging/) to the build environment.

PR #1240, opened 2023-02-18
Don't use distutils

jarrodmillman:
`disutiils` is going to be removed in Python 3.12 and these are the standard replacements (I believe).

PR #1324, opened 2023-08-21
MAINT: Add an install group for `virtualenv`

HaoZeke:
Needed since `distutils` deprecation in https://github.com/airspeed-velocity/asv/pull/1240.

Closes https://github.com/airspeed-velocity/asv/issues/1323.

Also https://github.com/sympy/sympy/pull/25562 and https://github.com/sympy/sympy/issues/25561.

**NOTE:** `asv` doesn't typically ensure any dependencies for its environment management plugins are present, and I am personally more of the opinion that this is not a bug and shouldn't be added to `asv`. However, `virtualenv` is pretty much the default for `asv` and so maybe an argument could be made for ensuring dependencies for it. In that case, it would be best to just depend on `virtualenv` as well.

**NOTE_II**: If `asv` shouldn't maintain dependencies for plugins (my preference) then this PR should be reworked to **just** have the changelog update.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,7 @@ Bug Fixes
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 - ``asv`` timestamps via ``datetime`` are now Python 3.12 compatible (#1331)
+- ``asv`` now provides ``asv[virtualenv]`` as an installable target
 
 0.6.0 (2023-08-20)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -46,6 +46,7 @@ Bug Fixes
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 - ``mamba`` and ``conda`` use ``environment.yml`` if it exists
+- ``virtualenv`` now requires ``packaging`` due to ``distutils`` deprecations (#1240)
 - Wheels are now built for CPython ``3.8, 3.9, 3.10, 3.11``
 
 0.5.1 (2021-02-06)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
     "asv-runner>=v0.0.9",
     "json5",
     "tabulate",
+    "packaging",
     "colorama; os_name == 'nt'",
     "pyyaml; platform_python_implementation != \"PyPy\"",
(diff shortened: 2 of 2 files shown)


mattip:
Hmm. If asv cannot be used without a package, that makes it a dependency. Maybe the error message should be extended:
> Could not find a package manager. Either virtualenv, conda, or mamba is required to run asv

On the other hand I think 90% of the users will be using virtualenv, and it is a "cheap" dependency, so this PR seems reasonable. Are there other indications virtualenv/conda/mamba should be used to build the benchmarking environments?

mattip:
What are all these "HaoZeke temporarily deployed to pypi — with GitHub Actions" messages? Should that job step only occur when merging?

mattip:
Doesn't the plugin system allow for specifying plugin-specific dependencies? like `asv[virtualenv]` or so?

HaoZeke:
> What are all these "HaoZeke temporarily deployed to pypi — with GitHub Actions" messages? Should that job step only occur when merging?

Yup, that's a bit of a mess, I think I fixed it in the repo settings (no way to fix it via the `yml` files).

> Doesn't the plugin system allow for specifying plugin-specific dependencies? like asv[virtualenv] or so?

Not yet, but that's a good feature, as noted by @brocksam as well.

> Hmm. If asv cannot be used without a package, that makes it a dependency. Maybe the error message should be extended:

> > Could not find a package manager. Either virtualenv, conda, or mamba is required to run asv

> On the other hand I think 90% of the users will be using virtualenv, and it is a "cheap" dependency, so this PR seems reasonable. Are there other indications virtualenv/conda/mamba should be used to build the benchmarking environments?


I think, given the discussions, I'll rework this to remove the dep from required (since it isn't), moving it to a new group `asv[virtualenv]` and making the error clearer.

mattip:
What happens if I run asv without `asv[virtualenv]` and without `asv[mamba]`? Do I get an appropriate error message that one of them is required?

HaoZeke:
> What happens if I run asv without `asv[virtualenv]` and without `asv[mamba]`? Do I get an appropriate error message that one of them is required?

No because `libmambapy` isn't on PyPI, and neither is `conda`. This will have to remain a runtime error.

mattip:
It is a bit weird to have `pip install asv` be broken since it cannot create virtual environments to run benchmarks. Would it be so bad to have virtualenv always be installed, and have conda as an optional addition?

PR #1329, opened 2023-08-23
MAINT: Handle conda environments in mamba plugin

### LINKED ISSUE #1325: mamba plugin in asv 0.6.0 doesn't consider dependencies in `conda_environment_file`
Shouldn't there be another branch?

https://github.com/airspeed-velocity/asv/blob/ab8804deb09a7bec692ce0ba9f1ce292994b014f/asv/plugins/mamba.py#L65-L69

Something like that:

```python
else:
   self._mamba_environment_file = conf.conda_environment_file
```

HaoZeke:
Closes #1325.

### DIFF
--- a/asv/plugins/mamba.py
+++ b/asv/plugins/mamba.py
@@ -32,8 +32,7 @@ class Mamba(environment.Environment):
     Manage an environment using mamba.
 
     Dependencies are installed using ``mamba``.  The benchmarked
-    project is installed using ``pip`` (since ``mamba`` doesn't have a
-    method to install from an arbitrary ``setup.py``).
+    project is installed using ``pip``.
     """
 
     tool_name = "mamba"
--- a/asv/plugins/mamba.py
+++ b/asv/plugins/mamba.py
@@ -66,6 +66,12 @@ class Mamba(environment.Environment):
             if (Path("environment.yml")).exists():
                 log.debug("Using environment.yml")
                 self._mamba_environment_file = "environment.yml"
+        else:
+            if (Path(conf.conda_environment_file)).exists():
+                log.debug(f"Using {conf.conda_environment_file}")
+                self._mamba_environment_file = conf.conda_environment_file
+            else:
+                log.debug(f"Environment file {conf.conda_environment_file} not found, ignoring")
 
         super(Mamba, self).__init__(conf, python, requirements, tagged_env_vars)
         self.context = libmambapy.Context()
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,7 @@ Bug Fixes
 - ``pip`` dependencies in ``environment.yml`` files for the ``mamba`` plugin are handled correctly (#1326)
 - ``asv.config.json`` matrix requirements no longer need ``pip+`` set explicitly
   for calling the ``pip`` solver for ``virtualenv``
+- ``asv`` will now use ``conda_environment_file`` if it exists (#1325)
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/asv/plugins/conda.py
+++ b/asv/plugins/conda.py
@@ -65,8 +65,7 @@ class Conda(environment.Environment):
     Manage an environment using conda.
 
     Dependencies are installed using ``conda``.  The benchmarked
-    project is installed using ``pip`` (since ``conda`` doesn't have a
-    method to install from an arbitrary ``setup.py``).
+    project is installed using ``pip``.
     """
     tool_name = "conda"
     _matches_cache = {}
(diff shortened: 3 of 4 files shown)


asv/plugins/mamba.py
```diff
@@ -66,6 +66,12 @@ def __init__(self, conf, python, requirements, tagged_env_vars):
             if (Path("environment.yml")).exists():
                 log.debug("Using environment.yml")
                 self._mamba_environment_file = "environment.yml"
+        else:
+            if (Path(conf.conda_environment_file)).exists():
+                log.debug(f"Using {conf.conda_environment_file}")
+                self._mamba_environment_file = conf.conda_environment_file
+            else:
+                log.debug(f"Environment file {conf.conda_environment_file} not found, ignoring")
```
mattip, line 74: Shouldn't line 74 be an error? How can one specify a mamba setup without a config file?

Is the comment on line 34 correct? `asv` should be using `pip install .` and not `setup.py`, which may not exist. 

Also: I think "magical" code like the stanza above (line 65) is problematic: the default value of `conf.conda_environment_file should be `"environment.yml"` which would both self-document the code and reduce a level of logic here, the lines 65-68 can be removed, and the new code shifted out one level.

HaoZeke: > Shouldn't line 74 be an error? How can one specify a mamba setup without a config file?

It isn't an error because `mamba` will be used to setup the matrix of dependencies in `asv.conf.json`


> Is the comment on line 34 correct? `asv` should be using `pip install .` and not `setup.py`, which may not exist.

Yup, the docstring is out of date, updating

> Also: I think "magical" code like the stanza above (line 65) is problematic: the default value of `conf.conda_environment_file should be `"environment.yml"` which would both self-document the code and reduce a level of logic here, the lines 65-68 can be removed, and the new code shifted out one level.

The problem is that users might want their benchmarks to run based on just the matrix of depedencies, instead of the environment file. Especially since the interaction of versions gets very complicated (e.g. matrix dependencies want to test `numpy` at two versions but `environment.yml` pins `numpy`). For those cases, `IGNORE` is the cleanest way forward.

mattip: Not sure I understand. According to [the documentation](https://asv.readthedocs.io/en/latest/asv.conf.json.html#conda-environment-file) (which uses `NONE`, not `IGNORE`), 
>  If there are packages present in matrix, an additional `conda env update` call is used to install them after the environment is created. 

Problem solved, no? The matrix spec will override the `environment.yml` one, or they can specify `NONE`/`IGNORE`

HaoZeke: > Not sure I understand. According to [the documentation](https://asv.readthedocs.io/en/latest/asv.conf.json.html#conda-environment-file) (which uses `NONE`, not `IGNORE`),

The feature was added in the latest release, should have been documented as `IGNORE`, this PR includes the doc fix.

> > If there are packages present in matrix, an additional `conda env update` call is used to install them after the environment is created.
> 
> Problem solved, no? The matrix spec will override the `environment.yml` one, or they can specify `NONE`/`IGNORE`

Yup, that's the idea, but also why the function can't be simplified further IMO.

mattip: - Change default to `environment.yml`
- Refactor code to be 
```
        if conf.conda_environment_file == "IGNORE":
            log.debug("Skipping environment file due to conf.conda_environment_file set to IGNORE")
            self._mamba_environment_file = None
        elif (Path(conf.conda_environment_file)).exists():
            log.debug(f"Using {conf.conda_environment_file=}")
            self._mamba_environment_file = conf.conda_environment_file
        else:
            raise ValueError(f"Environment file {conf.conda_environment_file=} not found")
```

HaoZeke: This does sort of break user expectations, they'd need to set it to IGNORE to get `asv` to build the environment matrix if `environment.yml` isn't present.

I think as a default `IGNORE` makes more sense then?

mattip: They chose to use a mamba/conda build, and doing `asv --help` will show what the default would be. Do you know of users who are using this option? Maybe we could ask them. I don't think ignore-by-default makes sense, but :shrug:

HaoZeke: I guess I was thinking more along the lines of, given an `asv` project (which is defined only by `asv.config.json` users should be able to use `asv` with any of the environment plugins as long as they have dependencies, so swapping `virtualenv` for `mamba` or `conda` should not (and currently does not) require having an `environment.yml` file.

mattip: OK. Maybe add that to the documentation somehow.

mattip: It is also easy to change the default if `IGNORE` proves unpopular.

mattip:
Thanks @HaoZeke

PR #1335, opened 2023-09-03
MAINT: Don't warn if valid environment is found

### LINKED ISSUE #1334: spurious warning about mamba plugin
If I have `"environment_type": "virtualenv"` in my `asv.conf.json`, I should not be seeing this:
```
Couldn't load asv.plugins.mamba because
No module named 'mamba'
```
As a design rule, no warnings should be emitted for idiomatic usage of a tool like `asv`. Please delete this warning, because it's potentially confusing when looking at (for example) CI logs.

HaoZeke:
Closes #1334.

### DIFF
--- a/asv/plugin_manager.py
+++ b/asv/plugin_manager.py
@@ -30,7 +30,7 @@ class PluginManager:
                 self.init_plugin(mod)
                 self._plugins.append(mod)
             except ModuleNotFoundError as err:
-                log.error(f"Couldn't load {name} because\n{err}")
+                log.debug(f"Couldn't load {name} because\n{err}")
                 continue  # Couldn't find mamba
 
     def _load_plugin_by_name(self, name):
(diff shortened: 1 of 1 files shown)


asv/plugin_manager.py
```diff
@@ -30,7 +30,7 @@ def load_plugins(self, package):
                 self.init_plugin(mod)
                 self._plugins.append(mod)
             except ModuleNotFoundError as err:
-                log.error(f"Couldn't load {name} because\n{err}")
+                log.debug(f"Couldn't load {name} because\n{err}")
                 continue  # Couldn't find mamba
```
mattip, line 34: This code path can only be hit if package == mamba? Maybe the error is fine and the problem is in the code that is calling this function?

HaoZeke: I reworked this to make it clearer. It starts to crop up with the `mamba` plugin, since that needs `libmambapy` and imports it. The error will also be triggered for `virtualenv` (but not `conda` since that will raise an `OSError` if it isn't found).

mattip:
Thanks @HaoZeke

PR #1338, opened 2023-09-09
CI: Enable wheel builds again

### LINKED ISSUE #1337: Wheels are missing on PyPI
I noticed this because of downloading dependencies and building a wheel for `asv` 0.6 in SciPy CI. It is good practice to release a wheel also for pure Python packages. No releases have one though. Is there any reason why they are missing?

HaoZeke:
Should close #1337. Well, once we have [TPM setup](https://docs.pypi.org/trusted-publishers/using-a-publisher/).


HaoZeke:
Close / reopen to reduce visual noise.

mattip:
Can we manually upload the wheels for now, instead of automatically publishing? We do not release that often, and getting the automatic publishing to work is a bit complicated.

PR #1339, opened 2023-09-09
CI: Enable wheel builds again

### LINKED ISSUE #1337: Wheels are missing on PyPI

HaoZeke:
Should close #1337. Well, once we have [TPM setup](https://docs.pypi.org/trusted-publishers/using-a-publisher/).

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -1,5 +1,8 @@
 # Build on every branch push, tag push, and pull request change:
 # From: https://github.com/pypa/cibuildwheel/blob/main/examples/github-deploy.yml
+# Also:
+# https://github.com/airspeed-velocity/asv/blob/main/.github/workflows/wheels.yml
+# include [wheel build] in the commit to trigger wheel builds
 name: Build wheels
 on: [push, pull_request, workflow_dispatch]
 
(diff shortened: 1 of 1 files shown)


mattip:
Please add artifact upload to this so we have a way to see the wheels before they are published

HaoZeke:
> Please add artifact upload to this so we have a way to see the wheels before they are published
Yup, already part of it, can be viewed on the actions tab:
https://github.com/airspeed-velocity/asv/actions/runs/6134080274

I will try to add a comment action with the link.

HaoZeke:
PyPy test failure is unrelated.

.github/workflows/build_wheels.yml
```diff
@@ -28,7 +55,7 @@ jobs:
           - [macos-12, macosx_x86_64]
           - [macos-12, macosx_arm64]
           - [windows-2019, win_amd64]
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11', 'pypy3.8', 'pypy3.9']
```
mattip, line 58: If PyPy3.8 causes problems, we should drop it.

mattip:
Thanks @HaoZeke

PR #1349, opened 2023-10-10
BUG: build/include only asv in bdist

### LINKED ISSUE #1348: test/ and benchmarks/ included as separate packages in the bdist
Building and installing a bdist wheel results in three separate packages: `asv`, `benchmarks` and `test`. Especially in a system distribution environment, the latter two package names carry a high clash risk.

vishwin:
setuptools [automatically detects the flat-layout](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#flat-layout) and excludes items accordingly. Fixes #1348

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,8 +89,6 @@ license-files = [
     "LICENSE.rst",
 ]
 
-[tool.setuptools.packages.find]
-namespaces = false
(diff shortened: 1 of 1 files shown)


HaoZeke:
Thanks @vishwin 🥳

Issue #1028, opened 2022-02-13
Possible regression with asv 0.5.1 using parameterized benchmark when some but not all skipped


ianthomas23:
Firstly, thanks for the fantastic library.

I am getting strange results when running benchmarks for which some parameter combinations are skipped and some not. This was OK for asv 0.4.2, but not for 0.5.1. I have a simple reproducer.

Test code:
``` python
class Simple:
    params = ([False, True])
    param_names = ["ok"]

    def time_failure(self, ok):
        if ok:
            x = 34.2**4.2
        else:
            raise NotImplementedError
```

Using a venv in Ubuntu MATE 20.10:
```
>>> pip list
Package       Version
------------- -------
asv           0.5.1  
distlib       0.3.4  
filelock      3.4.2  
pip           20.0.2 
pkg-resources 0.0.0  
platformdirs  2.5.0  
setuptools    44.0.0 
six           1.16.0 
virtualenv    20.13.1
>>> asv run -b Simple
Creating environments
<snip>
[100.00%] ··· step_detect.Simple.time_failure                         1/2 failed
[100.00%] ··· ======= =========
                 ok             
              ------- ---------
               False    failed 
                True   197±2ns 
              ======= =========
>>> asv show 2e59
Commit: 2e59 <master>

step_detect.Simple.time_failure [onion/virtualenv-py3.8-six]
Traceback (most recent call last):
  File "/home/iant/.venv/asv/bin/asv", line 8, in <module>
    sys.exit(main())
  File "/home/iant/.venv/asv/lib/python3.8/site-packages/asv/main.py", line 38, in main
    result = args.func(args)
  File "/home/iant/.venv/asv/lib/python3.8/site-packages/asv/commands/__init__.py", line 49, in run_from_args
    return cls.run_from_conf_args(conf, args)
  File "/home/iant/.venv/asv/lib/python3.8/site-packages/asv/commands/show.py", line 49, in run_from_conf_args
    return cls.run(
  File "/home/iant/.venv/asv/lib/python3.8/site-packages/asv/commands/show.py", line 93, in run
    cls._print_results(conf, commit, result_iter,
  File "/home/iant/.venv/asv/lib/python3.8/site-packages/asv/commands/show.py", line 169, in _print_results
    cls._print_benchmark(machine, result, benchmarks[name],
  File "/home/iant/.venv/asv/lib/python3.8/site-packages/asv/commands/show.py", line 179, in _print_benchmark
[...]

PR #1351, opened 2023-11-16
Handle partially-skipped benchmarks

ianthomas23:
This is a fix for #1028 so that partially-skipped benchmarks can still have their results displayed. The fix follows comment https://github.com/airspeed-velocity/asv/issues/1028#issuecomment-1793496681. It does not include a test though.


HaoZeke:
Thanks for this @ianthomas23 🥳

PR #1357, opened 2024-01-05
MAINT: fix instances of get in mamba plugin

hmaarrfk:
It expects a list, but with `.get` it can return None


HaoZeke:
Thanks for this @hmaarrfk, could you rename your commit to have `MAINT:` as a prefix?

hmaarrfk:
done.

HaoZeke:
Thanks @hmaarrfk 🥳

Issue #1353, opened 2023-12-08
BUG: "environment_type": "mamba" fails


jmccreight:
On both MacOs and Windows, I'm unable to use mamba (which is problematic since conda is timing out when run in asv on windows). On both I get the same error 

```
* Unknown environment type 'mamba'
* No environments selected
```

pip show indicates that I'm up-to-date with asv 0.6.1.

Maybe I'm missing something in my setup, but I've looked through the documentation as much as I can and dont see anything I find as germane. 

Thanks for your help.

HaoZeke:
Is `libmamba` present? Could you provide some more information about how the rest of the environment is populated?

jmccreight:
Hi,
Are you asking if libmamba is in my PATH? it is not. but "mamba" is as a
function.

I have been using mambaforge (which I see is now identical to miniforge as
of 9/2023 and specifically discouraged). If you have a suggestion for
installing mamba so that asv can use it, I'm open to it as I guess I'm
changing how I do things in the near future anyway.

thanks and happy new year!


-------------------------------------------------------
James L. McCreight
Research Hydrologist ~ Software Engineer
USGS Enterprise Capacity
UCAR CPAESS
cell:         831-261-5149

jmccreight:
Hi again, 

I think the simple solution is just `conda install libmambapy` in the environment with asv. Seems obvious. But I had to understand more on the relationship of conda and mamba to before I knew what to do. 

I think that saying (in the installation instructions) that:
* [libmambapy](https://mamba.readthedocs.io/en/latest/python_api.html), which is typically part of mamba

is kind of misleading because, at least as far as I could tell, while libmamba and libmambapy are available in the base env, you can not (easily) access them for other environments. I also did not want to install asv into base (which is forbidden by mamba). I was getting hung up on expecting libmamba to be available like conda is, from any environment. But maybe I missed another option here? As far as my solution is concerned, its irrelevant that these ship with mamba because you need a separate installation anyway.

Of course, libmambapy is not on pip and so cant be included in the pyproject.toml as an optional dependency either. 

I'm still curious, how do you suggest making libmambapy available in non-base environments when using mamba? And I think it could be very helpful to mention this in the docs in one or two sentences.

Thanks!

maximlt:
Have you tried running conda with the [conda-libmamba-solver](https://conda.github.io/conda-libmamba-solver/user-guide/), instead of mamba? This is now the new conda default solver.

[...]

PR #1370, opened 2024-01-17
Fix mamba environment option

dstansby:
I'm attempting to fix https://github.com/airspeed-velocity/asv/issues/1353. Currently this involves:
- adding some extra docs
- adding an import guard to check that `libmambapy` is installed
- making the `mamba` plugin discoverable by `asv` (by importing it in `plugins.__init__.py`)

However, I'm not convinced the mamba option has ever worked... The [mamba Python API](https://mamba.readthedocs.io/en/latest/python_api.html) does not contain `mamba.api.MambaSolver()`, which was added as an import in https://github.com/airspeed-velocity/asv/pull/1238. @HaoZeke do you know any more about this import and what package needs to be installed to access it?

### DIFF
--- a/asv/plugins/mamba.py
+++ b/asv/plugins/mamba.py
@@ -11,7 +11,13 @@ try:
 except ImportError:
     from yaml import Loader
 
-from mamba.api import libmambapy, MambaSolver
+try:
+    from mamba.api import MambaSolver
+    import libmambapy
+    _HAS_LIBMAMBAPY = True
+except ImportError:
+    _HAS_LIBMAMBAPY = False
+
 
 from .. import environment, util
 from ..console import log
@@ -51,6 +57,8 @@ class Mamba(environment.Environment):
             Dictionary mapping a PyPI package name to a version
             identifier string.
         """
+        if not _HAS_LIBMAMBAPY:
+            raise ImportError("Failed to import 'libmambapy' Python module.")
         self._python = python
         self._requirements = requirements
         self._mamba_channels = conf.conda_channels


HaoZeke:
> @HaoZeke do you know any more about this import and what package needs to be installed to access it?

I will take a closer look after next week (am traveling). It was certainly available at the time of that PR, though `libmambapy` has changed quite a bit over the past year or so.

HaoZeke:
So there used to be a `mamba.api` which shipped a default implementation of `MambaSolver` as noted in an [older fork](https://github.com/jonashaag/mamba/blob/a0f81393dd8eff38fde04f59cb5e6830f2995193/docs/source/python_api.rst#L85). However it has been removed so I'll setup a different default implementation for us (`boa` moves far too quickly and is too heavy to be a dependency).

asv/plugins/__init__.py
```diff
@@ -1 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+# Import to make sure plugins are discoverable
+import asv.plugins.conda
+import asv.plugins.mamba
```
HaoZeke, line 5: This shouldn't be required? (plugins are auto-discovered, failures mean they are dropped from valid options)

asv/plugins/mamba.py
HaoZeke, line 1: Rather than explicitly raise from here the current approach is to not let `mamba` be a valid option for an environment if the import fails.

The reason behind this is to allow for external plugins (it is also why we don't want to have an explicit list of allowed values).

docs/source/using.rst
```diff
@@ -80,7 +80,8 @@ suite are:
 - ``environment_type``: The tool used to create environments.  May be
   ``conda`` or ``virtualenv`` or ``mamba``.  If Conda supports the
   dependencies you need, that is the recommended method. Mamba is faster
-  but needs a newer Python version (3.8 or greater).
+  but needs a newer Python version (3.8 or greater) and the ``libmambapy``
+  package must be installed to provide access to the mamba Python API.
```
HaoZeke, line 84:
```suggestion
  but needs a newer Python version (3.8 or greater) and the ``libmambapy``
  package must be installed (typically through ``conda``) to provide access to the mamba Python API.
```

Clarification since `libmambapy` is not present on PyPI.

dstansby:
Fab, thanks for picking this up!

PR #1371, opened 2024-01-25
Remove `asv dev` from docs

CodyCBakerPhD:
#1200 and subsequent releases removed `asv dev` from the CLI, but the docs still had remnant mentions of it

### DIFF
--- a/docs/source/writing_benchmarks.rst
+++ b/docs/source/writing_benchmarks.rst
@@ -64,6 +64,9 @@ these features together with::
 
        asv run --python=same --quick --show-stderr --dry-run
 
+.. versionchanged:: 0.6.0
+   This replaces the now removed ``asv dev`` command.
+
 You may also want to only do a basic check whether the benchmark suite
 is well-formatted, without actually running any benchmarks::
 
--- a/docs/source/writing_benchmarks.rst
+++ b/docs/source/writing_benchmarks.rst
@@ -59,8 +59,8 @@ that your benchmarks may produce, pass the ``--show-stderr`` flag::
 
        asv run --show-stderr
 
-Finally, there is a special command, ``asv dev``, that uses all of
-these features and is equivalent to::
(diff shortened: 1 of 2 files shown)


HaoZeke (changes requested):
Thanks for this @CodyCBakerPhD. Could you rename the commits so they have `DOC: ` prepended to them?

CodyCBakerPhD:
@HaoZeke Thanks, fixed my commit messages

docs/source/writing_benchmarks.rst
```diff
@@ -59,8 +59,8 @@ that your benchmarks may produce, pass the ``--show-stderr`` flag::
 
        asv run --show-stderr
 
-Finally, there is a special command, ``asv dev``, that uses all of
-these features and is equivalent to::
+Finally, a quick way to test out the benchmark suite before doing a full run is to use all of
+these features together with::
```
HaoZeke, line 63: Maybe a [versionchanged directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionchanged) here to mention that there used to be `asv dev` for the same?

HaoZeke (changes requested):

docs/source/writing_benchmarks.rst
CodyCBakerPhD: That's up to you.

Personally I prefer docs to refer to the latest version of things; since the argument was removed, there must have been a reason why, so why bother educating users about a feature that is now unsupported?

Likewise, if they are intentionally downpinning to a past version, the past version of the docs at that point in time would still be valid in that respect

HaoZeke: For upgrade ease typically, if users come by trying to migrate from a workflow with `asv dev` it would make sense to highlight the alternative. I'll just push a minor update and get this through :)

HaoZeke:
Thanks @CodyCBakerPhD 🚀

PR #1372, opened 2024-01-27
MAINT: Vendor `mamba.api`

HaoZeke:
Closes #1353. Also addresses comments in #1370. There's an upstream clarification issue in https://github.com/mamba-org/mamba/issues/3159. However, I think this makes sense to go in as-is. Will cut a release soon with the fix.

PR #1373, opened 2024-01-28
BUG: Add `MAMBARC` support to `mamba`

### LINKED ISSUE #1358: ENH: Override default channels for `conda` / `mamba`
In the mamba plugin, the conda-forge channel is added by default

https://github.com/airspeed-velocity/asv/blob/b8cbf81ccb808158102692a6fc42b8a9b2f2861e/asv/plugins/mamba.py#L59

I understand that this is convenient, but it seems to me that this "removes" choice from the end user's control.

I don't see why users should be required to have this channel enabled.

HaoZeke:
~**Cannot be reviewed until #1372 is in**~

Closes #1358. Also explicitly removes `conda-forge` as a default for `mamba`. Users are free to either:
- Specify it in their `.condarc` and use `export MAMBARC=$HOME/.condarc`
or (better)
- Explicitly provide an `environment.yml` or set the corresponding variable in `asv.conf.json`

The documentation has also been updated to reflect the same.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
         r-version: ['release']
     timeout-minutes: 30
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
@@ -65,7 +65,7 @@ jobs:
       fail-fast: false
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
@@ -96,7 +96,7 @@ jobs:
   docs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - uses: actions/setup-python@v4
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Linux CI
+name: Linux and MacOS CI
 
 on: [push, pull_request]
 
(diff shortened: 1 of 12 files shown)


mattip:
Does this need a rebase now that #1372 is in? Does this need a specific test?

HaoZeke:
> Does this need a rebase now that #1372 is in? Does this need a specific test?

@mattip, I've rebased and added more tests covering all the additional features for the `mamba` solver, should be good to go.

```bash
test/test_environment_bench.py::test_asv_mamba[with_conda_forge] PASSED  [ 50%]
test/test_environment_bench.py::test_asv_mamba[empty_conda_channels] PASSED [ 62%]
test/test_environment_bench.py::test_asv_mamba_condarc[with_proper_condarc_and_mambarc] PASSED [ 75%]
test/test_environment_bench.py::test_asv_mamba_condarc[with_condarc_but_no_mambarc] PASSED [ 87%]
test/test_environment_bench.py::test_asv_mamba_condarc[without_condarc_and_mambarc] PASSED [100%]
```

mattip:
CI is not happy. It would be nice to have at least one windows (and maybe one macos) CI run to prove the code works on those platforms as well, since path parsing differs.

.github/workflows/ci.yml
```diff
@@ -60,7 +60,7 @@ jobs:
 
   test_env:
     name: test_environments
-    runs-on: "ubuntu-latest"
+    runs-on: [ ubuntu-latest, macos-latest ]
```
mattip, line 63: I think you need a `matrix:` section somewhere to make this work

.github/workflows/ci_win.yml
```diff
@@ -0,0 +1,72 @@
...
+    name: test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["windows-latest"]
+        python-version: ["3.7"]
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v3
```
mattip, line 20: checkout v4 should be used to avoid a warning about deprecated node 16

HaoZeke:
@mattip I think this now ready.

mattip:
thanks @HaoZeke

Issue #1368, opened 2024-01-17
MAINT: Update default branch from master to main


dstansby:
Since 2020 GitHub has been using `main` for the default branch name. See https://github.com/github/renaming/ for more info. It would be good to update:

- The default branch name in this repository
- Any instances of `master` in the sample configuration and docs

HaoZeke:
Thanks @dstansby. I'm ok with this, and have the permissions to do the needful, but I'll leave it till Monday, if there is no dissent then I'll go ahead :)

HaoZeke:
Pinging some other maintainers with the commit bit as well, maybe @mattip, @fangchenli, @pv, @mdboom..

HaoZeke:
Also @datapythonista :)

mattip:
Seems like we should do this.

fangchenli:
+ 1

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

HaoZeke:
Done, thanks all :)

PR #1375, opened 2024-02-01
MAINT: Finalize rename to `main`

HaoZeke:
Part of #1368.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -3,7 +3,7 @@ name: pre-commit
 on:
   pull_request:
   push:
-    branches: [master]
+    branches: [main]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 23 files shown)


mattip:
Could you add a release note? I think this could have an impact on projects that still use `master' as the default branch: they will need to now use a non-default configuration, right?

HaoZeke:
> Could you add a release note? I think this could have an impact on projects that still use `master' as the default branch: they will need to now use a non-default configuration, right?

I'm not sure it will though, if the projects have an existing `asv.config.json` they shouldn't be affected. Internally `asv` still assumes `master` even after this simply because `asv` shouldn't be responsible for making changes to `git` configurations and that still defaults to `master`. The default example config has changed, but that shouldn't be a breaking change.

I will add a release note.

mattip:
Thanks @HaoZeke

mattip:
The release note can be part of the general release

PR #1376, opened 2024-02-06
DOC : added `default_benchmark_timeout` in `using.rst` (proposal)

Schefflera-Arboricola:
having `default_benchmark_timeout` in the docs might be helpful. (origin: https://github.com/networkx/nx-parallel/pull/43)

Thank you :)


HaoZeke:
This looks great, thank you for the contribution! For future reference, please note that our commit convention requires commits to be formatted as `ID: ` which can be `DOC: ` or `MAINT:` or similar.

Excellent work @Schefflera-Arboricola 🥳

PR #1377, opened 2024-02-10
MAINT,DOC: `pip` and network connectivity defaults

### LINKED ISSUE #1356: Bad out-of-box support for PDM-based builds
I am adding ASV to [mismo](https://github.com/nickcrews/mismo/), which uses PDM as the project manager and build tool.

With no `build_command` specified in the asv.conf.json, then since the project has a pyproject.toml, the following default build command is used:

https://github.com/airspeed-velocity/asv/blob/b8cbf81ccb808158102692a6fc42b8a9b2f2861e/asv/environment.py#L918-L923

However, this fails, because the build dependency of `pdm.backend`, but the various flags there disallow going to PyPI to fetch it (as I understand it, perhaps the root cause is different). In order to get the build to work, I have to use
`python -mpip wheel -w {build_cache_dir} {build_dir}`

Why do we disallow fetching to PyPI? It seems like fetching the build backend is going to be needed in many cases, including pdm.backend, poetry-core, and hatchling? Is there something else we can do to make the out-of-box experience better for these users?

HaoZeke:
Closes #1356.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,4 +30,4 @@ By using the following markdown::
 License: `BSD three-clause license
 <http://opensource.org/licenses/BSD-3-Clause>`__.
 
-Authors: Michael Droettboom, Pauli Virtanen
+Authors: Michael Droettboom, Pauli Virtanen, asv Developers
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -920,15 +920,13 @@ class Environment:
         if cmd is None:
             if has_file('pyproject.toml'):
                 cmd = [
-                    ("PIP_NO_BUILD_ISOLATION=false "
-                     "python -mpip wheel --no-deps --no-index "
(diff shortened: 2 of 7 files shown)


mattip:
Is there a sample test project with a `pyproject.toml`?

HaoZeke:
> Is there a sample test project with a `pyproject.toml`?

Yes, but it needs a newly released `asv_runner` first, will prepare the release ASAP.

HaoZeke:
@mattip this should now be good to go.

HaoZeke:
CI failure is real, needs `importlib.metadata` on 3.7, cutting a new bugfix release for asv_runner.

HaoZeke:
@mattip finally ready ^_^

mattip:
Thanks @HaoZeke

PR #1380, opened 2024-02-11
MAINT: Switch to 3.8

### LINKED ISSUE #1318: MAINT,TST: Update minimum requirements and tests
Python 3.7 is EOL. 3.8 would require significant rewrites to the test suite.

HaoZeke:
Closes #1318 and https://github.com/airspeed-velocity/asv_runner/issues/28

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.7", "3.12", "pypy-3.8", "pypy-3.9"]
+        python-version: ["3.8", "3.12", "pypy-3.8", "pypy-3.9"]
         r-version: ['release']
     timeout-minutes: 30
     steps:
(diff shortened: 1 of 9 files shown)


mattip:
The disadvantage to this is that projects who wish to reproduce results for 3.7 can no longer do so. Are there features from higher version pythons that asv wants to utilize? If not then I think we should reject the issue. I am not sure what you meant by (from the issue)
>  3.8 would require significant rewrites to the test suite.

I don't see significant rewrites here ...

HaoZeke:
> The disadvantage to this is that projects who wish to reproduce results for 3.7 can no longer do so. Are there features from higher version pythons that asv wants to utilize? If not then I think we should reject the issue. I am not sure what you meant by (from the issue)
> 
> > 3.8 would require significant rewrites to the test suite.
> 
> I don't see significant rewrites here ...

I've been careful to not use any features beyond `3.7`, so we can reject this (though I will need to slightly modify the documentation workflow for `asv_runner`, the `myst` parser needs `3.8`).

HaoZeke:
I'm going to go ahead and close this for now.

PR #1391, opened 2024-03-19
DEP: only depend on tomli in Python 3.10 and older, use standard lib otherwise

neutrinoceros:
Just noticed this in passing.
`tomli` has been added in the standard library under `tomllib` in Python 3.11, so it's not needed as a third party dependency thereafter.


HaoZeke (approved):

HaoZeke:
Thanks @neutrinoceros

PR #1392, opened 2024-03-20
Add codespell config, pre-commit hook, and GH action (to catch new typos) + make it fix some it finds

yarikoptic:
TODO
- [x] remove TEMP commit after seeing if pre-commit workflow catches it
- [x] remove github action if pre-commit wf catches it

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,3 +18,10 @@ repos:
     hooks:
     -   id: ruff
         args: ["--fix"]
+-   repo: https://github.com/codespell-project/codespell
+    # Configuration for codespell is in pyproject.toml
+    rev: v2.2.6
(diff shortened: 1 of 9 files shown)


yarikoptic:
ok, pre-commit works but needs duplication for excludes. I might just exclude that specific name... and we do not need dedicated GH action -- will remove it.

HaoZeke (changes requested):
LGTM, could the commits be reworded with a MAINT: tag? (and also squashed, I think one MAINT: commit adding the config and one DOC: using it should suffix).


Sorry for the delayed review!

yarikoptic:
done -- one commit rebased and squashed now

yarikoptic:
Failing CI runs seems to not relate

PR #1418, opened 2024-07-13
Add TAGS as range parameter, fixes #488 fixes #1027

### LINKED ISSUE #488: asv run "ALL_TAGS"
If I create a new benchmark I generally want to run the benchmarks for all existing releases (in my case git tags). The `-b` optional already helps to select only the new test but it's either not easy to run on all tags or I missed the option.

Nevertheless: I think it would be really nice to have an `"ALL_TAGS"` option equivalent to the already existing `"ALL"`, `"EXISTING"` and `"NEW"` option for `range` in `asv run`.

### LINKED ISSUE #1027: Sampling each tag in the repo
Is it possible to run the benchmark for each of published tag of a repo? 

Currently I am using --steps to limit the number of commits it selects. It will be better if one can run against each of the tag.

timkpaine:


asv/commands/run.py
```diff
@@ -232,12 +236,15 @@ def run(cls, conf, range_spec=None, steps=None, date_period=None,
...
-        elif range_spec == 'EXISTING':
+        elif range_spec == "EXISTING":
             commit_hashes = get_existing_hashes(conf.results_dir)
         elif range_spec == "NEW":
             # New commits on each configured branches
             old_commit_hashes = get_existing_hashes(conf.results_dir)
             commit_hashes = repo.get_new_branch_commits(conf.branches, old_commit_hashes)
+        elif range_spec == "TAGS":
+            # All tags on each configured branches
+            commit_hashes = list(reversed(repo.get_tags().keys()))
```
HaoZeke, line 247:
```suggestion
            commit_hashes = reversed(list(repo.get_tags().keys()))
```

Should fix CI for 3.7, otherwise this looks good to me.

HaoZeke (changes requested):
Minor CI related nit.

mattip:
Rebased off main to get an updated CI, and merked ready for review. I think this is a nice feature.

HaoZeke (approved):
LGTM, and CI. Thanks @timkpaine

PR #1420, opened 2024-08-06
Allow `asv.conf.jsonc` file by default

lucascolley:
In SciPy, we have lots of comments in [our `asv.conf.json` file](https://github.com/scipy/scipy/blob/main/benchmarks/asv.conf.json). But comments are not allowed in JSON files (hence developers on GitHub or VSCode see lots of red highlighting).

This PR allows the config file to be named `asv.conf.jsonc`(a JSON with comments file), without having to pass the name to `--config`.

### DIFF
--- a/asv/config.py
+++ b/asv/config.py
@@ -1,6 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-import os
 import sys
 from pathlib import Path
 
@@ -22,7 +21,7 @@ def _get_config_path():
     num_matches = 0
     for e in extensions:
         p = path.with_suffix(e)
-        if p.exists:
+        if p.exists():
             num_matches += 1
             path = p
 
@@ -37,7 +36,7 @@ def _get_config_path():
         )
     else:
         # exactly one valid config file found
-        return path
+        return str(path)
 
 
 class Config:
--- a/asv/config.py
+++ b/asv/config.py
@@ -2,11 +2,42 @@
 
 import os
 import sys
+from pathlib import Path
 
 from .console import log
 from . import util
 
 # TODO: Some verification of the config values
+
+
+def _get_config_path():
+    """
+    Check the default config file path for all valid extensions.
+    Raise if no file is found, or if multiple are found for the valid extensions.
+    Return the path of the config file if exactly one is found.
+    """
+    extensions = ['.json', '.jsonc']
+    path = Path('asv.conf')
+
+    num_matches = 0
+    for e in extensions:
+        p = path.with_suffix(e)
+        if p.exists:
+            num_matches += 1
+            path = p
+
+    if num_matches == 0:
+        raise util.UserError(
+            f"No `asv.conf` file found for valid extensions: {extensions}."
+        )
+    elif num_matches > 1:
+        raise util.UserError(
+            f"Multiple `asv.conf` files found for valid extensions: {extensions}. "
+            f"Please specifiy which file to use with `--config=FILEPATH`."
+        )
+    else:
+        # exactly one valid config file found
+        return path
 
 
 class Config:
@@ -48,13 +79,13 @@ class Config:
     def load(cls, path=None):
         """
         Load a configuration from a file.  If no file is provided,
-        defaults to `asv.conf.json`.
+        defaults to `asv.conf` with valid extensions `['.json', '.jsonc']`.
         """
-        if not path:
-            path = "asv.conf.json"
-
-        if not os.path.isfile(path):
-            raise util.UserError(f"Config file {path} not found.")
+        if path:
+            if not Path(path).exists():
+                raise util.UserError(f"Config file {path} not found.")
+        else:
+            path = _get_config_path()
 
         d = util.load_json(path, cls.api_version, js_comments=True)
         try:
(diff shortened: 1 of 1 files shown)


asv/config.py
```diff
@@ -48,13 +48,16 @@ def __init__(self):
...
         """
         if not path:
             path = "asv.conf.json"
 
         if not os.path.isfile(path):
-            raise util.UserError(f"Config file {path} not found.")
+            # also check for `asv.conf.jsonc` (JSON with comments)
+            path += 'c'
+            if not os.path.isfile(path):
+                raise util.UserError(f"Config file {path} not found.")
```
lucascolley, line 60:
```suggestion
            path_c = path + 'c'
            if not os.path.isfile(path_c):
                raise util.UserError(f"Config file {path} not found.")
```

lucascolley:
@HaoZeke @mattip does this sounds like something you'd be happy to support by default? Or do you think comments in the config file is a niche use case?

HaoZeke (commented):
Is the `.jsonc` file extension widely known / used? If not I don't know that this would make sense to explicitly search for from an ASV perspective, given that this is mostly run in CI anyway and since passing the file works via --config

lucascolley:
500ish stars on Microsoft's parser (https://github.com/microsoft/node-jsonc-parser), so not widely known, but also not niche IMO.

My reason for opening this PR is that appending `c` to the filename appears to be the idiomatic solution to all the red highlighting that appears on the GitHub interface. Unless I am missing a setting.

But no worries if this seems like more hassle than it is worth - I was just hoping that I could change to `jsonc` downstream without breaking peoples' local asv invocations.

asv/config.py
```diff
@@ -48,13 +48,16 @@ def __init__(self):
...
         """
         if not path:
             path = "asv.conf.json"
 
         if not os.path.isfile(path):
-            raise util.UserError(f"Config file {path} not found.")
+            # also check for `asv.conf.jsonc` (JSON with comments)
+            path_c = path + 'c'
+            if not os.path.isfile(path_c):
+                raise util.UserError(f"Config file {path} not found.")
```
HaoZeke, line 60: I guess my main concern with this is that it'll silently ignore the `.jsonc` file if there's a `.json` with comments as well. Could you refactor this a bit more? Maybe a separate helper function which has a list of Path().extensions to check and raises if more than one is found.

HaoZeke (changes requested):

HaoZeke:
> 500ish stars on Microsoft's parser ([microsoft/node-jsonc-parser](https://github.com/microsoft/node-jsonc-parser)), so not widely known, but also not niche IMO.
> 
> My reason for opening this PR is that appending `c` to the file extension appears to be the idiomatic solution to all the red highlighting that appears on the GitHub interface. Unless I am missing a setting.
> 
> But no worries if this seems like more hassle than it is worth - I was just hoping that I could change to `jsonc` downstream without breaking peoples' local asv invocations.
> 
> The same applies to code blocks in conversation messages:
> 
> json:
> 
> ```json
> // a comment
> ```
> 
> jsonc:
> 
> ```js
> // a comment
> ```

OK then I'd say it'd be in scope, since anything to make using ASV more ergonomic is a win, left some suggestions for a refactor though.

lucascolley:
let me know if this is what you had in mind with `Path`!

asv/config.py
```diff
@@ -2,13 +2,44 @@
...
+            f"No `asv.conf` file found for valid extensions: {extensions}."
+        )
+    elif num_matches > 1:
+        raise util.UserError(
+            f"Multiple `asv.conf` files found for valid extensions: {extensions}. "
+            f"Please specifiy which file to use with `--config=FILEPATH`."
+        )
+    else:
+        # exactly one valid config file found
+        return path
```
lucascolley, line 40: Might need `str(path)`

HaoZeke: Yup I think `asv` expects path, but otherwise this looks great

asv/config.py
```diff
@@ -2,13 +2,44 @@
...
+    Raise if no file is found, or if multiple are found for the valid extensions.
+    Return the path of the config file if exactly one is found.
+    """
+    extensions = ['.json', '.jsonc']
+    path = Path('asv.conf')
+
+    num_matches = 0
+    for e in extensions:
+        p = path.with_suffix(e)
+        if p.exists:
```
lucascolley, line 25:
```suggestion
        if p.exists():
```

asv/config.py
lucascolley:
```suggestion
        return str(path)
```

asv/config.py
```diff
@@ -2,13 +2,44 @@
 
 import os
```
lucascolley, line 3:
```suggestion
```

lucascolley:
thanks @HaoZeke !

Let me know if I can help make an asv release happen, btw - I would like to patch SciPy once this is released.

HaoZeke:
> thanks @HaoZeke !
> 
> Let me know if I can help make an asv release happen, btw - I would like to patch SciPy once this is released.

Sure, I'm planning to cut a new release today, and add some documentation on how we manage them and what the tasks generally are (fairly standard but good to have written down). Maybe a `REL` issue or something.

PR #1421, opened 2024-08-11
BUG: Fix compare and continuous

### LINKED ISSUE #1414: QUERY: "Change" column not populated when running `asv compare`
Is the "Change" column supposed to be blank when running `asv compare`? I was under the impression that this would include the difference in the benchmark values between the two results. 

Below is a screenshot of a couple entries from when I ran `asv compare`. 

image

Here is the full output: https://github.com/UXARRAY/uxarray/pull/831#issuecomment-2201583704

### LINKED ISSUE #1417: MAINT: asv defaults to not using the expected module state / version
I have set up a minimal example here:

https://github.com/astrofrog/debug-asv

This contains a simple Python module 'mycode.py', and in the last commit (0b600d8f8a78ffaba69a8c102a2491282e08ac6e) I have added a sleep call in the ``myfunc`` function, which should slow down the benchmark. If I then run:

```
 asv continuous 0b43769a77c90d98c094bd1f6866f15eaffe2c58 0b600d8f8a78ffaba69a8c102a2491282e08ac6e
```

i.e. on the last two commits, I get:

```
Couldn't load asv.plugins._mamba_helpers because
No module named 'libmambapy'
· Fetching recent changes
· Fetching recent changes
· Creating environments
· Discovering benchmarks
·· Uninstalling from virtualenv-py3.12
·· Installing 0b600d8f <master> into virtualenv-py3.12
· Running 2 total benchmarks (2 commits * 1 environments * 1 benchmarks)
[ 0.00%] · For project commit 0b43769a <master~1> (round 1/2):
[ 0.00%] ·· Building for virtualenv-py3.12
[ 0.00%] ·· Benchmarking virtualenv-py3.12
[25.00%] ··· Running (benchmarks.time_myfunc--).
[25.00%] · For project commit 0b600d8f <master> (round 1/2):
[25.00%] ·· Building for virtualenv-py3.12
[25.00%] ·· Benchmarking virtualenv-py3.12
[50.00%] ··· Running (benchmarks.time_myfunc--).
[50.00%] · For project commit 0b600d8f <master> (round 2/2):
[50.00%] ·· Benchmarking virtualenv-py3.12
[75.00%] ··· benchmarks.time_myfunc                                                                                                                                                                                              1.00±0s
[75.00%] · For project commit 0b43769a <master~1> (round 2/2):
[75.00%] ·· Building for virtualenv-py3.12
[75.00%] ·· Benchmarking virtualenv-py3.12
[100.00%] ··· benchmarks.time_myfunc                                                                                                                                                                                              1.00±0s

BENCHMARKS NOT SIGNIFICANTLY CHANGED.
```

However, this is not right commit 0b43769a77c90d98c094bd1f6866f15eaffe2c58 does not contain the sleep call and should run a lot faster. It seems to be running the latest commit twice instead of the second two last commits. Am I using ``asv continuous`` incorrectly?

HaoZeke:
Closes #1417. Closes #1414 by improving the usage documentation.

### DIFF
--- a/asv/environment.py
+++ b/asv/environment.py
@@ -890,7 +890,9 @@ class Environment:
             # Run pip via python -m pip, avoids shebang length limit on Linux.
             # Don't run it in build directory, because it may contain Python packages
             # that pip believes to be already installed.
-            cmd = ["in-dir={env_dir} python -mpip install {wheel_file}"]
+            # --force-reinstall is needed since versions may not change between
+            # asv runs (esp. for compare), e.g. gh-1421
+            cmd = ["in-dir={env_dir} python -mpip install {wheel_file} --force-reinstall"]
 
         if cmd:
             commit_name = repo.get_decorated_hash(commit_hash, 8)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.7", "3.12", "pypy-3.8", "pypy-3.9"]
+        python-version: ["3.7", "3.12", "pypy-3.9"]
         r-version: ['release']
     timeout-minutes: 60
     steps:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,12 @@
+0.6.4 (TBD)
+------------------
+
+Other Changes and Additions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- ``asv`` defaults to ``--force-reinstall`` now to facilitate ``compare`` and ``continuous`` for manually versioned projects (#1421)
+
+
 0.6.3 (2024-02-25)
 ------------------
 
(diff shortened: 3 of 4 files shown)


asv/environment.py
```diff
@@ -890,7 +890,9 @@ def _install_project(self, repo, commit_hash, build_dir):
             # Run pip via python -m pip, avoids shebang length limit on Linux.
             # Don't run it in build directory, because it may contain Python packages
             # that pip believes to be already installed.
-            cmd = ["in-dir={env_dir} python -mpip install {wheel_file}"]
+            # --force-reinstall is needed since versions may not change between
+            # asv runs (esp. for compare), e.g. gh-1421
+            cmd = ["in-dir={env_dir} python -mpip install {wheel_file} --force-reinstall"]
```
HaoZeke, line 895: `--force-reinstall` (basically the same as `pip uninstall && pip install` is used here, but `-I` might also be the "right thing to do" for testing commits as long as the files have not changed.

HaoZeke: I'd have also been open to simply documenting the fix described in https://github.com/airspeed-velocity/asv/issues/1417#issuecomment-2282754045 but it seems like an issue a lot of people might have.

mattip:
I think it is fine to drop pypy3.8. It is no longer supported, and tests are timing out.

HaoZeke:
Makes sense, should be good to go now.

mattip:
Thanks @HaoZeke

PR #1424, opened 2024-08-12
REL: Prepare for `0.6.4`

HaoZeke:
This is a bugfix release, so not much else, I think the steps here are what I'll need to follow post-merge. @lucascolley let me know if this seems like a reasonable amount of detail.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -89,24 +89,3 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           path: dist/*.tar.gz
-
-  upload_pypi:
-    needs: [build_wheels, build_sdist]
-    runs-on: ubuntu-latest
-    environment:
-      name: pypi
-      url: https://pypi.org/p/asv_runner
-    permissions:
-      id-token: write  # for trusted publishing
-    # upload to PyPI on every tag starting with 'v'
-    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v') && github.repository == 'airspeed-velocity/asv'
-    # alternatively, to publish when a GitHub Release is created, use the following rule:
-    # if: github.event_name == 'release' && github.event.action == 'published'
-    steps:
-      - uses: actions/download-artifact@v3
-        with:
-          # unpacks default artifact into dist/
-          # if `name: artifact` is omitted, the action will create extra parent dir
-          name: artifact
-          path: dist
-      - uses: pypa/gh-action-pypi-publish@release/v1
--- a/asv/_rangemedian.cpp
+++ b/asv/_rangemedian.cpp
@@ -348,52 +348,14 @@ static PyMethodDef RangeMedian_methods[] = {
 
 static PyTypeObject RangeMedianType = {
     PyVarObject_HEAD_INIT(NULL, 0)
-    "RangeMedian",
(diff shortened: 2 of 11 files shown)


HaoZeke:
I'm open to removing `rangemedian` from here in this release and moving it to a new repo, or that can also wait on the next, this one has some important changes (including the new `.jsonc` support)

lucascolley:
sounds good to me, thanks!

asv/_rangemedian.cpp
```diff
@@ -348,52 +348,14 @@ static PyMethodDef RangeMedian_methods[] = {
...
-    0,                          // tp_del
-    0,                          // tp_version_tag
+    .tp_name="RangeMedian",
+    .tp_basicsize=sizeof(RangeMedianObject),
+    .tp_dealloc=(destructor)RangeMedian_dealloc,
+    .tp_flags=Py_TPFLAGS_DEFAULT,
+    .tp_doc = NULL,
+    .tp_methods=RangeMedian_methods,
+    .tp_init=(initproc)RangeMedian_init,
+    .tp_new=RangeMedian_new
```
mattip, line 358: Over [at cython](https://github.com/cython/cython/pull/5435#issuecomment-2090608598), using this in c++ requires c20, so it errored out on windows with `error C7555: use of designated initializers requires at least '/std:c++20'`. Does it work here?

.github/workflows/build_wheels.yml
```diff
@@ -89,24 +89,3 @@ jobs:
...
-    # alternatively, to publish when a GitHub Release is created, use the following rule:
-    # if: github.event_name == 'release' && github.event.action == 'published'
-    steps:
-      - uses: actions/download-artifact@v3
-        with:
-          # unpacks default artifact into dist/
-          # if `name: artifact` is omitted, the action will create extra parent dir
-          name: artifact
-          path: dist
-      - uses: pypa/gh-action-pypi-publish@release/v1
```
mattip, line 112: Are you doing this differently now?

mattip (approved):
LGTM, just a few questions.

asv/_rangemedian.cpp
HaoZeke: Designated initializers is a [C++20 feature](https://en.cppreference.com/w/cpp/language/aggregate_initialization) but also [C99 feature](https://rgambord.github.io/c99-doc/sections/6/7/8/index.html#p34), so here this should be fine (this is also now the [recommended style](https://docs.python.org/3/c-api/typeobj.html#examples) in the Python documentation). Although in the rewrite it would be much much better to cleanly separate the C++ parts (`vector` etc) from the Python bindings `C`.

.github/workflows/build_wheels.yml
HaoZeke: We never set up trusted publishing so this wasn't running, and shouldn't have been here. The release notes clarify this, explaining that we use `twine` for manual management.

I prefer manual management since you can test the wheels before uploading, the trusted publisher workflow means the only way to fix a bug is either a new release or a yank.

asv/_rangemedian.cpp
HaoZeke: Actually thinking about this a bit more I'm just going to revert the change here, there's a lot of weirdness here including the use of a template in a `.cpp` :'D

Issue #1425, opened 2024-08-13
JSONC support is bugged


lucascolley:
@HaoZeke there was an oversight in my PR - the default is actually defined at the level of the command as well as in `config.py`:

https://github.com/airspeed-velocity/asv/blob/bc2d07e8fab8926706e9d8ee7bf506dea6479b41/asv/commands/common_args.py#L26-L29

I think we should make it so that the command defaults to `None`, and let `config.py` handle the `None`?

HaoZeke:
> @HaoZeke there was an oversight in my PR - the default is actually defined at the level of the command as well as in `config.py`:
> 
> https://github.com/airspeed-velocity/asv/blob/bc2d07e8fab8926706e9d8ee7bf506dea6479b41/asv/commands/common_args.py#L26-L29
> 
> I think we should make it so that the command defaults to `None`, and let `config.py` handle the `None`?

Sure that makes sense. Sorry I didn't catch it in the review.

PR #1426, opened 2024-08-14
BUG: fix JSONC support

lucascolley:
@HaoZeke I think this should be it! close gh-1425

### DIFF
--- a/asv/commands/common_args.py
+++ b/asv/commands/common_args.py
@@ -26,7 +26,7 @@ def add_global_arguments(parser, suppress_defaults=True):
     parser.add_argument(
         "--config",
         help="Benchmark configuration file",
-        default=(argparse.SUPPRESS if suppress_defaults else 'asv.conf.json'))
+        default=(argparse.SUPPRESS if suppress_defaults else None))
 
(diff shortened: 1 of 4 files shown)


lucascolley:
CI isn't happy although I'm struggling to decipher the error messages.

HaoZeke (approved):

lucascolley:
I want to say the CI failure looks unrelated?

HaoZeke:
It is odd, especially since it isn't surprising at all that the extension (a **C-Python** extension) will fail on PyPy.. I'd have thought it wouldn't even try to build it there..

mattip:
TL;DR: Please pin setuptools to a known good version and only update it after checking that it works. 

This was reported as pypy/pypy#5004 and is due to a new setuptools release, which vendors pypa/distutils, which started requiring a default sysconfig value that PyPy did not supply pypa/distutils#280.

lucascolley:
CI is happy after pinning setuptools in this PR. This should be good to go.

HaoZeke:
Thanks @lucascolley and @mattip
Will cut another release tomorrow probably.

PR #1428, opened 2024-08-17
BUG: fix extraction of conda version

sfmig:
When creating a conda environment, the conda version is checked to determine the syntax of the command.

However, the version is currently extracted incorrectly. The command `conda --version` returns `conda xx.xx.xx` rather than the version string `xx.xx.xx` as the code assumes. 

As a result, for conda version `23.3.1`, we have `conda_version = "conda 23.3.1"` and `conda_version >= "24.3.0"` is evaluated as `True`, when it should be `False`.

This PR:
- adds a regexp to extract the conda version from the output of the `conda --version` command, and
- uses `packaging.version.Version` for a safer comparison of version strings.

It also pins setuptools following the recommendation given in PR #1426 to ensure CI passes.

**Question**: 
Should we test that environments are created correctly for different conda versions?

### DIFF
--- a/asv/plugins/conda.py
+++ b/asv/plugins/conda.py
@@ -5,6 +5,8 @@ import tempfile
 import contextlib
 from pathlib import Path
 
+from packaging.version import Version
+
 from .. import environment, util
 from ..console import log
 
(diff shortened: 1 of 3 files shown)


HaoZeke (approved):
LGTM, thanks, could you add yourself to the author list as well? That'd be in `docs/source/credits.rst`

HaoZeke:
Also, my apologies, I didn't see the question earlier, a test would be nice, but I think it can be added in later in a follow up (since I'm cutting a bugfix release now).

PR #1429, opened 2024-09-08
MAINT: Clean and update documentation

### LINKED ISSUE #1427: `asv dev` section still present in documentation after deprecation
The section for `asv dev` in the commands section is still present and hasn't been updated even though this feature has been deprecated. 

https://asv.readthedocs.io/en/stable/commands.html#asv-dev

Should this section be removed, or would it be better to add a warning (such as what NumPy does)

https://numpy.org/doc/1.25/reference/generated/numpy.find_common_type.html

image

Could include the following (taken from the RunetimeError when running `asv dev`

> asv dev has been removed. It was a shortcut of `asv run`, which you can use instead. It was removed because it caused confusion, in particular after changing what it was running in asv 0.5.x
> 
> You can be interested in the next `asv run` arguments:
>  - `--python=same`: to use the current environment and not create one
>  - `--quick`: to run benchmarks just once
>  - `--dry-run`: to not save the results

HaoZeke:
Closes #1427 and re-hauls the documentation for the new release. For the new release.


mattip:
This is a very large PR, I will find it hard to review. Is there something in particular you would like me to look at?

HaoZeke:
> This is a very large PR, I will find it hard to review. Is there something in particular you would like me to look at?

Ah sorry, I'll highlight a few places, in general there aren't too many differences, mostly cosmetic / documentation changes, that is:
- Uses `bibtex` for references, adds a logo (light/dark) and switched to `furo`
- Removed the spurious makefile stuff (which accounts for most of the deletions)
- Moved long module documentation (step detection) into its own page
- Uses `AutoAPI` and `viewcode` for more through documentation
  - This also means the `sphinx` build doesn't import `asv` anymore 
  - Uses `importlib_metadata` to consistently get the version as suggested by `setuptools_scm`

Here is what the documentation looks like at this commit:
https://tmpasv.surge.sh/

mattip:
It all seems quite straightforward, go ahead an merge if you are satisfied.

PR #1444, opened 2024-11-24
BUG: Fix multiple Python version usage

### LINKED ISSUE #1294: BUG: Package requires a different Python
Hi. Thanks for ASV. I'm really enjoying it. I have my conf and my suite of benchmarks mostly ready to go, but I have one issue

```
ERROR: Package 'mypackage' requires a different Python: 3.11.3 not in '<3.11,>=3.9'
```
In my conf, i have
```
    "environment_type": "conda",
    "pythons": ["3.9", "3.10"],
```
(among others) so this is surprising.

Indeed, mypackage is not supported for 3.11 and that information is in the pyproject.toml (which is what is apparently triggering the error on package install). I have older commits where this requires-python constraint was not present. In those cases asv runs the benchmarks, supposedly for python 3.10 (say) but if I insert `print(sys.version)` into a benchmark, i get back 3.11.3 in stdout. 

I thought maybe I'd installed asv improperly. So I went back to a fresh conda env and `pip install asv`. This step installed python 3.11 again and I have the same issues.  

Thanks in advance for clues on how I can resolve this.

HaoZeke:
Closes #1294. Basically ensures that the `asv.config` defined `python` takes precedence over the one written into an `environment.yml` file. This may cause resolution issues if there are a lot of pinned packages in the file, but that's basically user error.

Also fixes a subtle bug within the `mamba` plugin which caused incorrect environments to be created for multiple python versions.

### DIFF
--- a/asv/plugins/mamba.py
+++ b/asv/plugins/mamba.py
@@ -138,8 +138,8 @@ class Mamba(environment.Environment):
         env.update(self.build_env_vars)
         Path(f"{self._path}/conda-meta").mkdir(parents=True, exist_ok=True)
         if not self._mamba_environment_file:
-            # Construct payload, env file sets python version
-            mamba_pkgs = [f"python={self._python}", "wheel", "pip"]
+            # Construct payload
+            mamba_pkgs = ["wheel", "pip"]
         else:
             # For named environments
             env_file_name = self._mamba_environment_file
@@ -155,6 +155,11 @@ class Mamba(environment.Environment):
                 except KeyError:
                     raise KeyError("Only pip is supported as a secondary key")
         mamba_pkgs += mamba_args
+        # Changed in v0.6.5, gh-1294
+        # previously, the user provided environment was assumed to handle the python version
+        mamba_pkgs = [
+            f"python={self._python}" if a.startswith("python") else a for a in mamba_pkgs
+        ]
         self.context.prefix_params.target_prefix = self._path
         solver = MambaSolver(
             self._mamba_channels, None, self.context  # or target_platform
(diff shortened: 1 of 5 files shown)


asv/plugins/mamba.py
```diff
@@ -155,6 +155,12 @@ def _setup(self):
                 except KeyError:
                     raise KeyError("Only pip is supported as a secondary key")
         mamba_pkgs += mamba_args
+        # Changed in v0.6.5, gh-1294
+        # previously, the user provided environment was assumed to handle the python version
+        mamba_pkgs = [
+            f"python={self._python}" if a.startswith("python") else a for a in mamba_pkgs
+        ]
+        self.context.prefix_params.target_prefix = self._path
```
HaoZeke, line 163: This should probably be a more strict check than `startswith` to prevent inadvertently breaking things.

mattip:
Thanks @HaoZeke

asv/plugins/mamba.py
```diff
@@ -155,6 +155,12 @@ def _setup(self):
                 except KeyError:
                     raise KeyError("Only pip is supported as a secondary key")
         mamba_pkgs += mamba_args
+        # Changed in v0.6.5, gh-1294
+        # previously, the user provided environment was assumed to handle the python version
+        mamba_pkgs = [
+            util.replace_python_version(pkg, self._python) for pkg in mamba_pkgs
+        ]
```
jder, line 162: @HaoZeke @mattip sorry if I'm missing something, but it looks like if you do not have an environment file, the python version specified in asv.conf.json no longer gets passed to mamba after this change. (ie if there is no python= constraint before this line runs, we never add one) Perhaps we should restore the `python={self._python}` in the initialization of mamba_pkgs?

PR #1462, opened 2025-01-23
BUG: Apply ASV_PYTHONPATH to discovery

HaoZeke:
Companion to comparing across environments with `asv-spyglass` for situations like `scipy` where `dev.py` sets `PYTHONPATH`.


mattip:
Still draft?

mattip:
Is there a test that fails before, passes after this change?

PR #1467, opened 2025-02-18
docs: document the `version` field in `asv.conf.json.rst`

JuroOravec:


docs/source/asv.conf.json.rst
```diff
@@ -22,6 +22,10 @@ this file and their expected values.
 
 .. only:: not man
 
+``version``
+-----------
+Version of the `asv.conf.json` spec. Currently only `1`.
```
HaoZeke, line 27: Could you add the version instead of currently? i.e. as of version ... since it might change.

HaoZeke:
CI failure is unrelated here, will track it separately.

docs/source/asv.conf.json.rst
JuroOravec: Sounds good, updated!

HaoZeke (approved):

PR #1474, opened 2025-02-24
add space between python & args

KRRT7:
this messed me up when copy pasting :)


DimitriPapadopoulos:
I created the new PR #1487 to change all occurrences of `-mod` to `-m mod`.

But then both should work. Doesn't `python -mod` work in your case?

mattip:
Closing, this was done in #1487. Thanks KRRT7.

Issue #1484, opened 2025-05-06
BUG: virtualenv-20.31.0 breaks `asv setup` command with error "virtualenv: error: unrecognized arguments: --wheel=bundle"


JuroOravec:
Latest virtualenv version ([v20.31.0](https://virtualenv.pypa.io/en/latest/changelog.html#v20-31-0-2025-05-05)) drops the support for the `--wheel=bundle` argument on (I believe) py3.12+, resulting in following error when running `asv setup`.

```txt
Couldn't load asv.plugins._mamba_helpers because
No module named 'libmambapy'
· Creating environments
·· Creating virtualenv for virtualenv-py3.13-django5.1-djc-core-html-parser
·· Running '/opt/hostedtoolcache/Python/3.13.3/x64/bin/python -mvirtualenv --wheel=bundle --setuptools=bundle -p /opt/hostedtoolcache/Python/3.13.3/x64/bin/python3.13 /home/runner/work/django-components/django-components/.asv/env/b42e02a1e4eb987ea13a52915b8134f0'
   OUTPUT -------->
   usage: virtualenv [--version] [--with-traceback] [-v | -q] [--read-only-app-data] [--app-data APP_DATA] [--reset-app-data] [--upgrade-embed-wheels] [--discovery {builtin}] [-p py] [--try-first-with py_exe]
                     [--creator {builtin,cpython3-posix,venv}] [--seeder {app-data,pip}] [--no-seed] [--activators comma_sep_list] [--clear] [--no-vcs-ignore] [--system-site-packages] [--symlinks | --copies] [--no-download | --download]
                     [--extra-search-dir d [d ...]] [--pip version] [--setuptools version] [--no-pip] [--no-setuptools] [--no-periodic-update] [--symlink-app-data] [--prompt prompt] [-h]
                     dest
   SystemExit: 2

   virtualenv: error: unrecognized arguments: --wheel=bundle
·· Error running /opt/hostedtoolcache/Python/3.13.3/x64/bin/python -mvirtualenv --wheel=bundle --setuptools=bundle -p /opt/hostedtoolcache/Python/3.13.3/x64/bin/python3.13 /home/runner/work/django-components/django-components/.asv/env/b42e02a1e4eb987ea13a52915b[81](https://github.com/django-components/django-components/actions/runs/14843081536/job/41703441498?pr=1167#step:6:82)34f0 (exit status 2)
·· Failure creating environment for virtualenv-py3.13-django5.1-djc-core-html-parser
[...]

PR #1485, opened 2025-05-06
refactor: fix virtualenv setup for py3.9+

JuroOravec:
Closes https://github.com/airspeed-velocity/asv/issues/1484


CHANGES.rst
```diff
@@ -1,3 +1,11 @@
+0.6.5 (2025-05-06)
+------------------
+
+Bug Fixes
+^^^^^^^^^
+
+- Fix virtualenv setup for py3.12+ (#1485)
```
DimitriPapadopoulos, line 7:
```suggestion
- Fix virtualenv setup for virtualenv 20.31 (#1485)
- ```

File: asv/plugins/virtualenv.py [hunk 2]
```diff
@@ -133,7 +133,9 @@ def _setup(self):
         util.check_call([
             sys.executable,
             "-mvirtualenv",
-            "--wheel=bundle",
+            # Omit `--wheel=bundle` from py3.9+
+            # See https://github.com/airspeed-velocity/asv/issues/1484
+            *(["--wheel=bundle"] if sys.version_info < (3, 9) else []),
```
DimitriPapadopoulos, line 138: Don't make a special case for Python < 3.9: setuptools [20.31](https://virtualenv.pypa.io/en/latest/changelog.html#v20-31-0-2025-05-05) removes option `--wheel` entirely, regardless of the version of Python, as can be seen in https://github.com/pypa/virtualenv/pull/2868:

https://github.com/pypa/virtualenv/pull/2868/files#diff-34e3905158c5a4b777ef45dd13670361f3dd81c02aa4bf0d36f3e5cac777ee6a

DimitriPapadopoulos (commented):
I recommend removing this line:
```python
            "--wheel=bundle",
```
Not even sure this is needed at all, but you could require setutptools ≥ [20.31](https://virtualenv.pypa.io/en/latest/changelog.html#v20-31-0-2025-05-05) to be on the safe side:
https://github.com/airspeed-velocity/asv/blob/01f7cfa3da86f6d98ef68984ecc663ac6ff7dfaa/pyproject.toml#L34
```diff
-    "virtualenv",
+    "virtualenv>=20.31",
```

asv/plugins/virtualenv.py
DimitriPapadopoulos: Wait, the special case for Python 3.8 is actually part of https://github.com/pypa/virtualenv/pull/2876 which partially reverts https://github.com/pypa/virtualenv/pull/2868:

https://github.com/pypa/virtualenv/pull/2876/files#diff-34e3905158c5a4b777ef45dd13670361f3dd81c02aa4bf0d36f3e5cac777ee6a

The complete release notes can be found here:
https://github.com/pypa/virtualenv/releases/tag/20.31.0
**not** here:
https://virtualenv.pypa.io/en/latest/changelog.html#v20-31-0-2025-05-05

This is really confusing.

JuroOravec: Yeah, I'm confused. However, Python 3.8 has reached its end of life in [late 2024](https://devguide.python.org/versions/), so perhaps we can drop support for 3.8 and simply remove the option?

Don't know how others use `asv`, but at least in django-components the situation is that:
- The library still supports py3.8, because Django 4.2 still does so.
- However, `asv` is not part of the distribution, so we run benchmarks only during development and CI, and thoes run only on the latest Python version.

DimitriPapadopoulos:
About the CI error, see https://github.com/rpy2/rpy2/issues/1164.

asv/plugins/virtualenv.py
DimitriPapadopoulos: asv supports Python 3.7 but I don't know if that's on purpose.

JuroOravec: Ok, in that case let's just play it safe and just check for the version of `virtualenv`:

```py
from importlib.metadata import version
version('virtualenv')
```

See https://stackoverflow.com/a/17767076/9788634

But unfortunately the `version()` function returns version as plain string, and there are some cases where `virtualenv` released beta versions like [`20.0.0b2`](https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-0b2-2020-02-04).

Then again, we really need to check only major / minor parts, so something like this could work:

```py
from importlib.metadata import version
venv_version = map(int, version('virtualenv').split(".")[:2])
use_wheel = venv_version <= (20, 31)

util.check_call([
    sys.executable,
    "-mvirtualenv",
    # Omit `--wheel=bundle` from virtualenv v20.31 and later
    # See https://github.com/airspeed-velocity/asv/issues/1484
    *(["--wheel=bundle"] if use_wheel else []),
    "--setuptools=bundle",
    "-p",
    self._executable,
    self._path
], env=env)
```

HaoZeke (approved):
LGTM, thanks.

PR #1501, opened 2025-05-21
MAINT: Enforce ruff/pyupgrade rules (UP)

DimitriPapadopoulos:
The CI failure seems unrelated:
```
=========================== short test summary info ============================
FAILED test/test_run.py::test_run_spec - Failed: Timeout (>300.0s) from pytest-timeout.
============ 1 failed, 159 passed, 94 skipped in 1488.70s (0:24:48) ============
```

Maybe related to:
https://github.com/airspeed-velocity/asv/blob/d5f068b41018bbf73247551188ef0d0805d83784/.github/workflows/ci_win.yml#L36
https://github.com/airspeed-velocity/asv/blob/d5f068b41018bbf73247551188ef0d0805d83784/.github/workflows/triggered.yml#L60
PyPy runs might need a higher timeout than `300`.

### DIFF
--- a/asv/plugins/summarylist.py
+++ b/asv/plugins/summarylist.py
@@ -23,7 +23,8 @@ def benchmark_param_iter(benchmark):
     if not benchmark['params']:
         yield None, ()
     else:
-        yield from enumerate(itertools.product(*benchmark['params']))
+        for item in enumerate(itertools.product(*benchmark['params'])):
+            yield item
 
 
 class SummaryList(OutputPublisher):
--- a/asv/plugins/summarylist.py
+++ b/asv/plugins/summarylist.py
@@ -23,8 +23,7 @@ def benchmark_param_iter(benchmark):
     if not benchmark['params']:
         yield None, ()
     else:
-        for item in enumerate(itertools.product(*benchmark['params'])):
-            yield item
+        yield from enumerate(itertools.product(*benchmark['params']))
 
 
 class SummaryList(OutputPublisher):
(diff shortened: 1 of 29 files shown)


asv/plugins/summarylist.py
```diff
@@ -23,8 +23,7 @@ def benchmark_param_iter(benchmark):
     if not benchmark['params']:
         yield None, ()
     else:
-        for item in enumerate(itertools.product(*benchmark['params'])):
-            yield item
+        yield from enumerate(itertools.product(*benchmark['params']))
```
mattip, line 26: Hmm, is there a way to leave this? The implicit for loop is tricky here.

asv/util.py
```diff
@@ -819,8 +818,7 @@ def iter_subclasses(cls):
     """
     for x in cls.__subclasses__():
         yield x
-        for y in iter_subclasses(x):
-            yield y
+        yield from iter_subclasses(x)
```
mattip, line 821: Is there a way to leave the explicit loop?

asv/plugins/summarylist.py
DimitriPapadopoulos: Do you mean this change breaks the code? Both version should be functionally equivalent; `yield from` is more efficient because the loop is implemented inside Python itself.

asv/util.py
DimitriPapadopoulos: I can add `noqa: B028` as well.

But what's wrong with `yield from`?

asv/plugins/summarylist.py
DimitriPapadopoulos: I've reverted this change and disabled rule [`UP028`](https://docs.astral.sh/ruff/rules/yield-in-for-loop/#yield-in-for-loop-up028).

mattip:
Thanks @DimitriPapadopoulos

PR #1502, opened 2025-05-21
MAINT: Enforce ruff/flake8-bugbear rules (B)

DimitriPapadopoulos:


asv/template/benchmarks/benchmarks.py
```diff
@@ -13,11 +13,11 @@ def setup(self):
             self.d[x] = None
 
     def time_keys(self):
-        for key in self.d.keys():
+        for _ in self.d.keys():
```
mattip, line 16: Couldn't this be simplified to 
```
def time_keys(self):
    pass
```

Same with `time_values` below

mattip (approved):
LGTM, one small nit. It is fine to do it in another PR

DimitriPapadopoulos:
Cannot make sense of the CI error yet: 
```
FAILED test/test_continuous.py::test_continuous - KeyError: 'environments'
```

asv/template/benchmarks/benchmarks.py
mattip: ahh, right, sorry.

Issue #1453, opened 2024-12-18
"Failed to build the project and import the benchmark suite" with no apparent errors in output


sflis:
I was trying out `asv` with my project but failed to get it event to run getting this output 

``` 
$asv run
· Creating environments
· Discovering benchmarks
·· Uninstalling from virtualenv-py3.12
·· Building 698da480 <CAV-101376-expose-encoding-field> for virtualenv-py3.12...
·· Installing 698da480 <CAV-101376-expose-encoding-field> into virtualenv-py3.12
·· Failed to build the project and import the benchmark suite. 
```
when running in verbose mode I'm getting this: 

```
   Building wheels for collected packages: dcap
     Building wheel for dcap (pyproject.toml): started
     Building wheel for dcap (pyproject.toml): finished with status 'done'
     Created wheel for dcap: filename=dcap-0.0.4.post17+git.698da480-py3-none-any.whl size=22315 sha256=a208536bddb65150a963462139a4156ab570ce73abc7129e43a41357908e501e
     Stored in directory: /tmp/pip-ephem-wheel-cache-3qlrv7ij/wheels/d6/c4/aa/2fddef2f981ba26e6e5d96b79c993cb2203ead973b3aa818db
   Successfully built dcap
·· Running '/usr/bin/git name-rev --name-only --exclude=remotes/* --no-undefined 698da480f911c26d790dc1062f19da2d6ea96690'
   OUTPUT -------->
   CAV-101376-expose-encoding-field
·· Installing 698da480 <CAV-101376-expose-encoding-field> into virtualenv-py3.12
·· Failed to build the project and import the benchmark suite.
```
with still no apparent errors, this includes the previous steps omitted from the output shown above. What would be the next step to track down the issue?

HaoZeke:
Can you provide more details for reproducibility? Crucially, the repo (if it is open source) and also if the command used to install the project normally works outside ASV.

bdice:
I'm seeing the same problem, and another user reported the same. https://github.com/glotzerlab/signac/pull/1017#issuecomment-2634745588

`$ asv run` in https://github.com/glotzerlab/signac/ is enough to reproduce the failure.

Here is a more extensive reproducer with an older Python version (3.10 instead of 3.12):
```bash
git clone https://github.com/glotzerlab/signac
cd signac
conda create -n py310 python=3.10
conda activate py310
pip install -e .
pip install asv
asv run
```

Output:
```
$ asv run
[...]

PR #1505, opened 2025-05-25
Report the last error when failing to discover benchmarks

TalAmuyal:
While trying to use asv for the first time, I've got to a situation that was hard to debug without cloning the repo.
In this PR I'm trying to make it easier for the next person that makes the same (or similar) config mistake I did.

This might be related to https://github.com/airspeed-velocity/asv/issues/1453.

When running `asv run -v`, I got:

```
      Compiling geo-trace v0.0.1 ([...]/geo_trace/.asv/env/28ce2c79fdbca74891d3623705fc0783/project)
       Finished `release` profile [optimized] target(s) in 37.98s
   🚉 Overriding platform tag from _PYTHON_HOST_PLATFORM environment variable as linux_x86_64.
   📖 Found type stub file at geo_trace.pyi
   📦 Built wheel for CPython 3.10 to [...]/geo_trace/.asv/env/28ce2c79fdbca74891d3623705fc0783/project/geo_trace-0.0.1-cp310-cp310-linux_x86_64.whl
·· Running '[...]/bin/git name-rev --name-only --exclude=remotes/* --no-undefined 9851fde784e599ea9d027bf38aa351bb0b6284fd'
   OUTPUT -------->
   master
·· Installing 9851fde7 <master> into virtualenv-py3.10
```

After about 2 hours of tweaking configs and trying to debug it from the outside, I've cloned asv and installed it locally as editable so I could debug it from the inside.

Eventually, I've added:

```python
if last_err is not None:
    log.error("Failed to discover benchmarks: " + str(last_err))
```
just before `raise util.UserError("Failed to build the project and import the benchmark suite.")`, and got:
```
·· Failed to discover benchmarks: Configuration error: {wheel_file} not available when substituting into command 'in-dir={env_dir} python -m pip install {wheel_file} --force-reinstall' Available: {'project': 'geo-trace', 'conf_dir': '[...]/geo_trace', 'env_name': 'virtualenv-py3.10', 'env_dir': '[...]/geo_trace/.asv/env/28ce2c79fdbca74891d3623705fc0783', 'env_type': 'virtualenv', 'commit': '9851fde784e599ea9d027bf38aa351bb0b6284fd', 'build_dir': '[...]/geo_trace/.asv/env/28ce2c79fdbca74891d3623705fc0783/project', 'build_cache_dir': '[...]/geo_trace/.asv/env/28ce2c79fdbca74891d3623705fc0783/asv-build-cache/9851fde784e599ea9d027bf38aa351bb0b6284fd'}
```

This eventually led me to the understanding that the build command should use `build_cache_dir` instead of `build_dir` (the full command in my use-case now is `python -m maturin build --release --out {build_cache_dir}`).

### DIFF
--- a/asv/benchmarks.py
+++ b/asv/benchmarks.py
@@ -211,6 +211,8 @@ class Benchmarks(dict):
                 finally:
                     util.long_path_rmtree(result_dir)
             else:
+                if last_err is not None:
+                    log.error("Last error: " + str(last_err))
                 raise util.UserError("Failed to build the project and import the benchmark suite.")
 
         if check:


TalAmuyal:
Well, I don't really have the context for why it was made that way.
My main action-item here though, is that there needs to be some output that describes the problem.
If we can remove a bunch of code and replace it with a config, then I'm all in.
If we prefer to print something, just LMK what should be the logic and I'll do my best.

PR #1533, opened 2025-09-13
DOC: Add more on releases

HaoZeke:
Just some more fragments from the release, not urgent.


mattip:
CI failure unrelated. We should update the macos image from maocs-13 since it is going out of support soon.

PR #1539, opened 2025-09-14
before cd, change arg.config to full path

### LINKED ISSUE #1538: Cannot find `benchmarks/asv.conf.json` even though it exists on 0.6.5
On 0.6.5, [NumPy CI cannot run ASV](https://github.com/numpy/numpy/actions/runs/17712294880/job/50332944016#step:6:2). It works locally (with 0.6.5), when running
```
asv machine --yes --config benchmarks/asv.conf.json
```

mattip:
Closes #1538

In #1426 the logic changed, and providing `--config <file>` as a cmd-line argument was subtly changed. Now the path of that file is used as the cwd, but that path is not changed to accomodate the change in directory. A line that changed the value of `args.config` to the absolute path of `<file>` was removed. This PR restores it, and adds a test mimicing the failure in #1538. I verified that it fails before, passes after this PR.

### DIFF
--- a/asv/main.py
+++ b/asv/main.py
@@ -22,6 +22,7 @@ def main():
     # the run.
     # If using the default path, stay in the same dir.
     if args.config:
+        args.config = os.path.abspath(args.config)
         os.chdir(os.path.dirname(os.path.abspath(args.config)))
 
     try:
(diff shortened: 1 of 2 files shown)


mattip:
I also took the opportunity to 
- drop python3.7 and add python3.13 to the declared supported versions
- update r-tools to `release`, hopefully it will JustWork on macos-13 and not conflict with rpy2.

HaoZeke (approved):

PR #1544, opened 2025-09-27
Add plugin for uv to create virtualenvs

TimPansino:
## Overview

* Add plugin to create virtual environments with [uv](https://docs.astral.sh/uv/).
  * Note: The package resolver and installer used is still `pip` and this does not use `uv pip`.
* Help needed for adding tests. This PR has no tests currently was only tested on MacOS and not Linux or Windows.

### DIFF
--- a/asv/plugins/uv.py
+++ b/asv/plugins/uv.py
@@ -1,0 +1,137 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+import os
+import re
+import sys
+
+from .. import environment, util
+from ..console import log
+
+WIN = os.name == "nt"
+
+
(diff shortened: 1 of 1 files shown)


TimPansino:
I've rebased it

mattip:
Pending CI success, i think this is good. Woyld be nice to have @HaoZeke take a look

TimPansino:
Any update on this? @HaoZeke @mattip

mattip:
I see the uv environment is [passing the test](https://github.com/airspeed-velocity/asv/actions/runs/18247609650/job/51958750182?pr=1544#step:11:82) so it seems to be working. Let's put it in, we can iterate on the details as needed.

PR #1551, opened 2025-10-09
CI: Python 3.14

DimitriPapadopoulos:


mattip:
Close/reopen to retrigger CI against HEAD

.github/workflows/build_wheels.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           - [macos-15-intel, macosx_x86_64]
           - [macos-15, macosx_arm64]
           - [windows-2022, win_amd64]
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', 'pypy3.10']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14', 'pypy3.10']
```
mattip:
```suggestion
        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
```

mattip:
A test is failing
> `python -m pip install -v --upgrade asv_dummy_test_package_1' returned non-zero exit status 1`

.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-latest"]
-        python-version: ["3.9", "3.13", "pypy-3.11"]
+        python-version: ["3.9", "3.14", "pypy-3.11"]
```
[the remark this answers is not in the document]
DimitriPapadopoulos: I wonder why the test fails on `ubuntu` but not `macos`. And only with Python 3.14, not Python 3.13.

The failure occurs in `test/test_run.py::test_parallel`. I suspect `asv_dummy_test_package_1` might not be ready yet when pip attempts to install it. Improvements in Python 3.14 might uncover an existing race condition in the test.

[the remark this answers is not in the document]
DimitriPapadopoulos: With Python 3.13, I see an interesting `multiprocessing` warning:
```
=============================== warnings summary ===============================
test/test_gh_pages.py: 64 warnings
test/test_publish.py: 144 warnings
test/test_run.py: 20 warnings
test/test_util.py: 8 warnings
  /opt/hostedtoolcache/Python/3.13.11/x64/lib/python3.13/multiprocessing/popen_fork.py:67: DeprecationWarning: This process (pid=2442) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()
``

**Comment by DimitriPapadopoulos** [comment]
I think asv shouldn't mix threads and fork in general. From the [`os.fork()`](https://docs.python.org/3/library/os.html#os.fork) documentation:
> _Changed in version 3.12:_ If Python is able to detect that your process has multiple threads, [`os.fork()`](https://docs.python.org/3/library/os.html#os.fork) now raises a [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning).
> 
> We chose to surface this as a warning, when detectable, to better inform developers of a design problem that the POSIX platform specifically notes as not supported. Even in code that _appears_ to work, it has never been safe to mix threading with [`os.fork()`](https://docs.python.org/3/library/os.html#os.fork) on POSIX platforms. The CPython runtime itself has always made API calls that are not safe for use in the child process when threads existed in the parent (such as `malloc` and `free`).
> 
> Users of macOS or users of libc or malloc implementations other than those typically found in glibc to date are among those already more likely to experience deadlocks running such code.
> 
> See [this discussion on fork being incompatible with threads](https://discuss.python.org/t/33555) for technical details of why we’re surfacing this longstanding platform compatibility problem to developers.

Again, I don't know enough about asv internals to understand why threads and fork were introduced, and decide whether to get rid of threads or fork.

**Comment by DimitriPapadopoulos** [comment]
Also interesting, from the [documentation](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) of the _fork_ start method:
>  _Changed in version 3.14:_ This is no longer the default start method on any platform. Code that requires _fork_ must explicitly specify that via [`get_context()`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_context) or [`set_start_method()`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method).
> 
> _Changed in version 3.12:_ If Python is able to detect that your process has multiple threads, the [`os.fork()`](https://docs.python.org/3/library/os.html#os.fork) function that this start method calls internally will raise a [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning). Use a different start method. See the [`os.fork()`](https://docs.python.org/3/library/os.html#os.fork) documentation for further explanation.

and the _forkserver_ start method:
> _Changed in version 3.14:_ This became the default start method on POSIX platforms.

and moreover:
> _Changed in version 3.8:_ On macOS, the _spawn_ start method is now the default. The _fork_ start method should be considered unsafe as it can lead to crashes of the subprocess as macOS system libraries may start threads. See [bpo-33725](https://bugs.python.org/issue?@action=redirect&bpo=33725).
> 
> _Changed in version 3.14:_ On POSIX platforms the default start method was changed from _fork_ to _forkserver_ to retain the performance but avoid common multithreaded process incompatibilities. See [gh-84559](https://github.com/python/cpython/issues/84559).

Yet, in the case of asv, _forkserver_ appears to uncover more multithreaded process incompatibilities than _fork_.

File: asv/util.py [hunk 3]
```diff
@@ -1266,9 +1266,11 @@ def truncate_float_list(item, digits=5):
 _global_locks = {}
 
 
-def _init_global_locks(lock_dict):
-    """Initialize global locks in a new multiprocessing process"""
+def _init_global_locks(lock_dict, env):
+    """Initialize global locks in a new multiprocessing process
+    Also inherit the base environment even if using a forkserver"""
     _global_locks.update(lock_dict)
+    os.environ.update(env)
```
DimitriPapadopoulos: I understand this is the actual fix. I still don't get why it occurs only with Python 3.14. A change in the `multiprocessing` package?

asv/runner.py
```diff
@@ -421,12 +421,8 @@ def get_spawner(env, benchmark_dir, launch_method):
...
 
     if launch_method in (None, 'auto'):
-        # Don't use ForkServer as default on OSX, because many Apple
-        # things are not fork-safe
-        if has_fork and sys.platform not in ('darwin',):
-            launch_method = "forkserver"
-        else:
-            launch_method = "spawn"
+        # Don't use ForkServer as default as threading and fork are not compatible
+        launch_method = "spawn"
```
DimitriPapadopoulos: Do we keep `spawn` as the default in all cases, as suggested by the `fork()` documentation, although it's slower?

mattip: I think we should. The time of `spawn` should only be important when running tests, in "real" usage the benchmark run time should be much more than the difference between "spawn" and "fork"

PR #1558, opened 2025-11-24
Fix pip requirement building in new environments

hmaarrfk:
By default it seems that:

- pip+build
- pip+asv_runner
- pip+pympler

make it to the requirements of building your conda environment

Without this PR, the command gets executed and adds things like like a bogus environment.yml (temporary only) that looks lik

```
name: conda-py3.12
channels:
   - conda-forge
```

but this causes conda to error since there are no requirements for the "update".

We can debate as to whether or not these "pip dependencies" should be converted to conda ones, but lets leave that debate for later. This helps people run "asv setup"

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         os: ["ubuntu-latest", "macos-latest"]
         python-version: ["3.9", "3.13", "pypy-3.11"]
-    timeout-minutes: 60
+    timeout-minutes: 80
     steps:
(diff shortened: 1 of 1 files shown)


hmaarrfk:
is the rattler failure really from me?

hmaarrfk:
(i just rebased onto main since i saw that other activity may have made the CIs green)

mattip:
I think the timeout is being exceeded on a slow machine. Could you increase it for the macos tests?

mattip:
Thanks @hmaarrfk

PR #1560, opened 2025-11-25
Allow users to specify the launch_method in the config

hmaarrfk:
We find that with cupy at least, we are unable to allocate GPU memory without setting the method to spawn.

This is easy to forget to specify, and thus creates all these problems for us when we run longer benchmarks

Thanks for considering!


hmaarrfk:
Thank you. 

We might need a few more followups if that is ok. 

It might be that i didnt catch all ghe different codepatha

PR #1564, opened 2025-12-01
update cibuildwheel [wheel build]

mattip:
Replaces #1556, #1563 with the `[wheel build]` tag on the commit message


mattip:
Cool. The abi3 wheels work, moving the extension to heap types and using module state.

PR #1565, opened 2025-12-06
Fixup how the default launch method is chosen

hmaarrfk:
As a followup to #1560 this actually removes the "launch-method" from the args allowing it to be specified by the config file.

### DIFF
--- a/asv/commands/common_args.py
+++ b/asv/commands/common_args.py
@@ -289,8 +289,13 @@ def add_launch_method(parser):
         dest="launch_method",
         action="store",
         choices=("auto", "spawn", "forkserver"),
-        default="auto",
-        help="How to launch benchmarks. Choices: auto, spawn, forkserver",
+        default=None,
+        help=(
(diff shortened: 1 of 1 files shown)


hmaarrfk:
https://github.com/airspeed-velocity/asv/pull/1560/files#diff-2b2a4513be9046dbc9072fa60f71ce9a0606ba9001de77415d60523a68bbf841R443

Shows that we try to see if auto is specified.

hmaarrfk:
cc: @gianmarco-terrones

gianmarco-terrones:
It would be great if somewhere in the asv docs at https://asv.readthedocs.io/en/stable/index.html# it be mentioned that spawn is recommended for benchmarks involving CUDA/GPUs (even though spawn has more overhead associated it)

gianmarco-terrones:
Otherwise, running asv on CUDA-dependent benchmarks leads to unclear error messages

mattip:
Thanks @hmaarrfk

PR #1571, opened 2025-12-14
Remove per-file license headers

### LINKED ISSUE #1570: Use SPDX short-form identifiers in source file headers
How about changing header of source files from:
```python
```
to:
```python
# SPDX-License-Identifier: BSD-3-Clause
```

DimitriPapadopoulos:
Fixes #1570.

### DIFF
--- a/asv/commands/__init__.py
+++ b/asv/commands/__init__.py
@@ -1,4 +1,4 @@
-# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# SPDX-License-Identifier: BSD-3-Clause
 
 import abc
 import argparse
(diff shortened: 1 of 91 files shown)


asv/commands/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# SPDX-License-Identifier: BSD-3-Clause
```
mattip, line 1: I think we can remove the per-file license statement. The default is that the project license covers all the files, so only if there is a different license do we need a mention.

HaoZeke (changes requested):
I think I'd prefer @mattip's note about deleting these outright, I don't think they're very useful TBH

DimitriPapadopoulos:
Should be OK now after some rebasing.

Ci failures are unrelated.

PR #1575, opened 2026-01-30
Add --only-improved and --only-regressed flags for filtering benchmark results

### LINKED ISSUE #1496: Add built-in support to display only improved or regressed benchmarks
The current asv compare command provides detailed benchmark results, and the --split option already separates benchmarks into different tables for improved, unchanged, and worst-performing benchmarks. Additionally, the --only-changed option shows benchmarks that have changed between commits.

However, there is no built-in option to directly show only improved or only regressed benchmarks in a streamlined way. It would be extremely helpful to add flags like --only-improved and --only-regressed, allowing users to quickly filter and view only the benchmarks that have improved or regressed.

vprayag2005:
closes #1496

This PR added 2 flags, --only-improved and --only-regressed.With these flags, it easy to see which benchmarks has improved and which has gotten worse.


vprayag2005:
I hope this PR is useful, since there was no objection about this in the issue. The main reason why I am adding this is that in PR we can give comments like your PR caused regressions.

mattip:
> we can give comments like your PR caused regressions

Why would a reviewer only be interested in regressions, and not improvements?

vprayag2005:
That's just an example I gave. Most PRs mainly need regression testing, such as bug fixes, etc. Performance tests are usually only needed for PRs focused on performance improvements.

mattip:
I am trying to understand the motivation for the issue and the PR. The issue pointed out the ability to do result manipulation in spyglass, which would be easy to do in a CI environment. Where would this feature be useful, if not for PRs?

vprayag2005:
This feature isn't just for PR comments. After development,  we can run asv compare locally to check for regressions or verify if performance actually improved.

I don't know much about Spyglass, but as mentioned in the issue, having to install two packages just for basic filtering feels like extra work. We already have --only-changed and --split, but there's no simple option to get 'only regressed' or 'only improved.' Adding this directly to the compare command would be very helpful for everyday use.

trexfeathers:
> I am trying to understand the motivation for the issue and the PR. The issue pointed out the ability to do result manipulation in spyglass, which would be easy to do in a CI environment. Where would this feature be useful, if not for PRs?

My team and I were recently talking about how this functionality could enable us writing some simple automation highlighting regressions from previous benchmark results, without needing to call into ASV's internals.

That's not the same as "you must do this" - I understand the need to manage API bloat - I just wanted to share one possible motivation.

HaoZeke (changes requested):
I don't personally see the need for 2 new CLI flags just for this functionality... `asv_spyglass` is the right place for this IMO. Having a second package even in CI, especially a pure python one barely makes a difference in any environment. Without more clarifying use cases, I'm inclined to close this.

vprayag2005:
This is basic filtering. I don't think for this basic filtering we need to install two packages. I'm not sure if asv_spyglass has enough documentation or if users will know about it.

I think both maintainers have a similar opinion, but maybe adding it there will be good rather than adding more similar kind of flags.

However, if that is the case, I will close this and add it there in asv_spyglass.

vprayag2005:
closing since i have opened in https://github.com/airspeed-velocity/asv_spyglass/pull/10

PR #1580, opened 2026-03-08
ENH: Add free-threaded Python support

### LINKED ISSUE #1515: Support the free-threaded build and publish free-threaded wheels
When I try to use asv to run the numpy benchmarks on the free-threaded build, I see:

```
<frozen importlib._bootstrap>:488: RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'asv._rangemedian', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```

HaoZeke:
Declare `Py_mod_gil` with `Py_MOD_GIL_NOT_USED` so the extension does not force-enable the GIL on free-threaded builds.  The module holds no global mutable state (all data lives in per-object or per-module-state structs), so running without the GIL should be safe...

Enable free-threaded wheel builds in cibuildwheel so cp313t and cp314t wheels are published.

Closes #1515.

### DIFF
--- a/asv/_rangemedian.cpp
+++ b/asv/_rangemedian.cpp
@@ -92,12 +92,10 @@ private:
     };
 
     std::map<std::pair<size_t, size_t>, Item> items_;
-    mutable std::mutex mtx_;
 
 public:
     Cache() {};
     bool get(size_t left, size_t right, double *mu, double *dist) const {
-        std::lock_guard<std::mutex> lock(mtx_);
         auto it = items_.find(std::make_pair(left, right));
         if (it != items_.end()) {
             *mu = it->second.mu;
@@ -108,7 +106,6 @@ public:
     }
 
     void set(size_t left, size_t right, double mu, double dist) {
-        std::lock_guard<std::mutex> lock(mtx_);
         items_[std::make_pair(left, right)] = { mu, dist };
     }
 };
@@ -127,6 +124,9 @@ typedef struct {
     PyObject_HEAD
     std::vector<std::pair<double,double> > *y;
     Cache *cache;
+#ifdef Py_GIL_DISABLED
+    std::mutex *use_mtx;
+#endif
 } RangeMedianObject;
 
 #define RangeMedianObject_CAST(op)  ((RangeMedianObject *)(op))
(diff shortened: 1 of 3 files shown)


HaoZeke:
cc @ngoldbaum too for the assessment over in #1515 (nvm the failures they're from #1581)

mattip (approved):
LGTM

rgommers:
It'd probably be good to drop `cp313t` now, those wheels are no longer useful - `cp314t` is way better, and we're encouraging new projects to only publish for that. Also `cibuildwheel` defaults to adding `cp314t` but not `cp313t`.

asv/_rangemedian.cpp
```diff
@@ -91,10 +92,12 @@ class Cache
     };
 
     std::map<std::pair<size_t, size_t>, Item> items_;
+    mutable std::mutex mtx_;
 
 public:
     Cache() {};
     bool get(size_t left, size_t right, double *mu, double *dist) const {
+        std::lock_guard<std::mutex> lock(mtx_);
```
ngoldbaum, line 100: I think you have an [attached thread state](https://docs.python.org/3/c-api/threads.html#thread-states-and-the-global-interpreter-lock) here, since you don't detach inside `RangeMedian_mu_dist`, `RangeMedian_mu`, or `RangeMedian_dist`. That means this call can possibly lead to deadlocks with the interpreter on both the GIL-enabled and free-threaded build.

To avoid that, you need to detach from the interpreter before locking native mutexes.

That said, I'm not actually sure whether it makes sense to do that whole dance on the GIL-enabled, given that the GIL is held. So then you'd want to conditionally use locking on the free-threaded build, with `#ifdef Py_GIL_DISABLED` blocks.

I also suspect that despite the existence of the `set` method -- which I don't think is used at all from Python -- in practice this is a cache that is initalized once and then thereafter not modified, so using a single initialization API would probably make more sense than adding a mutex to the type, since you'd only have the possibility of blocking calls when the cache is created and not thereafter. Note that the C++ `call_once` API involves blocking calls, so you need to detach from the interpreter before calling any API that can block.

At a higher level - do you actually want to support sharing RangeMedian objects between threads? A simpler approach would be to not support that, replace the blocking calls to `lock` here with non-blocking `try_lock` calls. Then, if acquiring the lock fails, you can bubble up an error to Python saying that another thread is already using the RangeMedian.

Then it becomes the responsibility of the Python programmer to not share RangeMedian objects between threads. I don't think asv does that, so you'll never hit that error in practice until some future point when you want to add multithreaded parallelism using RangeMedian instances.

I'd also definitely add a multithreaded test. I bet if you wrote a multithreaded test that reads and writes to a RangeMedian from more than one thread using this current PR, you'd hit deadlocks. IMO it's always worthwhile to write tests for nontrivial code to handle multithreaded parallelism.

Issue #1542, opened 2025-09-16
How to make `matrix` mean anything?


flying-sheep:
## Use case

I’m developing a Python package and want to use ASV’s `matrix` feature

## Problem

In the config, `matrix` means little when `install_command` is set up to freely install dependencies:

To be sure that constraints mean something, they need to apply to all modifications to the environment. Ideally, one would only use exactly one solve to create an environment.

E.g. conda is fundamentally wrong in how it handles mixed pip and conda environments, as it *first* installs the conda packages, and *then* the pip packages (the second call is allowed to freely ignore the constraints passed in the first)

## Possible solutions

Therefore there are some ways to actually make the config make sense.

1. Have the single solve happen purely from the matrix deps:

   - Copy all your packages’ dependencies to `matrix`, duplicating them
   - Use an `install_command` that can’t modify any of them, like `["pip install --no-deps {wheel_file}"]`

   I hate the idea. So now I have to keep my deps in sync with the matrix, and I better invent some CI for it or never make a mistake? No thanks.

2. No `matrix`, install all dependencies in `install_command` together with your package’s wheel.

3. Have the solver solve your package’s dependencies together with the constraints from `matrix`.

   - Use rattler or virtualenv (as said, conda is fundamentally incapable of solving pip and conda deps together)
   - `install_command` needs to be `[]`, because
   - `{wheel_file}` is somehow passed to the solver together with the matrix deps.

   This is literally the only correct thing to do.

## Question

I’d like to know how to do 3., as it seems the only way to use asv’s matrix feature that actually makes sense.

flying-sheep:
Note that 2 also doesn’t work, since the command parser is really bad. Can’t even do something like `pip install "mypkg[some-extra] @ {wheel_file}"`, which results in `ValueError: No closing quotation` raised from shlex.

And the sane option of allowing the user to pass a list as command instead of relying on dodgy string parsing is also not available (could even do an object to pass options like `in-dir` instead of inventing ad-hoc syntax!)

PR #1581, opened 2026-03-08
BUG: Handle py-rattler 0.22.0 solve() API change

### LINKED ISSUE #1515: Support the free-threaded build and publish free-threaded wheels

HaoZeke:
py-rattler 0.22.0 renamed the 'channels' parameter to 'sources' in the solve() function.  Detect which name is accepted via inspect.signature so both old and new versions work.

.... I really need to strip the environments out of `asv`. This should unblock the Dependabot stuff like https://github.com/airspeed-velocity/asv/pull/1579 and then with #1580 we should be able to cut a new release with wheels. So it'll close #1515.

### DIFF
--- a/asv/plugins/rattler.py
+++ b/asv/plugins/rattler.py
@@ -1,4 +1,5 @@
 import asyncio
+import inspect
 import os
 import sys
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


flying-sheep:
> .... I really need to strip the environments out of asv.

also asv is fundamentally broken regarding environments, see #1542.

Issue #1582, opened 2026-03-19
asv quickstart fails silently if .gitignore already exists


abussy-pasqal:
Hi,

I'm new to `asv` and I was following the tutorial https://asv.readthedocs.io/en/stable/using.html. I couldn't generate `asv.conf.json` with the command `asv quickstart` which returned
```bash
root@07f9fb7a1dcf:~/asv-tutorial# asv quickstart
· Setting up new Airspeed Velocity benchmark suite.

Which of the following template layouts to use:
(1) benchmark suite at the top level of the project repository
(2) benchmark suite in a separate repository

Layout to use? [1/2] 1

· Template content already exists.
· Edit asv.conf.json to continue.
```
I found out that it was because a `.gitignore` file already existed in the directory.

### Steps to reproduce ###
In a docker container `python:3.14-slim` :
```bash
python -m pip install asv
touch .gitignore
asv quickstart --top-level
```
Nor the `asv.conf.json` or the `benchmarks` folder are created.

### Expected behavior ###

Either of :
1. `.gitignore` generation skipped
2. `.gitignore` merged
3. Failure with a meaningful error
4. or something else

The message `· Template content already exists.` hints at it to some extent, but is not explicit enough for newcomers.

PR #1583, opened 2026-03-28
quickstart: Fix silent failure when .gitignore existsThis fixes issue…

harshil2424:
… #1582 where asv quickstart aborted if a .gitignorewas present. The command now appends ASV-specific patterns to the existingfile instead of treating it as a template conflict.

### DIFF
--- a/asv/commands/quickstart.py
+++ b/asv/commands/quickstart.py
@@ -73,6 +73,8 @@ class Quickstart(Command):
 
         template_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', 'template')
         for entry in os.listdir(template_path):
+            if entry not in ("asv.conf.json", "benchmarks"):
+                continue
             path = os.path.join(template_path, entry)
             dest_path = os.path.join(dest, entry)
             if os.path.exists(dest_path):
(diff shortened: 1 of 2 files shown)


harshil2424:
Fixes #1582

HaoZeke (changes requested):
IMO this is a rather large change to the existing behavior. My preferred fix would be a noisier / clearer error message instead of changing the way `asv` works.

mtryan83:
I followed roughly the same steps in #1582 and wound up here. The problem I see with just changing the error message is the user then needs to 
1. create/find their own [`asv.conf.json`](asv/template/asv.conf.json)  template file
    - the template isn't provided in the documentation, but all the fields are listed
    - it wasn't _entirely_ clear where the config file went (top-level? benchmarks folder?), but that might have been just a misreading/misunderstanding on my part
3. Create a benchmarks folder and populate it with [`benchmarks.py`](asv/template/benchmarks/benchmarks.py) and empty `__init__.py`.
5. Check/create their `.gitignore` to verify they're ignoring the same things (slightly out-of-date, standard python `.gitignore` with a typo).

Note that this bug will also occur if the `benchmarks` folder already exists, even if `benchmarks.py`/`__init__.py` don't.

**Suggestions** 
1. The `"template content already exists"` message should be expanded to include *what* template content already exists.
2. Add link(s) to missing template files/contents somewhere
    -  Since the contents of `benchmarks.py` immediately follow in the documentation, and they'll need to be overwritten anyway, this is less of an issue
3. Instead of modifying the .gitignore file, copy the missing files and leave the already existing files alone.

I think (1) is the minimum to fix the issue, but (2) and/or (3) would be very helpful.

PR #1591, opened 2026-05-09
Add start and end hash to url

### LINKED ISSUE #1590: Subset commits via URLs
In https://github.com/pandas-dev/asv-runner/, we often have many commits when viewing benchmarks. It'd be great to be able to link to a certain range of commits when a regression is identified so that users do not have to subset it themselves for the time series to become more readable.

Concretely, I would like to request the ability to have `&start_hash=...` and/or `&end_hash=...` in the URL to subset the range of commits. Always happy to bikeshed the names.

In addition, it would be nice, but is not necessary, for the URL to automatically update when the user subsets the commit range themselves.

I've opened #1591 as a proof-of-concept of this feature.

rhshadrach:
Closes #1590
 
This PR will need tests and changelog if it is to go forward, holding off on those until there is feedback for the idea.

This PR was also entirely written by Claude Code. I have reviewed all the changes and it looks solid, but I haven't touched javascript since the early 2000s.

### DIFF
--- a/asv/www/graphdisplay.js
+++ b/asv/www/graphdisplay.js
@@ -42,6 +42,11 @@ $(document).ready(function() {
     var highlighted_revisions = null;
     /* Whether benchmark graph display was set up */
     var benchmark_graph_display_ready = false;
+    /* Commit-range filter, set via start_hash/end_hash URL params. */
+    var start_hash = null;
+    var end_hash = null;
+    var start_revision = null;
+    var end_revision = null;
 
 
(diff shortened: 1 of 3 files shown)


rhshadrach:
@HaoZeke - this should be ready. I expect the CI should pass but haven't gotten the tests to run tests locally (will put in more effort here if CI isn't green). Could you approve running?

rhshadrach:
Thanks @HaoZeke!

PR #1611, opened 2026-08-07
Add failure summary to the end of the run

### LINKED ISSUE #1201: Add failures to the end of the output
When running `asv run`, if anything fails, it's only available in the middle of a possibly huge log. Finding the errors can be quite time consuming, and it'd probably be helpful to have them repeated at the end of the log, similar to what pytest does.

@mroeschke implementing this in asv would be equivalent to what you proposed in pandas, right?

xref https://github.com/pandas-dev/pandas/pull/46599#pullrequestreview-929237272
xref https://github.com/scikit-image/scikit-image/blob/main/.github/workflows/benchmarks.yml#L81

larsoner:
Closes #1201

Related to #1610, it would have been really nice to have a failure summary at the end of the test run. This adds such failure reporting, like taking this on `main`:

```
               True     True    65536   1.38±0ms
               True     True    65537   4.76±0ms
               True     True    65538   1.65±0ms
             ======= ========= ======= ==========

$ echo $?
2
```

and turning it into this on this PR:

```
             ======= ========= ======= ==========

[50.00%] · Failures (4):
           FAILED fft_basic.FftThreading.time_fft -- exit status 1
           FAILED fft_basic.FftThreading.time_fftn -- exit status 1
           FAILED signal.Convolve.time_convolve2d -- exit status 1
           FAILED signal.Convolve.time_correlate2d -- exit status 1
```

The 50% above is fixed separately by #1610.

Changes drafted by Claude Opus 5, but I understand all code (this is the sort of thing I've previously plumbed myself in `pytest` for MNE-Python, etc.).

### DIFF
--- a/asv/commands/run.py
+++ b/asv/commands/run.py
@@ -14,7 +14,7 @@ from ..console import log
 from ..machine import Machine
 from ..repo import NoSuchNameError, get_repo
 from ..results import Results, get_existing_hashes, iter_results_for_machine_and_hash
-from ..runner import run_benchmarks, skip_benchmarks
+from ..runner import JSON_ERROR_RETCODE, run_benchmarks, skip_benchmarks
 from . import Command, common_args
 from .setup import Setup
 from .show import Show
@@ -310,6 +310,8 @@ class Run(Command):
 
         # Track failures across the run command
         failures = False
+        failed_benchmarks = []
+        failed_builds = []
 
         # Comparison period for date_period filtering
         old_commit_hashes = None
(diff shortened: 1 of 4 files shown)


HaoZeke (changes requested):
this will double-count failures when `--interleave-rounds` is on (the default unless you are using an existing env). collection is inside the outer `iter_rounds_commits()` loop, and `errcode` is not loaded back from disk, so a `rounds=2` timing fail shows up twice. same story for `FAILED build`.

I'd dedup on `(commit, env, name)` / `(commit, env)`. the new tests only use `existing_env_conf` so they miss it.

larsoner:
Okay I took at shot at implementing those changes!

PR #1614, opened 2026-08-15
results: Bail out gracefully when profile data is missing

### LINKED ISSUE #1459: MAINT: Bail out more gracefully if profile data isn't present
Currently loading a benchmark without profile data present and checking for it raises, which could well be a user warning instead.

```python
File ~/Git/Github/asvWork/asv/asv/results.py:564, in Results.get_profile(self, benchmark_name)
    549 """
    550 Get the profile data for the given benchmark name.
    551 
   (...)
    561 
    562 """
    563 profile_data = self._profiles[benchmark_name]
--> 564 profile_data = profile_data.encode('ascii')
    565 profile_bytes = zlib.decompress(base64.b64decode(profile_data))
    566 return profile_bytes

AttributeError: 'NoneType' object has no attribute 'encode'

In [64]: b1.get_profile('benchmarks.TimeSuite.time_add_arr')
```

Doesn't really matter but still, a better error / warning would be nicer.

chiruu12:
Closes #1459.

`get_profile` indexed `_profiles` directly and then called `.encode()` on the result, so a benchmark with no stored profile raised `KeyError`, and one stored as `None` raised `AttributeError`. `add_result` only writes the entry when the profile is truthy, so both shapes are reachable.

```
has_profile('absent')  -> None
get_profile('absent')  -> KeyError: 'time_absent'
get_profile(None-valued) -> AttributeError: 'NoneType' object has no attribute 'encode'
```

It now raises `UserError`, which `main.py` already turns into a logged error and a non-zero exit rather than a traceback. Of the three call sites, two guard with `has_profile` first, and the third is `asv profile` right after a profiling run, which is the one that surfaced this.

Also corrects the docstring, which promised `pstats.Stats` but returns raw bytes. `get_profile_stats` has been the Stats-returning method since #1253.

Added a test covering both shapes. It fails on main with `KeyError` and passes here.

`pytest test/test_profile.py test/test_results.py` passes, 11 tests. `pre-commit run --files` passes on all changed files.

### DIFF
--- a/asv/results.py
+++ b/asv/results.py
@@ -549,11 +549,22 @@ class Results:
 
         Returns
         -------
-        profile_data : pstats.Stats
-            Profile data
+        profile_data : bytes
+            Raw profile data. Use `get_profile_stats` for a `pstats.Stats`
+            object.
+
+        Raises
+        ------
+        UserError
+            If no profile was stored for the benchmark.
 
         """
-        profile_data = self._profiles[benchmark_name]
+        profile_data = self._profiles.get(benchmark_name)
+        if not profile_data:
+            raise util.UserError(
+                f"No profile data available for benchmark '{benchmark_name}'. "
+                f"Re-run the benchmark with --profile to collect it."
+            )
         profile_data = profile_data.encode('ascii')
         profile_bytes = zlib.decompress(base64.b64decode(profile_data))
         return profile_bytes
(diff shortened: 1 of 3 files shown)


chiruu12:
Anything you need from me on this one?

`get_profile` raised `KeyError` for an absent profile and `AttributeError` for one stored as
`None`. It now raises `UserError`, which `main.py` already turns into a logged error and a
non-zero exit rather than a traceback. Happy to rebase.

PR #1619, opened 2026-08-29
Add average_over to collapse a parameter across graphs

hmaarrfk:
I would like to simply "ignore" the python version in my benchmark plots. I find that it is just noise since for us, python doesn't affect the performance as much as other factors.

We let it float in our ASV CIs, so when they pickup different pythons for whatever reason, it is not really useful to plot.

Instead of trying to special case things, at visualization time, i'm trying to allow some improved "average over" tools to help customize how things are displayed, given the same backend storage of results.


hmaarrfk:
as much as i would love to see this, i don't think i have enough time to understand the current state of the ASV code to make a useful contribution along this front.

